PR #16, opened 2020-02-17
Function reference

martingallagher:


docs/func.md
```diff
@@ -0,0 +1,578 @@
...
+(sorted-map)
+elps> (assoc! test "1" 1)
+(sorted-map "1" 1)
+elps> test
+(sorted-map "1" 1)
+```
+
+## `dissoc`
+
+Disassociates a value from a map via a key, returning a copy without mutating
```
bmatsuo-at-luthersystems: It may be better to use the word "Dissociate" instead of "Disassociate" (also for `dissoc!` below).

docs/func.md
```diff
@@ -0,0 +1,578 @@
...
+evaluating from the right.
+
+```Lisp
+elps> (foldr - 0 '(1 2 3))
+2 ; (1 - (2 - (3 - 0)))
+```
+
+## `unpack`
+
+Unpacks the list supplying the values as parameters to the given function.
```
bmatsuo-at-luthersystems: `unpack` is actually a special case of `apply` and is more cumbersome. Instead of using `unpack` it is better to use `apply` and `funcall`. I would like to call this function deprecated.

docs/func.md
```diff
@@ -0,0 +1,578 @@
...
+Converts a string value to bytes.
+
+```Lisp
+elps> (to-bytes "ABC123")
+#<bytes 65 66 67 49 50 51>
+```
+
+## `to-int`
+
+Converts string digits ([0-9]+) and floats to integers. For floats the mantissa
```
bmatsuo-at-luthersystems: I don't think "mantissa" is the correct word to use. I would say "fractional part" or "fractional component."

PR #32, opened 2021-01-30
RFC - `format` implementation

reubent:
The implementation of `format` is cool but probably overcomplex for what it actually does. There's also a `TODO` to add positional parameters.

I've written a (much) simpler implementation that supports positional parameters. You'll find it in `format_test.go` as `quickformatBuiltin` at the moment. I've also written a series of tests and benchmarks for both implementations.

As it stands, there is code there to support unmatched brackets, but the more I think about it, the less desirable I think that actually is. Mistakes are obvious because the number of brackets won't match the number of parameters and this gives a way to use a literal curly bracket. Considering the performance impact of it as well, this seems reasonable.

With unmatched checking turned off, performance for this version is around 20% better for simple strings:

```
BenchmarkQuickFormatString
BenchmarkQuickFormatString-8                      	  727240	      1533 ns/op
```
versus
```
BenchmarkFormatString
BenchmarkFormatString-8                           	  631561	      1887 ns/op
```

With a lot of tokens, the gap is a lot wider:

```
BenchmarkQuickFormatStringWithATonOfTokens
BenchmarkQuickFormatStringWithATonOfTokens-8      	  116503	     10249 ns/op
```
versus
```
BenchmarkFormatStringWithATonOfTokens
BenchmarkFormatStringWithATonOfTokens-8           	   69565	     16725 ns/op
```

Performance is marginally slower with positional tokens, but not a lot:

```
BenchmarkQuickFormatStringWithPositionalToken
BenchmarkQuickFormatStringWithPositionalToken-8   	  727245	      1666 ns/op
```

Turning on the simple bad token check, performance drops to:
```
BenchmarkQuickFormatString
BenchmarkQuickFormatString-8                      	  599984	      1723 ns/op
BenchmarkQuickFormatStringWithATonOfTokens
BenchmarkQuickFormatStringWithATonOfTokens-8      	   88561	     13319 ns/op
```

and if I use the "comprehensive" check:
```
BenchmarkQuickFormatString
BenchmarkQuickFormatString-8                      	  380971	      2916 ns/op
BenchmarkQuickFormatStringWithATonOfTokens
BenchmarkQuickFormatStringWithATonOfTokens-8      	   18986	     63204 ns/op
```

### DIFF
--- a/lisp/format_test.go
+++ b/lisp/format_test.go
@@ -1,0 +1,185 @@
+package lisp
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFormatStringSimple(t *testing.T) {
+	env := NewEnv(nil)
+	format := String("Swap a symbol here {}")
+	arg1 := String("yeah")
+	args := &LVal{Cells: []*LVal{format, arg1}}
+	formatted := builtinFormatString(env, args)
+	assert.Equal(t, "Swap a symbol here yeah", formatted.Str)
+}
+
+func TestFormatStringComplex(t *testing.T) {
+	env := NewEnv(nil)
+	format := String("Swap a symbol here ")
+	built := format.Str
+	args := []*LVal{format}
+	for x := 0; x < 100; x++ {
+		args[0] = String(fmt.Sprintf("%s {}, blah blah,", args[0].Str))
+		built = fmt.Sprintf("%s %d, blah blah,", built, x)
+		args = append(args, String(strconv.Itoa(x)))
+		formatted := builtinFormatString(env, &LVal{Cells: args})
+		assert.Equal(t, built, formatted.Str)
+	}
+}
+
+func TestFormatStringUnmatched(t *testing.T) {
+	env := NewEnv(nil)
+	format := String("Swap a symbol here }")
+	arg1 := String("yeah")
+	args := &LVal{Cells: []*LVal{format, arg1}}
+	formatted := builtinFormatString(env, args)
+	assert.Equal(t, LError, formatted.Type)
+}
+
+func TestTokenizeFormatString(t *testing.T) {
+	x := tokenizeFormatString("This is a formatted {} string yeah {}.")
+	assert.Equal(t, []formatToken{
+		{typ: formatText, text: "This is a formatted "},
+		{typ: formatOpen, text: "{"},
+		{typ: formatClose, text: "}"},
+		{typ: formatText, text: " string yeah "},
+		{typ: formatOpen, text: "{"},
+		{typ: formatClose, text: "}"},
+		{typ: formatText, text: "."},
+	}, x)
+}
+
+var formatTokenRegexp = regexp.MustCompile(`\{(\d+)\}`)
+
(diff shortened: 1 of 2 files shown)


lisp/format_test.go
```diff
@@ -0,0 +1,185 @@
...
+
+var formatTokenRegexp = regexp.MustCompile(`\{(\d+)\}`)
+
+// var formatBadTokenPattern = regexp.MustCompile(`(?:\{[^\d\}]|[^\{\d]\})`) // This is slooooow...
+// var formatBadTokenPattern = regexp.MustCompile(`\{[^\d\}]`) // This is also slow enough that I'd ditch it
+
+func quickformatBuiltin(env *LEnv, args *LVal) *LVal {
+	format := args.Cells[0]
+	//if formatBadTokenPattern.MatchString(format.Str) {
+	//return env.Errorf("unmatched token")
```
bmatsuo-at-luthersystems, line 66: I think there should be a test/error for unmatched tokens.  The existing format-string does this.

bmatsuo-at-luthersystems (changes requested):
Interesting benchmarks. The existing format-string does seem a bit excessive given that it hasn't really needed any extension since it was written. So I'd be open to a simplification.

I think there are some cut corners with your replacement functions. I would be interested to know what the benchmark numbers would look like if those issues were to be addressed.

lisp/format_test.go
bmatsuo-at-luthersystems: Ah. Maybe this was trying to catch unclosed curly braces? What I meant was that format-string will report an error if you don't supply enough values.

```
elps> (format-string "hello {}, it's {} to meet you" "world")
stdin:1: lisp:format-string: too many formatting direcives for supplied values
Stack Trace [1 frames -- entrypoint last]:
  height 0: stdin:1: lisp:format-string
```

reubent: This should do that - it's just `fmt.Sprintf` inside, so it'll handle it the same way that Go does with a warning in the output.

bmatsuo-at-luthersystems: Right. fmt just outputs junk when there's a mismatch between argument and substitutions. But I don't think that is very good behavior for a language like lisp. I think a runtime error is preferable.

The annoying thing about the fmt package is that format argument mismatches can literally go unnoticed for onths or years. I think they chose to do that instead of forcing programmers to check errors for every formatting call. I've never really seen another argument for why formatting _should_ work that way (though I guess I haven't looked hard if you had one).

In dynamic languages like python and ruby format argument mismatches cause runtime exceptions. Even if it's not completely preferable, I think it's more expected of a language like elps.

sam-at-luther:
Closing this in favor of #97, which implements the features discussed here:

- **Positional parameters** (`{0}`, `{1}`, etc.) — the original motivation for this PR
- **Performance improvements** — single-pass scanner replaces the 3-phase tokenizer, benchstat shows 41% faster (geomean)
- **`}}` escape bug fix** — the close-brace escape was broken in the original implementation
- **Proper error handling** — runtime errors for mismatched args, unclosed braces, invalid directives, and out-of-range indices (addresses Bryan's review feedback about not silently producing junk)

Thanks @reubent for the original RFC — the benchmarking approach and the idea of simplifying the implementation directly informed the new design.

PR #36, opened 2021-03-07
New LType, LTaggedVal, a container for user-defined data types

bmatsuo-at-luthersystems:
LTaggedVal is an arbitrary lisp value that has been "tagged" with a
custom type name (symbol).  That is, an LTaggedVal is an instance of a
user-defined type and the value contained within is the instance data.
The following functions and macros have been defined for working with
LTaggedVal values:

- ``deftype''
  Define a new type -- binding its name to a typedef
- ``new''
  Instantiate a tagged-value from a typedef
- ``type''
  Returns a symbol representing the type of its argument (works for
  any value)
- ``type?''
  Returns a boolean -- true iff the given type specifier matches the
  type of the given object
- ``tagged-value?''
  Returns a bool -- true iff the given object is a tagged-value
- ``tagged-value''
  Unwraps tagged-value and returns the underlying value

Examples of user-defined types: _examples/random/oop.lisp


sam-at-luther:
@reubent this can be the foundation for building a type system on elps. See discussion on elps matrix channel.

_examples/random/oop.lisp
```diff
@@ -0,0 +1,236 @@
...
+(debug-print r)
+(debug-print (var r :numer))
+(debug-print (var r :denom))
+(debug-print (invoke r :to-float))
+(debug-print (thread-first r
+                           (invoke :mul 5)
+                           (invoke :to-float)))
+(debug-print (thread-first r
+                           (invoke :mul (new rational 1 2))
+                           (invoke :to-float)))
```
bmatsuo-at-luthersystems: I think these should actually be assert statements

_examples/random/oop.lisp
```diff
@@ -0,0 +1,236 @@
...
+;; Tests using the oop:complex type.
+(set 'c (new complex 0 1))
+(debug-print c)
+(debug-print (invoke c :real))
+(debug-print (invoke c :imag))
+(debug-print (complex? c))
+(debug-print (methods c))
+(debug-print (methods complex))
+(debug-print (methods 'oop:complex))
+(debug-print (methods 'complex)) ;; nil -- requires qualified symbols
```
bmatsuo-at-luthersystems: These should also be assert statements

lisp/lisp.go
```diff
@@ -77,8 +84,13 @@ const (
...
 	LArray
 	// LNative values store a Go value in the LVal.Native field and can be used
-	// by builtin functions store references to values of any type.
+	// by builtin functions to store values of any type.
 	LNative
+	// LTaggedVal is a user-defined type that uses the following fields in an
+	// LVal:
+	// 		LVal.Str      The user-defined type name
+	// 		LVal.Cells[0] The user-data for the typed-value
+	LTaggedVal
```
bmatsuo-at-luthersystems, line 93: I'm not crazy about the name "tagged-value". But I wasn't sure what else to use and I think I got "tagged-value" from SICP. I didn't really want to call it an "instance".

I was also thinking about the name "user-data" -- data with a user-defined type. I think this name sounds better but in the future some "user data" may be part of the core language (arrays, maps, ...)

andrei-at-luther (approved):

lisp/lisp.go
bmatsuo-at-luthersystems: I think the term "tagged-value" is fine along with `LTaggedVal`. But after mulling it over I think I dislike the lisp function `(tagged-value v)` enough to change it -- it's a very confusing function name. I think the data in a tagged-value should be called "user-data", data managed by the "user" (program) that defined the type. So you'd write something like

```lisp
(if (tagged-value? v)
  (do-something (user-data v))
  (error 'type-error "argument is not a tagged-value")
```

And for the function `new` you would say "new takes a typedef and returns a tagged-value with the given type and user-data returned by the type constructor after being invoked with new's remaining arguments".

lisp/lisplib/libschema/README.md
```diff
@@ -16,26 +16,20 @@ Types are defined using the `s:deftype` keyword and validations are performed by
...
 
 If the value does not have the required type, an error of type `"wrong-type"` will be returned. If a constraint (see below)
 fails, an error of type `"failed-constraint"` will be the result.
 
-The type passed to `s:validate` con be either a primitive type validator
-function or it can be a lisp:typespec created by by the `lisp:deftype` macro,
-in which case `s:validate` checks the typedef matches the type of the given
-value and then run any type validation that was defined with `s:deftype`.
+If using `s:make-validator` to define a time then pass the validator it
+returned along weth the value to be validated.
```
sam-at-luther: `weth` is this a typo?

lisp/lisplib/libschema/README.md
```diff
@@ -59,17 +53,38 @@ Or, more usefully, if we want to define an enum, we can specify a list of permit
 (s:deftype "title" s:string (s:in ("Mr","Mrs","Miss","Ms","Mx","Dr","Prof")))
 ```
 
-For types defined with `lisp:deftype`, `s:deftype` accepts a lisp:typedef value
-and the following type declaration and constraints apply to the tagged-value
-user data.
+If working with tagged-values (user-defined types created with the core
+language `deftype` macro) then `s:deftype` often will not be usable because
+both deftype operations attempt to bind the same symbol.  Instead of using
+`s:deftype` in those situations`s:make-validator` will return a validotor which
```
sam-at-luther: `validotor` -> `validator`

lisp/lisplib/libschema/README.md
```diff
@@ -59,17 +53,38 @@ Or, more usefully, if we want to define an enum, we can specify a list of permit
...
 
-In this situation you cannot pass a string as the first argument of `s:deftype`
-or it will not recognize that it is working with tagged-values.
+The `s:make-validator` function works with any data type, not just
+tagged-values.  It can be used to create scoped validators with a limited
+lifetime.
+
+```lisp
+(let ([v (s:make-validator "sequence-elemeent" s:sorted-map)])
+    (map '() #^(s:valiadte v %) sequence))
```
sam-at-luther: `s:valiadte` -> `s:validate`

lisp/lisplib/libschema/README.md
```diff
@@ -59,17 +53,38 @@ Or, more usefully, if we want to define an enum, we can specify a list of permit
...
-```lisp
+```
+(deftype abc (s) (to-string s))
+(set 'abc-validator (s:make-validator abc s:string (s:in '("a" "b" "c")))
+```
+
+When `s:make-validator` is passed the typedef `abc` it automatically creates a
+tagged-value validator which validates the type's string contents.
+
+If the structure of a tagged-value is known but it's exact type is not then the
```
sam-at-luther: it's -> its

lisp/lisplib/libschema/libschema.go
```diff
@@ -296,18 +246,42 @@ func GenSymbol() string {
...
 			return lisp.ErrorConditionf(WrongType, "Input was not a boolean for type %s", name)
 		}
-		return builtinCheckAny(constraints).FunData().Builtin(env, input)
+		return builtinCheckAny(env, constraints).FunData().Builtin(env, input)
+	})
+}
+
+// Checks constraints and type for values with a user-defined type.
+func builtinCheckTaggedVal(env *lisp.LEnv, name string, constraints []*lisp.LVal) *lisp.LVal {
+	fmt.Fprintln(env.Runtime.Stderr, "TAGGED_VAL", constraints)
```
sam-at-luther: did oyu mean to keep this?

lisp/lisplib/libschema/README.md
bmatsuo-at-luthersystems: At this point I think I wrote too many docs for this feature and just gave up.

sam-at-luther (approved):

PR #40, opened 2021-05-31
JSON performance improvements

bmatsuo-at-luthersystems:
- Add json benchmarks using a github api response

- Added a custom json encoder which avoids serializing an intermediate
  representations and instead directly marshals LVals.

- Changed the implementation of `sorted-map` to use an interface which
  the json package is able to satisfy, avoiding extra map allocations.

### DIFF
--- a/lisp/embed.go
+++ b/lisp/embed.go
@@ -2,6 +2,8 @@
 
 package lisp
 
+import "reflect"
+
 // True interprets v as a boolean and returns the result.
 //
 // NOTE:  I don't like this name, really.  But I can't think of a better one.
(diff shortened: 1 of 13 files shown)


bmatsuo-at-luthersystems:
```
name                   old time/op  new time/op  delta
Package/$load-8        1.34ms ± 2%  1.49ms ± 2%  +10.76%  (p=0.000 n=10+9)
Package/load-object-8  3.64ms ± 2%  3.14ms ± 1%  -13.70%  (p=0.000 n=9+8)
Package/load-array-8   3.69ms ± 2%  3.86ms ± 1%   +4.62%  (p=0.000 n=10+9)
Package/load-nested-8  5.50ms ± 1%  4.36ms ± 1%  -20.68%  (p=0.000 n=9+10)
Package/load-github-8   113ms ± 3%    81ms ± 1%  -28.38%  (p=0.000 n=10+9)
Package/dump-object-8  4.02ms ± 1%  2.89ms ± 1%  -28.25%  (p=0.000 n=9+9)
Package/dump-array-8   2.62ms ± 1%  1.99ms ± 0%  -23.90%  (p=0.000 n=9+8)
Package/dump-nested-8  6.55ms ± 1%  4.22ms ± 1%  -35.48%  (p=0.000 n=10+8)
Package/dump-github-8   149ms ± 2%    56ms ± 1%  -62.23%  (p=0.000 n=10+9)
```

lisp/embed.go
```diff
@@ -142,16 +142,11 @@ func GoMap(v *LVal) (map[interface{}]interface{}, bool) {
...
-		case MapSymbol:
-			m[string(k.(MapSymbol))] = vgo
-		default:
-			m[k] = vgo
-		}
+	data := v.Map()
+	m := make(map[interface{}]interface{}, data.Len())
+	for _, pair := range sortedMapEntries(data).Cells {
+		// FIXME:  typecheck pair
+		m[GoValue(pair.Cells[0])] = GoValue(pair.Cells[1])
```
sam-at-luther: what's the consequence of not checking the type here? it only effects embedded use cases correct?

lisp/lisplib/libjson/sortedmap.go
```diff
@@ -0,0 +1,92 @@
+package libjson
```
sam-at-luther, line 1: stupid question, but is `key?` not needed in the interface here?

lisp/embed.go
bmatsuo-at-luthersystems: Well, if you assume that `pair.Cells[0]` could be *any* type then it would be possible for this to panic. Because GoValue can return things like `[]interface{}`, `map[string]interface{}`, `error` and potentially other things which are not valid map keys.

For the existing sorted-map implementations this can't actually panic because `GoValue(pair.Cells[0])` will always return a string. So if someone wanted to extend their application so that sorted-maps could use vectors as keys (or something like that) then this probably wouldn't be sufficient.

IDK. I guess I could at least elaborate on what risks this poses. This function signature doesn't have a lot of room for signaling an failure because of bad map keys... I guess you document that returning `nil, true`  means the map had invalid keys. That seems a bit random but I guess it could work.

There's a similar FIXME comment in lisp.go... I might be able to do some better error signalling there.

lisp/lisplib/libjson/sortedmap.go
bmatsuo-at-luthersystems: There is no direct equivalent to the lisp function `key?`. The Go interface uses a Get with two return values `Get(k *LVal) (*LVal, ok)`. So the go interface matches go map semantics, you can't check for the presence of a key without also looking up the value associated with the key.

I think in the future `key?` would probably be implemented through `get`. The only issue is that lisp doesn't directly support multiple return types but if `get` had an optional continuation argument then you can compensate for that.

```lisp
(defun get (m k &optional continuation) ...)

(set 'm (sorted-map "a" 123))

(defun key? (m k) (get m k (lambda (x ok) (if ok true false))))

(defun get-default (m k def) (get m k (lambda (x ok) (if ok x def))))
```

sam-at-luther (approved):

bmatsuo-at-luthersystems:
I did add a test helper, `elpstest.AssertSortedMap(*testing.T, lisp.Map)` that tests implementations of the new `lisp.Map` interface to make sure they will work with the core language and standard library.

PR #41, opened 2021-06-05
User types example code

bmatsuo-at-luthersystems:
This is the code from the talk I gave, along with an example which I didn't have time to cover in the main talk.

### DIFF
--- a/_examples/user-defined-types/address-book.lisp
+++ b/_examples/user-defined-types/address-book.lisp
@@ -1,0 +1,20 @@
+(set 'address-book (sorted-map))
+
+(defun make-contact-info (name phone)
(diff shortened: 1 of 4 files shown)


bmatsuo-at-luthersystems:
Ugh. Something happening on this branch. I meant to put this code under _examples.

jeremy-asher:
Nice, thanks! Is there any significance to the `_solved` suffix on some of the filenames?

bmatsuo-at-luthersystems:
Ah. Right. I guess I started that because address-book.lisp was representing the "problem" and the other files were the "solution" to the problem.

jeremy-asher:
Ah, gotcha.

jeremy-asher (approved):

PR #48, opened 2021-07-21
JSON encoding escapes html-unsafe characters &, <, >

bmatsuo-at-luthersystems:
This behavior is consistent with encoding behavior prior to v1.14.0 as
escapeHTML is enabled by default in encoding/json.


sam-at-luther (approved):

bmatsuo-at-luthersystems:
This fixes #49

Issue #46, opened 2021-06-18
documentation parsed from code


bmatsuo-at-luthersystems:
Common-lisp style (aka python style) docstrings are technically a breaking change but also seem drastically easier to implement. Perhaps if docstrings are used then a hypothetical "linter" could catch cases where a "constant function" is ambiguous about whether its intent is to return a constant string or a nil value.

Of course, go-style doc comments have significant benefits. Beyond being backwards compatible it becomes possible to document arbitrary symbols within a package.

Implementations have to consider how function documentation works after symbols have been "imported" to the user's package (i.e. via `use-package`). If a user in a repl imports `regexp-match?` from the regexp package and then tries to evaluate a hypothetical `(help regexp-match)` expression then how does that resolve to the original documentation for that function?

Even the implementation of such a repl-based doc lookup is not a certainty.

PR #50, opened 2021-08-11
Docstring implementation with repl help package and cli doc command

bmatsuo:
``` 
elps doc map
```

```
 $ elps repl
 > (help map)
```


bmatsuo-at-luthersystems:
Closes #46

lisp/builtins.go
```diff
@@ -39,121 +51,129 @@ func (fun *langBuiltin) Eval(env *LEnv, args *LVal) *LVal {
...
+	{"append-bytes!", Formals("bytes", "values"), builtinAppendBytesMutate, ""},
+	// NOTE:  ``append-bytes'' does not accept a type-specifier argume, ""nt
+	// because 'string would be equivalent to (concat 'string .., "".)
+	{"append-bytes", Formals("bytes", "byte-sequence"), builtinAppendBytes, ""},
+	{"aref", Formals("a", VarArgSymbol, "indices"), builtinARef, ""},
+	{"length", Formals("seq"), builtinLength, ""},
+	{"empty?", Formals("seq"), builtinIsEmpty, ""},
+	{"cons", Formals("head", "tail"), builtinCons, ""},
+	{"not", Formals("expr"), builtinNot, ""},
+	// true? is potentially needed as a boolean conversion function (for jso, ""n)
```
jeremy-asher: Looks like a quoting problem came up in some of the comments.

lisp/lisplib/libhelp/libhelp.go
```diff
@@ -0,0 +1,199 @@
...
+		will also be printed.
+		`),
+}
+
+func opHelp(env *lisp.LEnv, args *lisp.LVal) *lisp.LVal {
+	name := args.Cells[0]
+	if name.Type != lisp.LSymbol {
+		return env.Errorf("argument is not a symbol: %v", lisp.GetType(name))
+	}
+	err := RenderVar(env.Runtime.Stderr, env, name.Str)
```
jeremy-asher, line 62: Would it be useful to have these functions return the docs in some format for programmatic use or would that be too detrimental to the REPL UX?

lisp/lisplib/libhelp/libhelp.go
```diff
@@ -0,0 +1,199 @@
...
+	if lisp.True(printAll) {
+		for sym := range pkg.Symbols {
+			_, err := fmt.Fprintln(env.Runtime.Stderr, sym)
+			if err != nil {
+				return env.Error(err)
+			}
+		}
+	} else {
+		exports := pkg.Externals
+		for _, exsym := range exports {
```
jeremy-asher, line 103: Looks like this should probably be ordered somehow if possible before printing to match the output without `printAll` being true.

lisp/lisplib/libhelp/libhelp.go
```diff
@@ -0,0 +1,199 @@
...
+
+func RenderPkgExported(w io.Writer, env *lisp.LEnv, query string) error {
+	pkg := env.Runtime.Registry.Packages[query]
+	if pkg == nil {
+		return fmt.Errorf("no package: %q", query)
+	}
+	exports := pkg.Externals
+	for i, exsym := range exports {
+		if i > 0 {
+			_, err := fmt.Fprintln(w)
```
jeremy-asher, line 157: Looks like this produces extra newlines, especially visible when a function has no documentation.

jeremy-asher (commented):
Looks good! Just found a few minor issues and one question.

lisp/lisplib/libhelp/libhelp.go
bmatsuo-at-luthersystems: Having looked at the output again I think it's better looking this way and I think a little more in line with python's `help` function. I guess one thing python does is group variables (non-functions).

If you remove the extra newlines I don't really think it looks good or consistent when you have a mix of document and undocumented procedures (it's like, take your pick of one). Maybe you are imagining something a little different from me.

The way I see it this text rendering isn't really meant to be held to any backcompat requirements for the time being (though I am fixing the nondeterminism you mentioned). I think we should be free to change the rendering/ordering and do something a bit fancier in the future. I'll just document that for now.

lisp/lisplib/libhelp/libhelp.go
bmatsuo-at-luthersystems: Presumably you are talking about lisp code accessing function docs. I think the exported go functions here should be adequate for go programs that embed elps to access documentation.

It's an interesting idea but I have some questions about how exactly that works as things evolve. I think there may be some time where documentation is somehow stripped out at "compile" time to speed up initialization and potentially break packages which were relying on runtime access to documentation. That's why I left things rather opaque to lisp code for now.

lisp/lisplib/libhelp/libhelp.go
bmatsuo-at-luthersystems: oh. I think I fixed a different line number with this. oops.

I actually made the printAll version deterministic because it was ranging over a map. This one should be deterministic, it's just not sorted.

I'm a little torn on whether or not to respect the definition-order of exported functions. I think this is the correct call because you might group a set of functions together logically even if their names aren't lexically adjacent. At least that's my reasoning at the moment.

lisp/lisplib/libhelp/libhelp.go
jeremy-asher: Gotcha. 👍

lisp/lisplib/libhelp/libhelp.go
jeremy-asher: Ah.  I guess I was comparing to `go doc` which appears to sort things alphabetically.

jeremy-asher (approved):

PR #59, opened 2023-12-16
Make REPL testable from go test

sam-at-luther:


sam-at-luther:
@jeremy-asher this last commit removes the ugly Stack data structure, using the optimization we discussed

lisp/profiler.go
```diff
@@ -4,17 +4,14 @@ const ElpsVersion = "1.7"
...
+	// SetFile sets the file to output to
 	SetFile(filename string) error
-	// End the profiling session and output summary lines
+	// Complete ends the profiling session and output summary lines
 	Complete() error
-	// Marks the start of a process
-	Start(function *LVal)
-	// Marks the end of a process
-	End(function *LVal)
+	// Start the start of a process, and returns a function to stop.
```
jeremy-asher: `Starts the process` perhaps?

lisp/x/profiler/opentelemetry_annotator.go
```diff
@@ -75,41 +72,27 @@ func contextTracer(ctx context.Context) trace.Tracer {
...
+	oldContext := p.currentContext
 	switch function.Type {
 	case lisp.LInt, lisp.LString, lisp.LFloat, lisp.LBytes, lisp.LError, lisp.LArray, lisp.LQuote, lisp.LNative, lisp.LQSymbol, lisp.LSortMap:
 		// We don't need to profile these types. We could, but we're not that LISP :D
-		return
+		return func() {}
 	case lisp.LFun, lisp.LSymbol, lisp.LSExpr:
 		fName := fmt.Sprintf("%s:%s", function.FunData().Package, getFunNameFromFID(p.runtime, function.FunData().FID))
-		p.contexts.Push(p.currentContext)
 		p.currentContext, p.currentSpan = contextTracer(p.currentContext).Start(p.currentContext, fName)
```
jeremy-asher, line 86: Would it be safer to just keep the context and span as locals that the returned function uses to avoid any kind of race?

sam-at-luther: i believe they are still needed so that on the recursive calls it does the right nesting, but maybe i'm not understanding the suggestion

jeremy-asher (approved):

PR #62, opened 2024-01-04
Add profiler span labeler

sam-at-luther:


lisp/x/profiler/funlabeler.go
```diff
@@ -0,0 +1,85 @@
...
+
+	// Eliminate duplicate underscores
+	parts := strings.Split(userLabel, "_")
+	var nonEmptyParts []string
+	for _, part := range parts {
+		if part != "" {
+			nonEmptyParts = append(nonEmptyParts, part)
+		}
+	}
+	return strings.Join(nonEmptyParts, "_")
```
jeremy-asher: This prefixing behavior seems a bit non-intuitive.  Would it make sense to just drop leading junk we don't want to allow or reject the label entirely and fall back to the function name?

jeremy-asher (approved):

PR #64, opened 2024-02-09
Tracing fixes

jeremy-asher:


lisp/x/profiler/callgrind.go
```diff
@@ -195,11 +197,15 @@ func (p *callgrindProfiler) end(fun *lisp.LVal) {
...
-		ref.line = fun.Source.Line
-		ref.file = fun.Source.File
+	if ref.line == 0 && loc != nil {
+		ref.line = loc.Line
+		ref.file = loc.File
 	}
 	// Output timing and line ref
+	line := 0
+	if loc != nil {
+		line = loc.Line
```
sam-at-luther, line 207: is it OK that `ref.line` may not be the same as this line here?

jeremy-asher: I'm pretty sure this should work as it did before unless I missed something.

sam-at-luther (approved):

PR #87, opened 2026-02-12
Fix handler-bind condition types reported as undefined

### LINKED ISSUE #85: Semantic analysis: handler-bind condition types and #^ prefix lambda % reported as undefined
## Summary

~~Two categories of~~ **One category** of false `undefined-symbol` diagnostics from semantic analysis:

### 1. handler-bind condition types ✅ Fixed in 8351fd3

```lisp
; repro_handler.lisp
(defun safe-divide (a b)
  (handler-bind
    ((condition (lambda (c) (debug-print "caught" c) 0)))
    (/ a b)))
```

```bash
elps lint --workspace=. repro_handler.lisp
```

**Expected:** No diagnostics
**Actual:** `undefined symbol: condition`

The `analyzeHandlerBind` implementation walks all children including the condition type binding pairs `(condition handler-fn)`. The first element of each pair is a condition type name (matched at runtime), not a variable reference. The analyzer should skip type positions in handler-bind bindings.

### ~~2. #^ prefix lambda `%` parameter~~ Not reproducible

Could not reproduce with standalone `elps lint` at v1.17.0. The analysis correctly handles `#^(* % 2)`. See [comment](https://github.com/luthersystems/elps/issues/85#issuecomment-3888849327).

## Impact

These cause ~15 false `undefined-symbol` diagnostics in substrate shirocore (handler-bind condition types only).

sam-at-luther:
## Summary

Fixes #85 (handler-bind part).

- `analyzeHandlerBind` was walking all children including condition type names in binding pairs like `(condition handler-fn)`. The condition type is a runtime matcher, not a variable reference — it should be skipped.
- Split analysis to skip the first element of each binding pair and only analyze handler functions + body forms.
- The prefix lambda `%` issue from #85 could not be reproduced — the analysis correctly handles `#^(* % 2)` and the lint test passes without changes.

## Test plan

- [x] `TestAnalyze_HandlerBind_ConditionTypeNotUndefined` — verifies condition type not flagged
- [x] `TestUndefinedSymbol_Negative_HandlerBindConditionType` — lint integration test
- [x] Existing `TestAnalyze_HandlerBind` (lambda param `c` resolved) still passes
- [x] `make test` — full suite passes
- [x] `make static-checks` — golangci-lint clean
- [x] `./elps lint --workspace=. ./...` — no false positives

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/analysis/analysis_test.go
+++ b/analysis/analysis_test.go
@@ -340,6 +340,20 @@ func TestAnalyze_HandlerBind(t *testing.T) {
 	}
 }
 
+func TestAnalyze_HandlerBind_ConditionTypeNotUndefined(t *testing.T) {
+	// The condition type name in handler-bind bindings is a runtime matcher,
+	// not a variable reference. It should not be flagged as undefined.
+	result := parseAndAnalyze(t, `
+(defun safe-divide (a b)
+  (handler-bind
+    ((condition (lambda (c) (debug-print "caught" c) 0)))
+    (/ a b)))`)
+	for _, u := range result.Unresolved {
+		assert.NotEqual(t, "condition", u.Name,
+			"condition type name in handler-bind should not be flagged as undefined")
+	}
+}
+
 // --- Analyze: test-let / test-let* ---
 
 func TestAnalyze_TestLet_BindingsResolved(t *testing.T) {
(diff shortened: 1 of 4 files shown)


sam-at-luther:
Both fixes from #85 are now on main: handler-bind condition types (041653c) and unqualified (expr ...) prefix lambda (158692b).

PR #253, opened 2026-03-27
feat: load-tree package inference, keyword hover docs, EndPos everywhere

### LINKED ISSUE #250: LSP: cross-file use-package not modeled — stdlib macros like 'when' reported as undefined
## Description

When opening a phylum codebase with `shirotester lsp`, files that rely on `(use-package 'utils)` from another file (typically `main.lisp`) get false `undefined-symbol` errors for symbols like `when`, `unless`, etc.

## Reproduction

Given a phylum project with two files:

**main.lisp:**
```lisp
(in-package 'my-service)
(use-package 'utils)
(use-package 'router)
```

**test-helpers.lisp:**
```lisp
(in-package 'my-service)

(defmacro assert-error (err msg &rest expr)
  (progn
    (when (nil? err) ...)))  ;; LSP reports: undefined symbol: when
```

The `when` macro is defined in substrate's shirocore `utils` package and is available via the registry. However, since `test-helpers.lisp` doesn't have its own `(use-package 'utils)`, the per-file analyzer cannot resolve `when`.

## Root Cause

The LSP's per-file analysis model doesn't propagate `use-package` effects across files:

1. `bootShiroRegistry()` loads the full shiro environment including `utils:when`
2. `ExtractPackageExports()` correctly extracts `when` into `PackageExports["utils"]`
3. Per-file analysis only imports `utils` symbols if the file itself contains `(use-package 'utils)`
4. At runtime, `use-package` in `main.lisp` affects all subsequently loaded files — the LSP doesn't model this

## Impact

This is a significant UX issue for substrate phylum development because:
- It's standard practice to have `use-package` declarations in `main.lisp` only
- Other files implicitly rely on the runtime's package import state
- Common macros like `when`, `unless`, `cond-let`, etc. show as undefined in every non-main file

## Possible Solutions

1. **Workspace-level use-package tracking**: During workspace prescan, collect all `use-package` declarations and apply them to files in the same `in-package` scope
2. **Auto-import common packages**: When the registry provides a package, auto-import its exports into all files that share the same `in-package`
3. **Per-file use-package hint**: The LSP could suggest adding `(use-package 'utils)` as a quick-fix code action
4. **Configuration**: Allow specifying default-imported packages in LSP settings

### LINKED ISSUE #251: LSP: 'view problem' navigates to wrong source location for shadowing diagnostics
## Description

When the LSP reports a shadowing diagnostic (e.g., `parameter 'expr' shadows special-op from enclosing scope`) on a defmacro parameter, clicking "View Problem" in VS Code navigates to an incorrect location — reportedly to a `when` definition further down in the file rather than to the `expr` parameter.

## Reproduction

In a file like:

```lisp
(defmacro assert-not-error (&rest expr)   ;; shadowing diagnostic reported here
  ...)

;; ... more code ...

(defmacro when ...)   ;; VS Code navigates HERE when clicking "view problem"
```

1. Open the file in VS Code with the ELPS LSP (via `shirotester lsp`)
2. See the shadowing warning on the `expr` parameter
3. Click "View Problem" or click the diagnostic in the Problems panel
4. VS Code navigates to the wrong location

## Analysis

The diagnostic position flows through:

1. `addParams()` (`analysis/analyzer.go:1092-1110`) — sets `sym.Source` by matching the cell name in the formals list
2. `AnalyzerShadowing` (`lint/analyzers.go:776-779`) — uses `posFromSource(sym.Source)` to set `Pos`
3. `convertLintDiagnostic()` (`lsp/diagnostics.go:203-230`) — converts to LSP Range with 0-based line/col

Potential causes:
- The `Source` location on the `expr` cell inside `(&rest expr)` might not be set correctly by the parser
- The `posFromSource` conversion might produce an incorrect position
- The diagnostic might lack an `EndPos`, causing VS Code to expand the range unexpectedly
- There could be a mismatch between the diagnostic the user clicks and the one that navigates

## Steps to Debug

1. Add a test case: `(defmacro foo (&rest expr) expr)` — verify the shadowing diagnostic's `Pos` points to the `expr` token in the formals
2. Check if `Source` is nil for symbols after `&rest` in the formals list (the `addParams` loop searches `formalsVal.Cells` for a cell matching `p.Name`)
3. Verify `convertLintDiagnostic` output for this case

This may be reproducible with a minimal test case in `lint/lint_test.go` by checking the `Pos.Line` and `Pos.Col` of the shadowing diagnostic.

sam-at-luther:
## Summary

Builds on PR #252 to address remaining issues from #250 and #251:

- **Load-tree package inference**: Walk `main.lisp`'s `load-file` chain to build a file→package map. Bare files inherit the package context from the load point.
- **Pre-in-package def remapping**: Defs before first `in-package` are remapped to load-tree package via `firstInPkgLine` tracking.
- **Cross-file unused-function detection**: Wire `WorkspaceRefs` through `Config → Result → lint Pass`. Checks cross-file references before flagging.
- **Macro-body undefined-symbol → warning**: Unresolved symbols inside user-defined macro bodies are warnings (not errors) since macros may introduce bindings at expansion time. Built-in macros still produce errors.
- **DiagnosticTag.Unnecessary**: Unused-variable/function diagnostics include `DiagnosticTag.Unnecessary` so VS Code renders them with faded text, visually distinguishing them from errors.
- **Keyword hover docs**: `&rest`, `&optional`, `&key`, `unquote`, `unquote-splicing`, `condition`, `else`.
- **EndPos on all diagnostics**: 15 diagnostics across 12 analyzers.
- **`lisp.DefaultUserPackage` constant**: Replace all hardcoded `"user"` strings. Document default package in `docs/lang.md`.

Closes #250
Closes #251

## Test plan

- [x] Load tree: `TestBuildLoadTree_Basic/PackageSwitch/Nested/CycleDetection`
- [x] Package inference: `TestPrescanWorkspace_DefaultPackage/ExplicitUserPackageNotRemapped/NoMainLisp/DefsBeforeInPackage/PhylumPattern`
- [x] Cross-file resolution: `TestAnalyze_DefaultPackage_BareFile`, `TestAnalyze_DefsBeforeInPackage_Resolve`
- [x] Unused-function: `TestUnusedFunction_Negative_CrossFileRef` + `_WithDefaultPackage` + `_Positive_OnlySameFileRef` + `_NoWorkspaceRefs`
- [x] Macro-body severity: `TestUndefinedSymbol_Warning_InsideMacroBody`, `_Error_InsideBuiltinMacro`, `_Error_InsideFunctionCall`
- [x] Hover: `TestHover_AllKeywords` (7 subtests) + `TestWordAtPosition/ampersand_keywords`
- [x] EndPos: `TestUndefinedSymbol_HasEndPos` + `TestUserArity_HasEndPos` (exact positions)
- [x] Full `go test ./...` passes

### DIFF
--- a/analysis/analysis.go
+++ b/analysis/analysis.go
@@ -38,6 +38,11 @@ type Config struct {
 	// (no in-package declaration). Derived from main.lisp's in-package.
 	DefaultPackage string
 
+	// WorkspaceRefs maps SymbolKey.String() to cross-file references.
+	// When set, analyzers can check whether a symbol is referenced from
+	// other files in the workspace.
+	WorkspaceRefs map[string][]FileReference
+
 	// Filename is the source file being analyzed.
 	Filename string
 }
@@ -73,6 +78,11 @@ type Result struct {
 	// for cross-file duplicates without relying on scope lookups that may
 	// have been overwritten by local definitions.
 	ExtraGlobals []ExternalSymbol
+
+	// WorkspaceRefs maps SymbolKey.String() to cross-file references.
+	// Copied from Config. Used by lint analyzers to check whether a symbol
+	// is referenced from other workspace files (e.g. unused-function check).
+	WorkspaceRefs map[string][]FileReference
 }
 
 // Analyze performs semantic analysis on a set of parsed expressions.
@@ -106,7 +116,7 @@ func Analyze(exprs []*lisp.LVal, cfg *Config) *Result {
 
 	a := &analyzer{
 		root:             root,
-		result:           &Result{RootScope: root, ExtraGlobals: cfg.ExtraGlobals},
+		result:           &Result{RootScope: root, ExtraGlobals: cfg.ExtraGlobals, WorkspaceRefs: cfg.WorkspaceRefs},
 		cfg:              cfg,
 		qualifiedSymbols: make(map[string]*Symbol),
 	}
--- a/analysis/analysis.go
+++ b/analysis/analysis.go
@@ -34,6 +34,10 @@ type Config struct {
 	// imported in other files (e.g. main.lisp) are available.
 	PackageImports map[string][]string
 
+	// DefaultPackage overrides the default "user" package for bare files
+	// (no in-package declaration). Derived from main.lisp's in-package.
+	DefaultPackage string
+
 	// Filename is the source file being analyzed.
 	Filename string
 }
@@ -111,7 +115,7 @@ func Analyze(exprs []*lisp.LVal, cfg *Config) *Result {
 	a.prescan(exprs, root)
 
 	// Phase 2: Deep recursive walk
-	currentPkg := "user"
+	currentPkg := a.defaultPackage()
 	for _, expr := range exprs {
 		a.analyzeExpr(expr, root, currentPkg)
 		if expr != nil && expr.Type == lisp.LSExpr && !expr.Quoted && len(expr.Cells) > 0 &&
--- a/analysis/analysis_test.go
+++ b/analysis/analysis_test.go
@@ -1561,6 +1561,28 @@ func TestAnalyzeFile_PreservesPackageImports(t *testing.T) {
 		"AnalyzeFile should forward PackageImports so 'when' resolves")
 }
 
+func TestAnalyze_DefaultPackage_BareFile(t *testing.T) {
+	// Regression test: a bare file (no in-package) with DefaultPackage set
+	// should use the default package, enabling cross-file imports to work.
+	source := "(defun f () (helper-fn 42))"
+	cfg := &Config{
+		PackageExports: map[string][]ExternalSymbol{
+			"helpers": {{Name: "helper-fn", Kind: SymFunction, Package: "helpers"}},
+		},
+		PackageImports: map[string][]string{
+			"svc": {"helpers"},
+		},
+		DefaultPackage: "svc",
+	}
+	result := parseAndAnalyzeWithConfig(t, source, cfg)
+	unresolvedNames := make(map[string]bool)
+	for _, u := range result.Unresolved {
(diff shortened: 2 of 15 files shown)


sam-at-luther:
## Testing feedback from acre phylum

The load-tree inference is a big improvement — most of the false positives are gone. But there are a few remaining edge cases:

### 1. Multi-package files with definitions before first `in-package`

`utils.lisp` has this structure:
```lisp
;; No in-package here — inherits 'acre from main.lisp load context

(defun get-created-index (obj) ...)     ;; defined in acre package at runtime
(defun strip-created-index! (obj) ...)  ;; defined in acre package at runtime

(in-package 'utils)                     ;; switches to utils
(set 'index-bimap-prefix "acr:bimap")
(in-package 'acre)                      ;; switches back
```

`scanFilePackage()` returns `"utils"` (the first `in-package` found), so the file is NOT treated as bare. But `extractDefinitions` starts with `currentPkg = "user"` and registers `get-created-index` and `strip-created-index!` under `"user"` (before hitting the `in-package 'utils` on line 19).

At runtime, these definitions are in `acre` (inherited from the `main.lisp` load context). The load-tree knows this file should be `acre`, but the `filePkg != ""` check prevents remapping.

**Possible fix:** For multi-package files in the load tree, remap definitions that were registered under `"user"` (before any `in-package`) to the load-tree package, even if the file has a later `in-package`.

### 2. Cross-file "unused function" false positives

`caller-has-acre-admin-api-key?` is defined in `auth.lisp` and called in `permission.lisp`. Both are bare files correctly remapped to `acre`. The LSP reports it as "unused function" because the unused-function analyzer only checks usages within the same file, not across files.

This is a pre-existing limitation (not caused by this PR), but it's more visible now that the undefined-symbol false positives are fixed.

Issue #105, opened 2026-02-13
Fix sorted-map Copy() and Equal() (HC-7)


sam-at-luther:
## Parent issue

Sub-ticket of #104 (HC-7).

### Copy() shares underlying map data

`LVal.Copy()` for `LSortMap` does a shallow struct copy (`*cp = *v`), which shares the `*MapData` pointer. Mutations via `assoc!`/`dissoc!` on the copy affect the original.

`copyMapData()` already exists and correctly creates a structurally independent map with shared value pointers — it's used by `assoc` and `dissoc`. It just needs to be wired into `Copy()`.

### Equal() unconditionally returns false for all sorted-maps

`Equal()` checks map lengths, then unconditionally returns `Bool(false)` — even for empty maps. The code has an explicit `BUG` comment.

### Copy()

Call `copyMapData()` for `LSortMap` in `Copy()`. This copies the map structure (new Go maps) while sharing value `*LVal` pointers — consistent with how `copyCells()` works for lists, and already proven performant via `assoc`/`dissoc`.

### Equal()

Iterate entries from both maps and recursively compare key-value pairs using `Equal()`. Short-circuit on first mismatch.

## PoC tests

Failing tests demonstrating both bugs are already written in `lisp/map_test.go`:
- `TestSortedMapCopyBug` — Go-level: mutating copy mutates original
- `TestSortedMapCopyDissocBug` — Go-level: deleting from copy deletes from original
- `TestSortedMapCopyViaSort` — ELPS-level: stable-sort comparison function mutation bleeds through
- `TestSortedMapEqualBug` — Go-level: identical maps not equal
- `TestSortedMapEqualELPS` — ELPS-level: `equal?` returns false for identical maps

## Locations

- `lisp/lisp.go:979-992` — `Copy()`
- `lisp/lisp.go:994-1007` — `copyMapData()` (already exists)
- `lisp/lisp.go:950-955` — `Equal()`

PR #308, opened 2026-08-06
Fix evaluation-limit design, sorted-map equality, panic handling, and lint false positives

sam-at-luther:
Audit follow-up on 343 AI-authored commits since 2025-02-01. Nine findings addressed, one dropped as incorrect, one confirmed and deferred. Also folds in the three open Dependabot PRs so this is one review and one merge.

**Supersedes #304, #305 and #306** — their heads are cherry-picked here as the first three commits.

## Behaviour changes to defaults

These are the risky part of the PR. Each one, with who is affected:

| # | Old | New | Who is affected |
|---|---|---|---|
| 1 | `MaxHeightLogical` defaulted to 50000, which capped every tail loop at ~25,000 iterations | `MaxHeightLogical` defaults to 0 (off); new `MaxTailIterations` defaults to 1,000,000 turns | Anyone whose tail loop ran past ~25k iterations was **broken** and is now fixed. Anyone relying on the logical bound must now set it explicitly. Runaway loops are still bounded, at 1M turns instead of ~25k. |
| 1 | `MaxHeightPhysical` = 25000 | **unchanged** | nobody |
| 3 | A recovered Go panic was an ordinary `error` condition, swallowed by `ignore-errors` and catch-all `handler-bind` | Tagged `internal-panic`; propagates unless a handler names it | Code that wrapped a panicking host builtin in `ignore-errors` and carried on now sees the error surface. That is the point — but it can turn a silently-degraded path into a visible failure. |
| 4 | `WithMaxSteps(n)` was a lifetime quota; once `n` steps had run in total, every later evaluation failed | Per-evaluation budget, reset at the outermost entry point | Anyone using `WithMaxSteps` on a long-lived `Runtime`. The old behaviour matched neither doc comment. |
| 2 | `(equal? (sorted-map "a" 1) (sorted-map 'a 1))` was `false` | `true` | Code depending on key *spelling* affecting equality. That branch is 3 months old (#105); before it, `equal?` returned `false` for **all** sorted-maps. Applies to string/symbol keys only — custom `lisp.Map` implementations with other key types keep full `Equal` semantics (review F1). |
| 5,6,8 | Three lint checks reported findings on correct code, two at `SeverityError` (exit 1) | No longer reported | Anyone whose `elps lint` was red or noisy. No check was weakened for real defects; see the tests. |
| 9 | `elps run -e` / `-p` were dead flags | Implemented as documented | `elps run -e '(+ 1 2)'` used to fail with "no such file or directory". |

## Review round (adversarial review of `b69a70c`)

Eight review findings (F1–F8); all addressed. Head is now `9d441a6`.

### F1 [HIGH] — REQUIRED, fixed. Confirmed regression vs main.

`equal?` compared sorted-map keys with `.Str` **unconditionally**. Safe for the stock `sortedmap` (string/symbol keys only) — but `lisp.Map` is an exported interface and `lisp.SortedMapFromData` an exported extension point, and `libjson` already uses that path in-tree. For any key type whose `.Str` is empty (`LInt`, `LFloat`, `LSExpr`, …) **every key compared equal**. Reproduced exactly as reported:

```
custom Map with integer keys
a = {1:"x", 2:"y"}   b = {7:"x", 9:"y"}
keys(a)='(1 2)  keys(b)='(7 9)
before fix: (equal? a b) = true    <- silent wrong answer
after  fix: (equal? a b) = false
```

Fixed by applying the name shortcut only when **both** keys are string-like, falling back to `Equal` otherwise. The intent — spelling is not part of key identity — is unchanged for the keys it was reasoned about, and is deliberately not extended past them. `TestSortedMapEqualNonStringKeys` adds a custom integer-keyed `lisp.Map`.

### F2 [MEDIUM] — fixed.

`collectBindingFormNames` dumped every locally bound name into a file-global set, so an unrelated `(let ([map ...]) ...)` in one function silently disabled `builtin-arity` for `map` across the whole file. Reproduced:

```lisp
(map 'list)                          ; genuine arity error
(defun unrelated ()
  (let ([map (sorted-map 'a 1)])     ; unrelated binding, different function
    (get map 'a)))
```

Shadowing is now scoped to the binding form's own subtree, so the diagnostic is reported again. The node-level `aritySkipNodes` half — which is what actually fixes the originally reported false positive — is unchanged. Pinned by `TestBuiltinArity_ShadowingIsScopedToTheBindingForm`, which also asserts the in-scope shadowed call is still exempt.

The reviewer's exposure datum is worth recording: in substrate's shirocore corpus, 8 of 20 `.lisp` files bind a name colliding with a builtin (`key` ×11, `min` ×2, `get` ×2, `max`, `map`). No error was masked today, but the check gates the build.

### F3 [MEDIUM-LOW] — fixed.

`MacroCall` was the only exported evaluation entry point without `beginEval()`. Reproduced: after a budget-exhausting `Load`, `MacroCall` still returned `step-limit-exceeded` while a control `FunCall` correctly reset. So a host driving macro expansion through it — formatter, expander, LSP feature — kept exactly the lifetime-quota poisoning this PR set out to remove. One-line fix, plus the entry-point lists in `config.go` and `docs/lang.md`. `TestMaxStepsResetAtEveryExportedEntryPoint` now covers `MacroCall`, `FunCall` and `Load`.

### F4 [NIT] — fixed with a non-forgeable marker rather than a doc note.

`(error 'internal-panic "...")` written in lisp escaped both `ignore-errors` and a catch-all `handler-bind`, so any library could raise an uncontainable error, and a real host panic was indistinguishable from a forged one.

Containment now keys off the Go stack snapshot the recover handler attaches, not the condition name. That marker is non-forgeable: `env.go:957` is the only assignment to `GoStack` in the tree and it writes to the error's `CallStack` **copy**, so the live runtime stack's `GoStack` is always nil and an `error`-raised condition always copies nil. Exported as `lisp.IsInternalPanic(v)` so embedders test the same way. A forged `internal-panic` is now an ordinary containable condition; a real one still escapes.

### F6 [NIT] — fixed, and this one changed the docs materially.

The tail-iteration backstop is **not** a wall-clock bound and the "~4s" figure only holds for an O(1) body. Verified locally:

```lisp
(defun grow (n acc) (if (= n 0) (length acc) (grow (- n 1) (cons n acc))))
(grow 60000 ())   ; 60,000 turns — far under the 1M backstop — but ~17s
```

`WithMaxSteps` does not help either: an O(n) builtin is one step. `docs/lang.md` and the `DefaultMaxTailIterations` doc comment now qualify the figure as trivial-body and point at **a context deadline** as the only limit here that measures elapsed time.

### F7 — boundary pinned.

Of the reviewer's 5 surviving mutations, the one worth pinning was the `<` vs `<=` off-by-one. Measured boundary: with `MaxTailIterations = 1000`, `(spin 1000)` succeeds and `(spin 1001)` fails reporting `1001` — a limit of N permits N turns and the error reports the turn that exceeded. `TestTailIterationLimitBoundary` now pins N-1 / N / N+1 and the reported number.

The other four survivors I agree should not be chased: `specialOpCall`'s `TailIterations++` sits on a branch the reviewer instrumented and found **unreachable** (0 hits across the whole suite), so it is unpinnable by construction — it stays for symmetry with its sibling site rather than as covered code; `CallStack.Copy()` carrying the two limits is a real latent correctness fix but currently unobservable (the copy is only attached to errors for trace rendering); and `CheckTailCall`'s `CheckHeight()` call is genuinely redundant with `PushFID` → `checkHeightPush`, which is defensible defence-in-depth on a cheap path.

### F8 [NIT] — fixed, slightly beyond "lines touched".

Verified every pinned action SHA against upstream tags with `git ls-remote --tags`. Four confirmed mislabels:

| pin | comment said | actually |
|---|---|---|
| `actions/checkout@3d3c42e5` | `# v6` | **v7** (v7.0.1) |
| `actions/setup-go@b7ad1dad` | `# v6` | **v7** (v7.0.0) |
| `actions/cache@55cc8345` | `# v5` | **v6** (v6.1.0) |
| `actions/setup-node@82076278` | `# v6` | **v7** (v7.0.0) |

(For reference: checkout v6 is `d23441a4`, v7.0.0 is `9c091bb2`.) The reviewer scoped this to "lines you touch", which is the `checkout` lines only. I fixed all four: leaving three confirmed-wrong labels beside a corrected one in the same file is the exact failure mode being flagged, and these are comment-only edits with no functional effect. Trivially droppable if you prefer the narrower scope.

Two pins I could **not** verify and did not touch: `actions/github-script@3a2844b7` (`# v9`) and `golangci/golangci-lint-action@ba0d7d2e` (`# v9`). Neither SHA is a tag tip, so `ls-remote` cannot confirm or refute the label.

**No pinned SHA was changed** — the diff is comments only.

### Mutation matrix, review round

| Mutation | Test(s) that failed |
|---|---|
| F1 restore unconditional `.Str` comparison | `TestSortedMapEqualNonStringKeys` |
| F1b drop the name shortcut entirely | `TestSortedMapEqualIgnoresKeySpelling`, `TestSortedMapEqualNonStringKeys` |
| F2 make the shadow set file-global again | `TestBuiltinArity_ShadowingIsScopedToTheBindingForm` |
| F3 remove `beginEval` from `MacroCall` | `TestMaxStepsResetAtEveryExportedEntryPoint` |
| F4 match `internal-panic` by name | `TestForgedInternalPanicIsContainable` |
| F7 flip the boundary `<` → `<=` | `TestTailIterationLimitBoundary` |

Re-verified after the review round on linux/amd64, go1.25.0, golangci-lint 2.6.2: `make test`, `make race`, `make test-elpscheck`, `make static-checks` (0 issues), `elps fmt`, `elps lint` (9 warnings, still byte-identical to `origin/main`), `elps doc -m`, `go vet`, `go build` — all green.

Also folded in from a parallel session: `b69a70c` pins the sticky symbol key spelling as accepted (cosmetic) behaviour, with the reasoning and the blast-radius trade recorded so it is not rediscovered as a bug.

### 1 — BLOCKER: tail loops hard-failed at ~25,000 iterations. Fixed.

Confirmed: `(spin 24000)` returned, `(spin 25000)` gave `logical stack height exceeded maximum: 50001`.

**I partly disagree with the framing in the brief, and the correction matters.** Commit `9e5fd6c74` was not a regression. Before it, `StandardRuntime()` built `&CallStack{}` with *both* limits zero, so there was no protection at all against unbounded non-tail recursion crashing the Go runtime. It fixed a real fatal bug; it just bundled a correct half (the physical guard) with an incorrect one (the logical guard used as an iteration cap).

Measured on linux/amd64, go1.25, with limits disabled:

| non-tail depth | time | Go stack | result |
|---|---|---|---|
| 25,000 | 0.67s | +85MB | ok |
| 100,000 | 2.68s | +340MB | ok |
| 200,000 | 4.06s | +678MB | ok |
| 400,000 | — | — | **uncatchable Go stack overflow, process dies** |

| tail iterations | time |
|---|---|
| 100,000 | 0.37s |
| 1,000,000 | 3.89s |
| 5,000,000 | 18.91s |

`MaxHeightPhysical = 25000` sits 8–16x below the crash threshold. That margin is load-bearing — the threshold moves with arch, Go build, frame size and container memory ceiling, so it was measured on one machine and must not be read as a portable constant. **It stays at 25000.**

The two limits are genuinely orthogonal: a 200,000-turn tail loop runs to completion under a **100-frame** physical ceiling, while non-tail recursion is refused at frame 101. Only non-tail recursion moves the physical stack, so raising the tail-loop bound carries zero stack-crash risk. Pinned in `TestTailCallsDoNotConsumePhysicalFrames`.

**On the unit.** `HeightLogical` counts *elided frames*, not turns: a trivial body elides 2 per turn, a body with one more level of terminal nesting elides 3. So the same numeric bound buys a different number of iterations depending on the shape of the loop — exactly the property that made the original bound fail data-dependently. Rather than ship a constant whose unit an operator cannot reason about, I took the second option offered: **count real tail iterations**.

- `CallFrame.TailIterations` increments by exactly one per turn, whatever the body looks like.
- `MaxTailIterations` (default 1,000,000) is the runaway-loop backstop, in turns.
- `MaxHeightLogical` keeps its documented "elided frames" meaning and is **off by default**, opt-in via `WithMaximumLogicalStackHeight`.

`MaxHeightLogical` is left off rather than raised to ~2,000,000 deliberately: with both bounds on, the logical one fires first for any body that nests terminal forms more deeply than the trivial case, reintroducing the same shape-dependent failure with the same confusing message. One bound, in honest units, is the whole point.

Both messages now name the limit and the knob:

```
tail-call iteration limit exceeded: 1000001 (a single tail-recursive loop ran this
many turns; this is a runaway-loop backstop, not a stack overflow — raise or
disable the limit with WithMaxTailIterations)

physical stack height exceeded maximum: 25001 (recursion too deep; the limit stops
the Go runtime from aborting the process with an unrecoverable stack overflow —
rewrite the recursion as a tail call, or raise the limit with
WithMaximumPhysicalStackHeight)
```

### 2 — `equal?` on sorted-maps. Verdict: **legitimate bug**. Fixed.

Investigated as asked. Evidence:

- **The behaviour contradicts documented key identity.** `docs/lang.md:526` already said sorted-map lookup works "with either a string or a symbol, regardless which type was used to insert/set the value originally". `get`, `key?`, `assoc` and `dissoc` all honour that. `Equal` compared the reconstructed key `LVal`s with `LVal.Equal`, which is type-sensitive, so it disagreed with every one of those accessors.
- **Nothing documents or asserts the divergence.** No doc comment, no `docs/lang.md` text, no test. The `equal?` docstring said only "structurally equal, deep comparison across all value types".
- **It is new and incidental.** The `LSortMap` branch of `Equal` was added 2026-02-13 in #105 (AI-authored). Before that, `equal?` returned `false` for *all* sorted-maps including empty ones. The commit message describes fixing "always false" and says nothing about key types — the type sensitivity is a side effect of reusing `LVal.Equal` on reconstructed keys, not a decision.
- **There is hidden state, and it is the leak rather than a justification.** `sortedmap` carries a `typemap` recording each key's spelling. `Set` with an `LString` key does **not** clear a `symbolkey` entry recorded earlier, so the spelling is sticky and reflects the *first* insertion, not the last. Equality therefore depended on construction history:

```lisp
(set 'm1 (sorted-map 'a 0)) (assoc! m1 "a" 3)   ; m1 prints (sorted-map 'a 3)
(set 'm2 (sorted-map "a" 3))
(equal? m1 m2)                                   ; was false
;; but delete-then-reinsert cleared the typemap, and then:
(set 'm3 (sorted-map 'a 0)) (dissoc! m3 'a) (assoc! m3 "a" 3)
(equal? m3 m2)                                   ; was true
```

Two maps with identical contents, reached by two documented routes, compared differently.

- **"Identical through every accessor" is not quite right, and I checked.** `keys` and the printed form *do* distinguish the two spellings, and that is long-standing intentional behaviour (present in the initial import, asserted by `TestMaps`). So key spelling is real, deliberate *presentation*. What was wrong is that it leaked into *identity*, where the language had already committed to name-based lookup.

Fix: compare key names, for string-like keys only (see review F1). `keys`, printing and the sticky typemap are untouched. Documented in `docs/lang.md`, the `equal?` docstring, and `TestSortedMapEqualIgnoresKeySpelling`.

### 3 — Go panics in host builtins silently swallowed. Fixed.

`eval`'s `recover()` produced `env.Errorf(...)`, i.e. the generic `"error"` condition — indistinguishable from `(error 'my-condition "...")`. `ignore-errors` returned `()` and a catch-all `handler-bind` intercepted it, so a nil dereference in a host builtin looked like a routine failure and the program continued over data structures the panicking code had left in an unknown state.

Now tagged `CondInternalPanic = "internal-panic"`. `ignore-errors` and the catch-all `condition` specifier do not intercept it; an explicit `(internal-panic ...)` handler does. Ordinary errors are unaffected. Containment keys off a non-forgeable marker — see review F4.

### 4 — `WithMaxSteps` cumulative, never reset. Fixed.

`Runtime.maxSteps` was documented "Per-evaluation step limit" and `WithMaxSteps` as "the maximum number of evaluation steps before evaluation returns"; `steps` only ever reset via an explicit `ResetSteps()`. So `WithMaxSteps(n)` was a **lifetime quota**: once a long-lived runtime had executed `n` steps in total, every subsequent evaluation failed however small.

Reset at the outermost evaluation entry (`Eval`, `EvalContext`, `EvalSExpr`, `FunCall`, `FunCallContext`, `SpecialOpCall`, `MacroCall`, and `load`, which funnels every `Load*`). Nested evaluation shares the enclosing budget — pinned by `TestMaxStepsNotRefilledByNestedEvaluation`, since refilling per nested entry would let a runaway loop evade the limit forever. Added `Runtime.TotalSteps()` for the lifetime count.

This matters for finding 1: a step budget is the only instrument here that bounds a loop which neither recurses nor tail-calls. (Base elps has no `while` — that is substrate's shirocore — so in base elps the loop shapes are recursive and the stack limits do cover them. In substrate they do not.) It is not a *time* bound either; see review F6.

### 5 — `builtin-arity` false positives on binding forms. Fixed.

`(let ((map (sorted-map 'a 1))) ...)` contains the s-expression `(map (sorted-map 'a 1))`, whose head names the builtin `map` (min arity 3). The walk read it as a call and reported `map requires at least 3 argument(s), got 1` at `SeverityError`, so `elps lint` exited 1 on correct code. Reproduced for `let`, `let*`, `flet`, `labels`, and for call sites of locally shadowing functions:

```
error: map requires at least 3 argument(s), got 1 (builtin-arity)
error: rest accepts at most 1 argument(s), got 2 (builtin-arity)
error: first accepts at most 1 argument(s), got 2 (builtin-arity)
error: nth requires at least 2 argument(s), got 1 (builtin-arity)
error: get requires at least 2 argument(s), got 1 (builtin-arity)   ; (flet ((get (x) x)) (get 1))
```

Two halves, both pinned separately: binding entries and their formals lists are excluded from arity checking, and names bound by `let`/`let*`/`flet`/`labels`/`macrolet` shadow the builtin **within the binding form's subtree** (see review F2 — the first version of this was file-global). Genuine arity errors still fire (`TestBuiltinArity_Positive_StillCatchesRealErrors`).

### 6 — `quote-call` flagged every correct `defconst`. Fixed.

`defconst` is a macro that quotes its own name argument — `(defconst x 42)` expands to `(progn (set 'x 42) (export 'x) ())` — so the unquoted form is correct. The check flagged all of them, and its suggested fix breaks the program:

```
$ elps run -e "(defconst 'my-const 42)"
error: lisp:set: first argument is not a symbol: quote
```

`defconst` removed from the check. `set` is unchanged; `set!` was already excluded. The runtime behaviour behind this is pinned in `TestDefconstQuotedNameIsARuntimeError`, so the rationale cannot rot.

### 7 — `undefined-symbol` severity. **Dropped — the finding is wrong.**

Severity does not depend on whether the macro is in the same file. A macro defined in another workspace file gets exactly the same `SeverityWarning` downgrade as one in the same file — both go through `insideMacroCall`:

```
$ elps lint --workspace ws --checks=undefined-symbol ws/other.lisp ws/same.lisp
warning: undefined symbol: thing            # macro in macros.lisp (cross-file)
warning: undefined symbol: undefined-here
warning: undefined symbol: thing            # macro in the same file
warning: undefined symbol: undefined-here2
```

The only divergence I could reproduce is between a macro that resolves *at all* and a head symbol that is entirely unresolved — and in that case the head is not known to be a macro (it could be an undefined function, in which case its arguments really are undefined), so reporting at error severity is defensible. There is a narrower related case — a macro registered from Go by an embedder has `Source == nil`, so `isUserMacro` is false and no downgrade applies — but that is "lisp macro vs Go macro", not "same file vs other file", and exercising it needs an embedder harness. Not shipping a speculative fix for a MINOR finding.

### 8 — `unused-nolint` false positives under `--checks`. Fixed.

```
$ elps lint --checks=set-usage file.lisp
warning: nolint directive does not suppress any diagnostic (unused-nolint)
  --> file.lisp:1:24     ; the directive is ; nolint:quote-call
```

`quote-call` never ran, so its directive could not possibly have suppressed anything. On a narrowed run a directive is now only reported if at least one analyzer it names actually ran; a bare `; nolint` is never reported. Unknown analyzer names are still reported — a typo is a typo regardless of which checks ran. **Zero change on a full run** (the exemption only activates when some default analyzer is absent), pinned by `TestUnusedNolint_FullRunStillFlagsStale`.

### 9 — `elps run -e` / `-p` dead flags. Fixed.

Both were declared, registered with cobra, and documented in the command's long help with worked examples, but never read: `elps run -e '(+ 1 2)'` treated the expression as a filename. Implemented as documented. Extracted `runElps` so it is testable (the previous body called `os.Exit` inline).

### 10 — INFO: mcpserver test cannot pass as root. Fixed.

`TestWorkspaceTraversalErrorsAreLoggedAndSkipped` chmods a directory to 0 and expects traversal to fail. Root bypasses that, so the "unreadable path" log never happens. Skipped under `os.Geteuid() == 0`. Common in CI and dev containers — `make race` fails on `origin/main` for this reason, and this PR fixes it.

### 11 — INFO: JSON integers above 2^53. **Confirmed, not fixed — out of scope as instructed.**

```lisp
(json:load-string "9007199254740993")        ; => 9.007199254740992e+15
(json:load-string "{\"id\": 9007199254740993}") ; => (sorted-map "id" 9.007199254740992e+15)
```

Integers decode through float64 and lose precision above 2^53, silently changing type from int to float. This is pre-existing (not AI-authored) and is the root cause of a wider class, so it needs its own decision about the intended contract — follow-up issue, not this PR.

## Drive-by

`TestEngine_RequestPause` had an ordering race: `WaitIfPaused` publishes `pausedEnv` (what `IsPaused()` reads) and releases the mutex *before* invoking the stopped-event callback, so the test could observe "paused" and then read an empty stop reason. It flaked under `-race` on a loaded 4-core box. Now waits for the event. Verified by 3/3 clean full `make race` runs after the fix, against repeated failures before it.

## Mutation matrix

Every fix was reverted individually and a named test confirmed to fail.

| Mutation | Test(s) that failed |
|---|---|
| M1 `DefaultMaxLogicalStackHeight` back to 50000 | `TestTailLoopIsNotBoundedByLogicalHeight`, `TestDefaultStackLimits` |
| M1b `DefaultMaxPhysicalStackHeight` to 0 | `TestDefaultStackLimits` |
| M1c `DefaultMaxTailIterations` to 0 | `TestDefaultStackLimits` |
| M1d drop the `TailIterations++` increment | `TestTailIterationLimit`, `TestTailIterationUnitIsOneTurnPerIteration` |
| M1e revert the logical-overflow message | `TestStack` |
| M2 restore type-sensitive map-key comparison | `TestSortedMapEqualIgnoresKeySpelling`, `TestSortedMapEqualELPS` |
| M3a revert the panic condition tag | `TestPanicHasInternalPanicCondition`, `TestIgnoreErrorsDoesNotSwallowPanic`, `TestHandlerBindCatchAllDoesNotSwallowPanic` |
| M3b let `ignore-errors` swallow panics again | `TestIgnoreErrorsDoesNotSwallowPanic` |
| M3c let catch-all `handler-bind` swallow panics again | `TestHandlerBindCatchAllDoesNotSwallowPanic` |
| M4 make `MaxSteps` cumulative again | `TestMaxStepsIsPerEvaluation`, `TestStepsAndTotalSteps` |
| M4b refill the budget on every nested evaluation | `TestMaxStepsNotRefilledByNestedEvaluation` |
| M5a revert the binding-form skip | `TestAritySkipNodes_BindingEntries` |
| M5b revert local-shadow name collection | `TestBuiltinArity_Negative_LocallyShadowedCalls` |
| M6 make `quote-call` flag `defconst` again | `TestQuoteCall_Negative_UnquotedDefconst` |
| M8 disable the narrowed-run exemption | `TestUnusedNolint_ChecksFilterDoesNotFlagKnown`, `TestUnusedNolint_ChecksFilterDoesNotFlagBare` |
| M9 revert `-e`/`-p` | `TestRunExpressionFlag`, `TestRunExpressionsShareOneEnvironment`, `TestRunPrintFlagWithFile` |

(Plus the six review-round mutations above.)

M5a and M8 each initially reported no failure — M5a because the two halves of the fix overlap, M8 because the naive mutation did not compile. Both were re-done: M5a is now pinned by a direct unit test on `aritySkipNodes`, and M8 by a mutation that keeps the build green. Finding 10 has no mutation (the fix *is* a skip); finding 7 was dropped; finding 11 was not fixed.

## Verification

Run locally on linux/amd64, go1.25.0, golangci-lint 2.6.2, **with the three dependency bumps applied**:

| Command | Result |
|---|---|
| `make test` (go test + `_examples`) | pass |
| `make race` | pass, 3 consecutive runs |
| `make static-checks` (golangci-lint v2.6) | 0 issues |
| `./elps lint --workspace=. --exclude 'grammar' --include '_examples' ./...` | 9 warnings, byte-identical to `origin/main` |
| `go vet ./...` | clean |
| `go build ./...` | clean |

Not verified: the Windows build job (no Windows host available), and the crash-threshold numbers are from one machine and one Go build.

## Follow-up, not in this PR

Substrate inherits these defaults silently — `elpsutil.go:130` calls `lisp.NewEnv(nil)` with no override, which is why the ~25k cliff went unnoticed until customer data grew past it. It should set `MaxHeightPhysical` and `MaxTailIterations` explicitly to match its transaction budget, so the domain-specific number lives in the repo that knows what a Fabric transaction costs. That is a substrate PR and is deliberately not touched here.

Also: finding 11 (JSON integers above 2^53) needs its own issue and its own decision.


sam-at-luther:
Superseded by #310, which consolidates this with #309 (the benchmark gate) into a single PR off latest `main`.

Everything here is carried across at `9d441a6`, including all eight adversarial-review fixes. Merged clean with #309 — no conflicts. Combined tree verified: `make test` 29 ok, `static-checks` 0 issues, `ci-gates-test` 45/45, `elps doc -m` clean, and `elps lint` at **17 diagnostics — identical to `main`**.

Closing so there is one thing to review; the branch is untouched.


---

PR #309, opened 2026-08-07
ci: make the benchmark regression gate able to fail

sam-at-luther:
## The bug

`.github/workflows/benchmark.yml:116`:

```bash
if grep -E '^\S.*\+$' benchstat-output.txt | grep -qv '^name'; then
  echo "has_regression=true" >> "$GITHUB_OUTPUT"
```

The regex requires a line **ending in a literal `+`**. No benchstat line ever does — a regression row ends in `(p=0.008 n=5)`, a geomean row ends in `%`. So `has_regression` was always `false` and the "Fail on regressions" step never executed.

```
$ grep -cE '^\S.*\+$' scripts/testdata/benchstat-task-sample.txt
0        # ...on a table containing +50.00% (p=0.000 n=10)
```

**Run-history evidence.** The workflow has **473 runs**. I sampled 143 via the Actions API (pages 1, 2, 5, 11, 16 — spanning 2026-02-08, the workflow's first day, through 2026-08-06): **137 success, 4 cancelled, 2 failure, 0 caused by the gate.** The two failures are both from 2026-02-08 on `feature/rethrow-errors`, the day the workflow was authored; their logs are past retention (`410 Gone`), but they cannot have been the gate — the same unmatchable regex was already present at those SHAs, so `has_regression` could not be `true` and the fail step could not run.

## The fix

Logic moves out of workflow YAML into `scripts/benchstat-gate.sh` (parses the table instead of pattern-matching a line ending), with `scripts/ci-gates-test.sh` proving on every PR that the gate can actually **fail**, plus a `make ci-gates-test` target. A gate nobody has seen fail is indistinguishable from a gate that cannot fail — which is exactly this bug.

## Ported from substrate vs. changed for elps

Ported in spirit from the equivalent fix in `luthersystems/substrate` (PR #361, branch `claude/audit-ci-gates`):

- awk table parser bounding the delta search after the last `±` and before `(p=`
- p-value parsed as a **full token** so `p=1.5e-05` is not truncated to `1.5` (the one path that failed *open*); a non-numeric p-value is a hard **exit 2**
- `~` rows **tallied separately**, so an all-`~` table is a successful comparison, not an unparseable one
- footnote filter anchored on the **leading superscript**, so an old-format `~ (all equal)` *data* row is not discarded as a footnote
- **exit 2, never 0**, when nothing can be interpreted — a gate that cannot read its input must never report success
- both benchstat table formats (new box-drawing, old `name`/`delta`)
- the "the old inline grep does NOT match a real regression" assertion

**Changed, because elps differs:**

**1. Metric direction.** elps' parser benchmarks call `b.SetBytes()` (`parser/rdparser/bench_test.go`, `elpstest/lisptest.go`), so benchstat emits a **`B/s` throughput column** substrate has no equivalent of. `B/s` is *higher-is-better*. Substrate's rule — any positive delta is a regression — ported verbatim would fail PRs for **getting faster**; a real elps run shows `Parser/diff.lisp-4 +178.17% (p=0.008)` on `B/s`, a 2.8x throughput *gain*. Metrics are now classified (`*/s` and old-format `speed` → higher-is-better; everything else → lower-is-better; unrecognised → lower-is-better, which can only over-report).

**2. Two thresholds instead of substrate's one.** I ran the workflow's exact benchmark command three times against **identical code** and compared pairwise. Counting only moves in the *bad* direction:

| metric | significant bad-direction moves | worst |
|---|---|---|
| `allocs/op` | 0 | — |
| `B/op` | 1 | **0.19%** |
| `B/s` | 8 | 23.48% |
| `sec/op` | 15 | **33.83%** |

Allocation metrics are effectively deterministic; timing metrics are noisy at `-benchtime=100ms -count=5` across two runner jobs. A single threshold is the wrong shape: low enough to be useful for allocations reds every PR on timing noise; high enough to survive timing noise lets real, perfectly-measurable allocation regressions through.

**3. Thresholds calibrated against a real CI null comparison.** This PR changes **no Go code**, so this branch's own first CI run is a genuine null comparison on the real infrastructure (`ubuntu-latest`, AMD EPYC 7763) — preserved verbatim as `scripts/testdata/benchstat-clean-ci.txt`. Worst bad-direction moves there:

```
Package/dump-array-4   +1.52% (p=0.008)  [sec/op]
Package/load-nested-4  +0.88% (p=0.032)  [sec/op]
Parser/sicp.lisp-4     +0.18% (p=0.016)  [B/op]
```

Real CI timing noise is **~1.5%**, not the ~34% my contended sandbox showed — the sandbox bounds noise from above but does not predict it. My initial 50% timing gate was ~30x looser than the environment needs, i.e. close to no gate at all, so the second commit recalibrates:

- `BENCH_REGRESSION_THRESHOLD_PCT` **15** (timing: `sec/op`, `B/s`) — ~10x headroom over the observed CI floor, and still above the 7.14% worst case substrate measured on its noisier self-hosted runners, so an unlucky run does not red an innocent PR.
- `BENCH_ALLOC_THRESHOLD_PCT` **5** (`B/op`, `allocs/op`) — observed 0.18%, ~26x headroom.

Both are env vars in the workflow; tuning needs no script change. I deliberately did **not** touch the benchmark sampling parameters (a longer `-benchtime`/higher `-count` is the root fix for timing headroom) so this stays a gate fix. I also did not drop the timing gate straight to the 5% that `.claude/skills/benchmark/SKILL.md` quotes — that number is for local, same-machine comparisons, and betting on one quiet CI run risks the gate being switched off again.

## Sibling defects found and fixed in the same workflow

All the same fail-open class as the headline bug:

- **`benchstat ... || true`** discarded a benchstat crash, making it indistinguishable from a clean comparison. Status is now captured and the gate adjudicates the output (crash text → exit 2 → job fails).
- **`go test -bench ... | tee`** took `tee`'s exit status. GitHub's default shell is `bash -e` *without* `pipefail`, so benchmarks that fail to compile or panic went **green** with error text in `bench-current.txt`. `set -o pipefail` added.
- **`has_comparison` was set twice and read by nothing** — a step output no one consumed. Replaced with `gate_status`, which is consumed.
- **The fail step keyed off an output an earlier cancelled step may never set.** It now runs under `always()` and passes only on `skipped` (no baseline) or `0`.

## Other workflows — reported, not fixed

- `elps.yml`, `govulncheck.yml`, `tree-sitter.yml`, `vscode-publish.yml`: **no dead gate of this class.** They are action-driven with no shell gate logic. `elps.yml`'s `elps fmt -l` step *looks* like a `gofmt -l` no-op gate but is genuinely live — `cmd/fmt.go` sets `exitCode = 1` when `fmtList && changed`.
- **5 unpinned actions** (`govulncheck.yml`: `actions/checkout@v4`, `actions/setup-go@v5`; `release-tag.yml`: `actions/checkout@v4`, `anthropics/claude-code-action@v1`, `actions/upload-artifact@v4`). The repo pinned everything to SHAs in f5b12e3; these two later workflows drifted back to floating tags. Out of scope here, so the self-test **allowlists them with a printed `WARN` and hard-fails on any new one** — an allowlist that can only shrink, rather than a silent omission.

## Verification

**Confirmed by CI.** Actions turned out to be working for this repo, so both commits were exercised end-to-end:

| run | commit | `CI Gate Self-Test` | `Benchmark Comparison` |
|---|---|---|---|
| [#474](https://github.com/luthersystems/elps/actions/runs/31134407529) | `5ae2cd5` (gate fix) | success | success |
| [#475](https://github.com/luthersystems/elps/actions/runs/31134811349) | `0dce112` (recalibration) | success | success |

Run #474's `Benchmark Comparison` ran a real comparison end-to-end through the new gate, posted its PR comment, and passed — no false positive. Run #475 confirms the same at the tightened 15% threshold.

**Local** (`bash -n`, `shellcheck -S warning` clean on both scripts; `make ci-gates-test` → **45 passed, 0 failed**):

| proof | result |
|---|---|
| gate **fires** on the reported `+50.00% (p=0.000 n=10)` sample | exit 1 |
| gate does **not** fire on a clean comparison (real CI null) | exit 0 |
| gate does **not** fire on a `~`-only table | exit 0 |
| gate does **not** fire on `B/s` throughput gains (even at a 0% gate) | exit 0 |
| dead-pattern guard catches a real reintroduction (injected) | caught at `benchmark.yml:229` |

**Negative-control matrix** — each fix reverted individually, self-test re-run:

| fix reverted | self-test | first failing assertion |
|---|---|---|
| metric direction (`B/s` higher-is-better) | **DETECTED** | the real CI noise IS seen; only the threshold holds it back |
| per-metric-class thresholds | **DETECTED** | +8% allocation regression fires the tight gate |
| p-value parsed as full token | **DETECTED** | sci-notation p-value read as significant |
| `~` rows tallied as interpreted | **DETECTED** | real CI comparison fully parsed |
| footnote filter anchored on superscript | **DETECTED** | old-format all-`(all equal)` table is clean, not exit 2 |
| exit 2 when nothing interpretable | **DETECTED** | benchstat crash is an error, not "clean" |
| gate invoked from the workflow | **DETECTED** | benchmark.yml no longer invokes the script |
| self-test wired into CI | **DETECTED** | benchmark.yml does not run ci-gates-test.sh |

**8/8 detected.** Worth noting: the first run of this matrix was **5/8** — the footnote-filter and both workflow-shape controls were *missed*, because the two shape guards were satisfied by the script names appearing in explanatory *comments*, and no assertion covered an all-`(all equal)` table. Those were real holes in the self-test, found only by running the negative controls. Both shape guards now anchor on a non-comment invocation, and `benchstat-allequal-old.txt` was added.

## A note on audit coverage

The elps audit that produced #308 examined 343 AI-authored commits and **did not find this**, even though the identical bug had already been found and fixed in substrate. That is a finding about audit coverage, not just about the gate: a defect class already confirmed in a sibling repo should be searched for directly across the others, rather than left to be rediscovered.

## Not verified

- **A real regression failing a real PR.** The gate has been proven to exit 1 on regression fixtures and to run green in CI, but no CI run has yet gone red on a genuine regression — by construction, since this PR contains none. The self-test is what stands in for that.
- **The distribution of CI noise.** The 15% timing threshold rests on one real null comparison plus substrate's independent 7.14% figure. More runs may justify tightening it; the fixture assertion is written so a retune below the real noise floor flips the test rather than silently reintroducing false positives.
- **The two 2026-02-08 failures' actual failing step** (logs past retention, `410 Gone`); ruled out as gate-caused by code inspection at those SHAs, not by log evidence.


sam-at-luther:
Superseded by #310, which consolidates this with #308 (the audit fixes) into a single PR off latest `main`.

Carried across at `0dce112` — both commits, including the threshold recalibration. Merged clean with #308.

Worth restating in the consolidated PR, since it is the part that would be easiest to lose: porting substrate's rule verbatim would have **failed PRs for getting faster**, because elps' parser benchmarks call `b.SetBytes()` and emit a `B/s` throughput column substrate has no equivalent of — a real run shows `+178.17%` on `B/s`, a 2.8× gain. And one threshold cannot serve both metrics: bad-direction noise measured 0.19% for `B/op` and 0% for `allocs/op`, but 33.83% for `sec/op`.

Closing so there is one thing to review; the branch is untouched.


---

PR #310, opened 2026-08-07
Audit fixes: tail-call budget, equal? on sorted-maps, panic surfacing, step limits, lint false positives, benchmark gate

sam-at-luther:
Consolidates the two outstanding elps branches into one PR off latest `main` (`6fb9f4d`). Supersedes **#308** and **#309**. Also folds in the three open Dependabot PRs (**#304** go-deps, **#305** vscode-deps, **#306** actions), which it supersedes too.

From an audit of 343 AI-authored commits on `main` since 2025-02-01: 9 findings fixed, 1 dropped as a non-finding, 1 confirmed and deferred, plus the dead benchmark gate found separately.

## The one with downstream blast radius

`9e5fd6c74` (Feb 2026) set `DefaultMaxLogicalStackHeight = 50000` in `StandardRuntime()`, where previously `&CallStack{}` meant unlimited. But `HeightLogical` **accumulates elided tail frames** (`env.go:1191`, `:1335`), so that constant is a *total tail-iteration budget*, not a nesting-depth guard:

```
n=24000   'done
n=25000   error: lisp:if: logical stack height exceeded maximum: 50001
n=100000  error: lisp:if: logical stack height exceeded maximum: 50001
```

A correct constant-space tail loop over a 30k-element collection was rejected, with a message that reads like runaway recursion — so whoever debugs it hunts for a missing base case that does not exist. It is data-dependent: passes every test and every small customer, then starts failing when one customer's data grows. Substrate pins elps and inherits this today.

**Fixed with a new `MaxTailIterations` in honest units, and `MaxHeightLogical` left off** rather than merely raised. With both enabled the logical bound fires first for any loop body nesting terminal forms more deeply than trivial, reintroducing the same shape-dependent failure with the same confusing message.

`MaxHeightPhysical` stays at 25,000, and that is deliberate — it guards a **fatal, uncatchable** failure. Measured with limits disabled: 200,000 non-tail frames complete (2.6 KB/frame, 7.0s); 400,000 gives `runtime: goroutine stack exceeds 1000000000-byte limit` / `fatal error: stack overflow`, killing the process. Crash point ≈372k, so the default sits ~15× below it. Note this reframes `9e5fd6c74`: before it, *neither* limit was set, so there was no protection against that crash at all. It fixed a real bug and bundled a wrong one with it.

The two limits are orthogonal — proven by setting the physical cap to **100 frames** with no tail bound: a 1,000,000-iteration tail loop completes and the stack unwinds to 0, while non-tail recursion 500 deep is refused at frame 101. So raising the tail bound carries no stack-crash risk.

## Also fixed

- **`equal?` returned false for sorted-maps identical through every accessor.** A real bug, not intentional: `docs/lang.md` already documented that lookup ignores key spelling and `get`/`key?`/`assoc`/`dissoc` all honour it — only `equal?` didn't. Equality depended on a map's *construction history*, because `Set` with a string key does not clear a symbol spelling recorded earlier. The `.Str` shortcut now applies only when **both** keys are string-like, falling back to `Equal` otherwise — an adversarial review caught that comparing `.Str` unconditionally made *every* key equal for embedder-supplied `Map` implementations with non-string keys, since `.Str` is empty there.
- **Go panics in host-registered builtins** were silently swallowed by `ignore-errors`/`handler-bind`; now surfaced as a non-forgeable `internal-panic` condition, exported as `lisp.IsInternalPanic`.
- **`WithMaxSteps`** was documented as per-evaluation but was process-cumulative and never reset — one budget-exhausting evaluation poisoned the runtime permanently. Now per-evaluation, including on `MacroCall`, which was the one exported entry point missing the reset.
- **Lint:** `builtin-arity` false-positived on `let`/`let*`/`flet`/`labels` (and its shadow set is now scoped to the binding form's subtree rather than file-global); `quote-call` flagged every correct `defconst` and suggested a fix that breaks the program; `unused-nolint` false-positived when `--checks` narrows the analyzer set.
- **`elps run -e` / `-p`** were documented in help text but dead — `-e` treated the expression as a filename.
- **`mcpserver` test** cannot pass as root; now skipped under `os.Geteuid() == 0`.

**Dropped:** `undefined-symbol` severity depending on macro location — investigated and rejected as a non-finding rather than shipping a fix for a non-bug.

## The benchmark gate could never fire

`.github/workflows/benchmark.yml:116` tested `grep -E '^\S.*\+$'`, requiring a line **ending** in a literal `+`. Benchstat lines never do — regression rows end in `(p=0.000 n=10)`, geomean rows in `%`. A 50% regression at p=0.000 produces **zero matches**, so `has_regression` was always false and the fail step never ran. 473 runs, no gate-attributable failure.

Ported from substrate's fix (same bug, fixed there in luthersystems/substrate#361), with two elps-specific corrections found by measurement:

1. **Metric direction.** elps' parser benchmarks call `b.SetBytes()`, so benchstat emits a `B/s` **throughput** column substrate has no equivalent of. Substrate's "any positive delta is a regression" rule, ported verbatim, would have **failed PRs for getting faster** — a real run shows `+178.17%` on `B/s`, a 2.8× gain.
2. **Two thresholds, not one.** Measured on identical code: bad-direction noise is 0.19% for `B/op`, 0% for `allocs/op`, but 33.83% for `sec/op`.

Sibling defects fixed in the same workflow: `benchstat … || true` swallowing a crash; `go test | tee` taking tee's exit status without `pipefail`, so **failing benchmarks went green**; a `has_comparison` output set twice and consumed by nothing; and a fail step keyed off an output a cancelled step may never set.

## Verification

Run on this exact tree (`e183a2e`):

| Check | Result |
|---|---|
| `make test` | **29 packages ok, 0 fail** |
| `make static-checks` | **0 issues** |
| `make ci-gates-test` (the new gate self-test) | **45 passed, 0 failed** |
| `elps doc -m` | all functions and exports documented |
| `elps lint` (CI invocation) | **17 diagnostics — identical to `main`** |
| `elps fmt -l` (CI invocation) · `go vet` | clean · clean |

Mutation-verified throughout — 16 in the first pass, 6 more after adversarial review, each failing a named test. Two initially came back clean and were fixed rather than waved through: one because the two halves of a fix overlapped, one because the naive mutation didn't compile so the "pass" was meaningless.

## Not fixed here

The JSON decoder rounds integers above 2^53 (pre-existing since 2018, not AI-authored). It is the root cause of the class substrate works around in its migration cache keys, and it needs its own decision: `statedb:get` deserializes through the same decoder, so erroring there would make already-written state **unreadable** rather than merely imprecise. Deliberately out of scope for a PR that already changes runtime defaults.

---


sam-at-luther:
## The benchmark failure is a true positive, not a false one

I was asked to confirm the diagnosis against the script rather than take it from the numbers. Having done that: the proposed mechanism (`B/s` magnitudes being tested against `BENCH_ALLOC_THRESHOLD_PCT`) is **not** what happened, and the gate needs no change.

### What actually fired

The summary I was given says "`B/op` everything ~ or +0.00%". The `lisp` package's `B/op` table also contains this row:

```
EnvFunCallRecursion-4   3.768Mi -> 4.086Mi   +8.44% (p=0.008 n=5)   [B/op]
```

That is an allocation metric at 8.44%, above the 5% allocation gate. Reproduced locally against the exact script on `e183a2e`:

```
REGRESSION  …/elps/lisp  B/op  EnvFunCallRecursion-4  delta=+8.44% p=0.008 (gate 5%)
below-gate  …/elps/lisp  B/s   Parser/sicp.lisp-4     delta=-5.97% p=0.008 (gate 15%) (higher is better)
below-gate  …/elps/lisp  B/s   Parser/stream.lisp-4   delta=-2.54% p=0.008 (gate 15%) (higher is better)
```

The `B/s` rows print `(gate 15%)` — they were already being judged against the **timing** gate and were correctly logged `below-gate`. `is_alloc_metric()` requires a `/op` suffix, so `B/s` never enters the allocation bucket.

### Why it's a real regression, not noise

`allocs/op` for the same benchmark is **identical** — `40.08k → 40.08k, +0.00% (p=0.008)`. The allocation *count* did not move; only the bytes. Same allocations, each bigger.

That points at a struct, and the diff confirms it: this PR adds `TailIterations int` to `CallFrame` (`lisp/stack.go`), which is allocated per call.

```
40,080 allocations/op × 8 bytes ≈ 313 KiB
observed:  4.086Mi − 3.768Mi     ≈ 325 KiB
```

The arithmetic lands. `p=0.008`, on the metric class that is near-deterministic in elps (measured noise ≤0.19%). The gate priced a real memory cost of the safety bound.

### This needs a decision, not a tuning

I have deliberately **not** changed any threshold or classification — silently widening the allocation gate to make this green would discard the gate's single most reliable signal, on its first real firing. The options are yours:

1. **Accept the cost** — the memory is the honest price of the tail-call bound, as the ~2% timing cost is. Raise `BENCH_ALLOC_THRESHOLD_PCT` (10% clears it) with a comment citing this run.
2. **Avoid the cost** — `CallFrame` may have padding a smaller type could occupy, or the counter could live on the `Stack` rather than per-frame.
3. **Scope it** — bound only the frames that can tail-loop.

Note the timing side is genuinely fine: worst timing delta is `EnvFunCallRecursion-4 +13.48%`, under the 15% gate. Raising *only* the allocation gate makes the whole table pass, which is pinned as an assertion.

### What I did change

`e707367` — two fixtures and eight assertions, no gate logic or workflow change:

- `benchstat-bps-dip-only.txt` — a several-percent `B/s` dip with flat `B/op`/`allocs/op` must exit 0, must print `gate 15%` on the `B/s` rows, and must be unaffected even at `BENCH_ALLOC_THRESHOLD_PCT=0`. Verbatim rows from this run.
- `benchstat-alloc-regression-live.txt` — this run's table; must exit 1 and name `EnvFunCallRecursion-4`.
- `B/s` **gains** re-verified as still never firing (even at a 0% gate), per the reverse check requested.
- A new negative control that deliberately mis-classifies `B/s` into the allocation bucket — the bug that was suspected. It is now **DETECTED**, so if it ever becomes real the suite catches it.

**9/9 negative controls detected; 53/53 assertions pass.** `bash -n` and `shellcheck -S warning` clean.

### The lesson stands, with a correction

The observation that "the self-test passed 45/45 while the live wiring was wrong" is the right instinct, and it is worth keeping — but it does not apply here: the wiring was right, and the suite's gap was that it could not *demonstrate* that, so the question had to be re-derived from raw numbers under time pressure. That is a real cost, and it is what the new fixtures remove. The sharper version of the lesson: **a fixture suite should be able to answer "is this a false positive?" without anyone re-reading the parser.** Until this commit, mine could not.

Worth flagging separately: the first summary of this run omitted the one row that mattered. When a gate fires, the gate's own report names the metric, the row and the threshold it breached — that line is worth reading before reconstructing the cause from the benchstat table.

---

PR #311, opened 2026-08-07
Struct alignment guard, 9 fuzzing fixes, debugger race, 16 more linters, and a benchmark gate that actually measures

sam-at-luther:
Started as a struct-alignment guard and grew into the wider hardening pass that
came out of it. Five separate workstreams; each is its own commit or run of
commits, and the commit messages carry the reasoning in full.

| # | Workstream | Commits |
|---|---|---|
| 1 | Align hot structs; gate `fieldalignment` in CI | `fb3fe27`, `04c8898`, `7cf9e4c` |
| 2 | Native `go test -fuzz` targets; 9 parser/formatter/minifier defects repaired | `0dd0036`, `2b5a23c`, `fde78f3` |
| 3 | Debugger pause race | `05cdedc` |
| 4 | 16 more golangci-lint checks, plus `exhaustive` and the `elpscheck` build | `1acdc3e`, `b9f610e`, `1fae820` |
| 5 | Benchmark gate: make it measure, and make it fail when it cannot | `7e011c5`, `0a97a6a`, `2ca1a26`, `493e6f6`, `ea40979`, `a52a0a0`, `7cd139d` |

**There is no speedup here.** Workstream 1 is a regression guard, not an
optimisation. At n=10 interleaved, 46 benchmark rows: `allocs/op` 0 significant,
`B/op` 0 with non-zero magnitude, `sec/op` 2 marginal rows where ~2.3 false
positives are expected by chance. Numbers and the mechanistic explanation are in
[this comment](https://github.com/luthersystems/elps/pull/311#issuecomment-5222996661).

---

## 1. Struct alignment

A new `int` field on `lisp.CallFrame` grew it 72 -> 80 bytes with `allocs/op`
unchanged, and only the benchmark gate caught it. `fieldalignment` was not
enabled anywhere. This turns that class into a lint error at review time.

| struct | size before -> after | GC scan extent before -> after |
|---|---|---|
| `LVal` | 112 -> 112 | **112 -> 64** |
| `LEnv` | 64 -> 64 | 64 -> 56 |
| `Runtime` | 184 -> 184 | 144 -> 112 |
| `CallStack` | 64 -> 64 | 48 -> 32 |
| `FunctionInfo` | 80 -> 80 | 80 -> 64 |
| `SourceMeta` | 112 -> **104** | 64 -> 56 |
| `Token` (parser) | 40 -> 40 | 24 -> 16 |
| `libhelp.SymbolDoc` | 56 -> 56 | 56 -> 48 |

Only `SourceMeta` shrank. `LVal` keeps its 112 bytes, so this is a **GC-scan
change, not an allocation-size change** — which is exactly why `B/op` and
`allocs/op` are unchanged.

`fieldalignment -fix` **deletes every doc comment on a reordered struct** (it
nuked all of `LVal`'s field documentation in a trial run). Every struct here was
reordered by hand or hand-repaired, and each hot struct gained a comment saying
its order is load-bearing. `make fieldalign-fix` uses `betteralign` instead,
which preserves comments.

### Is reordering a breaking change?

Field order is not part of a struct's semantics, so reordering is safe except in
five specific cases. All five were checked, not assumed:

| Risk | Method | Result |
|---|---|---|
| Positional (unkeyed) struct literals | AST scan of this repo + **substrate** | **None in substrate** (see below) |
| `unsafe` pointer arithmetic over offsets | grep `unsafe.`, `Offsetof`, `Sizeof`, `Alignof`, `uintptr` | None. The only `unsafe.` use is a CGO handle cast in `tree-sitter-elps` (separate module), not layout-dependent |
| Hand-rolled binary marshalling | grep `encoding/gob`, `encoding/binary` | Neither is imported anywhere |
| Reflection iterating fields by index | grep `.Field(`, `NumField()`, `FieldByIndex` | No occurrences |
| Struct comparison / map keys | reviewed | Unaffected by order |

Within elps the only exported multi-field types ever constructed positionally
are `parser/token.Rune` and `elpsutil.Builtin` — neither is flagged by
fieldalignment, so neither is touched.

**The substrate scan gave a false "clean" the first time**, and it is worth
recording, because a scanner that silently analyses nothing and reports clean is
the same failure class as a CI gate that cannot fail. The first scan used
`go/packages` with full type checking and reported zero positional elps
literals. Substrate declares Go 1.26; the machine had Go 1.25, so **190 of 192
packages failed to type-check**. With no `TypesInfo` every lookup missed and the
walker fell through — it reported zero because it had examined nothing. Caught
by asserting on the load itself: 0 fully type-checked out of 192 is the tell.

The replacement is a **syntax-only AST scan** that needs no type information, so
toolchain skew cannot silently disable it, and it was validated against a
positive control of planted literals before being trusted. Result: **246 files
parsed, 0 parse failures, 0 positional literals of elps types**, across 102
files importing 27 distinct elps packages. A third cross-check by grep looked
alarming — 99 hits shaped like `lisp.LVal{...}` — but all 99 are `[]*lisp.LVal{...}`
**slice** literals whose elements are function calls. Grep alone would have
raised a false alarm; the AST scan is what distinguishes them.

### Scope of the guard

Enforced on **all of `lisp/...`** and **all of `parser/...`** — whole trees, not
today's hot files, so a new struct in `lisp/lisplib/libjson` is covered the day
it lands. Narrowing to today's hot files is how `CallFrame` grew unnoticed in
the first place.

Two exemptions, both rules rather than uncovered paths, so the allowlist can
only shrink: **`lisp/x/`** (experimental debugger/profiler, off by default — the
tool's ordering dissolves ~15 documented field groupings to reclaim 8-16 bytes
once per session) and **`_test.go`**. Everything outside `lisp/` and `parser/`
is excluded because it is cold and the churn is not worth it — not for any
ordering reason.

### Negative controls

| # | Scenario | Expected | Result |
|---|---|---|---|
| 1 | Misaligned struct in `lisp/lisp.go` | red | **exit 1**, `size 32 could be 24` |
| 2 | `bool` inserted mid-`LVal` (the real regression class) | red | **exit 1**, `size 120 could be 112` |
| 3 | Misaligned struct in excluded `mcpserver/` | green | 0 hits |
| 4 | Misaligned struct in `lisp/*_test.go` | green | 0 hits |
| 5 | Misaligned struct in `lisp/lisplib/libjson` | red | **exit 1** |
| 6 | Misaligned struct in `lisp/lisplib/libhelp` | red | **exit 1** |
| 7 | Misaligned struct in exempt `lisp/x/debugger` | green | 0 hits |

For every "green" row the raw `fieldalignment` binary was run separately to
confirm the planted struct genuinely **is** misaligned — so those rows prove the
scoping works, not that the linter is off. All controls removed; tree is clean.

---

## 2. Fuzzing

Native `go test -fuzz` targets covering the strict, fault-tolerant and
format-preserving parsers, the lexer and byte scanner, the formatter and
minifier round-trips, and the JSON decoder. Seeds come from `internal/fuzzseed`
(the repo's real `.lisp` sources plus hand-written adversarial input); each
package's `testdata/fuzz/` holds the crashers found so far, which plain
`go test` replays as regression cases. `make fuzz` locally;
`.github/workflows/fuzz.yml` runs 30s/target on PRs and 10m/target nightly.

Nine defects found and repaired, including a **parser hang on the 4-byte input
`"abc\xff"`** (verified: hangs on `main`, returns on this branch) and minifier
nondeterminism (`main` produced 5 distinct outputs in 8 runs of the same input;
this branch produces 1 in 8).

`fde78f3` is worth a note because two obvious fixes were wrong before the third
worked, and the commit records both: `strconv.ParseInt`/`ParseFloat` misses
`-1abc`, and a lexer-level guard wrongly rejects the legitimate symbols `--` and
`-a`. The fix re-parses the name and asks whether it reads back as a symbol.

---

## 3. Debugger race

`TestEngine_FunctionBreakpoint` was flaky. `pausedEnv` was published in one
critical section and the stopped event dispatched outside it, so a test could
observe a pause that had not been announced.

The first fix — gating `IsPaused()` on callback completion — **deadlocked five
dapserver tests**: the callback writes to a pipe that blocks until the test
reads, while the test waits on `IsPaused()`. That is recorded in the commit so
nobody retries it. The shipped fix sets `stopDispatched` in the same critical
section that publishes `pausedEnv`, before the callback runs, and adds
`PausedUnannounced()` for tests that need the stricter predicate.

---

## 4. Linters

16 additional golangci-lint checks plus `exhaustive` (every no-default switch
over an enum is now closed), and `make static-checks` gained a second pass over
the `elpscheck` build tag — golangci-lint only analyses the default build, so
tagged code was invisible to it. That second pass immediately found a real
issue, fixed in `1fae820`.

---

## 5. Benchmark gate

The gate was **dead for 473 runs**: an inline `grep -E '^\S.*\+$'` that no
benchstat line can match, so the workflow reported success by construction
rather than by measurement. Fixing it exposed three further ways it failed
without going red, all now closed:

- **Cross-machine baselines.** It compared against a cached baseline recorded on
  a different machine on a different day, attributing the hardware delta to the
  code delta. Consecutive runs of identical code drew an EPYC 7763 and an EPYC
  9V74 and paired nothing. Both arms are now measured **in one job on one
  runner**, interleaved round by round so a runner that slows partway through
  biases both arms equally rather than pooling the drift into one.
- **GOMAXPROCS suffix.** `go test` appends GOMAXPROCS to every benchmark name
  and benchstat pairs on the full name, so arms at different core counts share
  no names and pair nothing — from a table that looks entirely normal. Sprung by
  editing `runs-on`, which does not look like touching this gate. Prevented by
  pinning `GOMAXPROCS`; detected by `scripts/bench-arms-check.sh`, which names
  the cause instead of the symptom.
- **Jobs that queue forever.** A `runs-on` label no runner available to this
  repo carries does not fail — it queues, publishing **no status at all**, so
  the board shows the remaining checks green and reads as a pass. That cost five
  consecutive pushes here (`ubuntu-arm-4core-150gb`, real but substrate's, not
  elps'). `timeout-minutes` cannot cover it; that clock starts when a job
  *starts*. `scripts/ci-queue-watchdog.cjs` polls the run's own job list and
  fails if the watched job has not left `queued` within 15 minutes.

Benchmarks now run on **ARM**, since the consumers pinning this module ship ARM
and struct layout, padding and GC scan extent are not arch-invariant.

The gate itself lives in `scripts/benchstat-gate.sh` so it can be unit-tested,
with two thresholds because elps' metrics have wildly different noise (measured:
allocation ~0.18%, timing ~1.5% on real CI). It also classifies metric
direction — elps emits a `B/s` throughput column via `b.SetBytes()`, and a naive
"any positive delta is a regression" rule flags throughput *gains* as
regressions.

Everything above is guarded by `scripts/ci-gates-test.sh`, which runs as its own
CI job and asserts, among other things, that the gate fires on known-regression
fixtures, stays quiet on a real null comparison, and that the workflow still
invokes the script rather than reinlining it. **Every guard was verified to fail
by reverting the thing it guards in a scratch copy** — a guard nobody has
watched fail is indistinguishable from one that cannot.

---

## Verification

- `golangci-lint run ./...` — 0 issues; second pass under `-tags elpscheck` also clean
- `go vet ./...`, `./elps doc -m`, `./elps fmt -l`, `./elps lint` — all clean
- `go test ./...` — passes but for the pre-existing root-permission case below
- `scripts/ci-gates-test.sh` — 82 passed, 0 failed
- Live CI: 9/9 green, with the benchmark gate reaching a real verdict (40
  benchmarks paired on both arms, n=10 each, 82 delta rows interpreted, 0 at or
  above the gate)

**Pre-existing on `main`, not fixed here:**
`TestWorkspaceTraversalErrorsAreLoggedAndSkipped` fails when run as root (it
chmod-000's a directory; root ignores permission bits — it passes in CI, which
runs unprivileged), and 34 files are not `gofmt`-clean under Go 1.25's
comment-alignment rules.

**Known limitation, documented in the code:** `go test` emits no `cpu:` header
on linux/arm64, so `bench-arms-check.sh`'s cross-CPU check is inert on this
runner and logs "skipping" rather than "match". The one-job design is what
actually prevents that failure; the check is a backstop for amd64.


sam-at-luther:
## Benchmark measurement: there is no speedup

**Headline: this change is performance-neutral. It is a correctness/regression guard, not an optimisation.** Reporting that plainly rather than fishing for a win.

### Method

- `benchstat`, **n=10** per arm (CI's `BENCH_COUNT: 5` is below the 6 benchstat needs for a confidence interval — at n=5 the floor p-value is 0.008, which means only "all samples ranked one way" and is exactly how environment bias masquerades as significance).
- Base = `origin/main` @ `6fb9f4d`, branch = this PR, same machine (4 vCPU Xeon @ 2.80GHz).
- **Interleaved**, arms alternating order every round, so slow drift hits both arms equally instead of pooling into one.
- `-benchtime=300ms`, whole suite (`./lisp/... ./parser/...`), 46 benchmark rows per metric.

### Result across all 46 rows

| metric | rows | flagged significant | with non-zero magnitude |
|---|---|---|---|
| `sec/op` | 46 | 2 | **2** |
| `B/op` | 46 | 2 | **0** (both `+0.00%`/`-0.00%`) |
| `allocs/op` | 46 | **0** | 0 |

The only two `sec/op` rows that cleared p<0.05 are `libjson/Encode_stringNumbers +6.20% (p=0.035)` and
`rdparser/Parser/complex.lisp +3.87% (p=0.023)` — both marginal, both in the *slower* direction, and
**at α=0.05 across 46 rows you expect ~2.3 false positives by chance. Exactly 2 were observed.** That is a
null result, not a regression.

Geomeans (`sec/op`), this run vs the CI run on a different runner:

| pkg | this run (n=10, interleaved) | CI (n=5) |
|---|---|---|
| `lisp` | **-0.82%** | +0.47% |
| `libjson` | **+1.61%** | -0.14% |
| `rdparser` | **+2.33%** | +0.26% |
| `regexparser` | **+0.17%** | +0.29% |

The two independent runs **disagree on the sign** for `lisp` and `libjson`. That is the signature of noise,
and it is the strongest single piece of evidence that there is no real effect either way.

`B/op` and `allocs/op` geomeans are `+0.00%` everywhere, which is expected **by construction**: no struct
except `SourceMeta` changed size, so the allocator sees identical requests.

### Why the GC-scan win doesn't show up in wall clock

`LVal`'s GC scan extent really does drop 112 -> 64 bytes. It just doesn't translate, for three reasons:

1. **Size is unchanged (112 bytes).** Same size class, same allocator pressure, same heap footprint — so
   `B/op` and `allocs/op` cannot move, and neither can allocation-path time.
2. **Scan extent only costs anything for *live* objects during mark.** These benchmarks generate almost
   entirely short-lived garbage; dead objects are never scanned. The surviving set that actually benefits
   is small.
3. **These are short benchmarks with small heaps.** The mark phase is a minor fraction of total runtime, so
   even a 43% cut in bytes-scanned-per-live-`LVal` moves a small share of a small share.

A workload with a large *live* `LVal` graph (a big long-lived `sorted-map`, a long-running phylum holding
state across many calls) is where this would pay off. Nothing in the current benchmark suite looks like that.

### So what is this PR worth?

The `fieldalignment` gate, not the reorder. The reorder is what makes the gate green today; the gate is what
stops `LVal` silently growing tomorrow. Negative controls proving it can actually fail are in the PR body.

---

### Incidental finding: the benchmark regression gate is dead

Worth flagging separately — **not fixed here, and not related to this change.**

`benchmark.yml` decides whether to fail with:

```bash
if grep -E '^\S.*\+$' benchstat-output.txt | grep -qv '^name'; then
```

That regex requires a line to **end** with a literal `+`. benchstat rows end with `)` (per-benchmark) or
`%` (geomean) — never `+`. Tested against this PR's own CI benchstat output, which contains a genuine
`+3.84% (p=0.008)` row: **0 matches.** So `has_regression` can never be set, and the "Fail on regressions"
step is unreachable.

This is visible on this very PR: the CI comment above flagged `EnvFunCallOpt-4 +3.84% (p=0.008)`,
`EnvFunCallPos-4 +3.62%`, `EnvFunCallKey-4 +3.47%` and `EnvFunCallTailRec-4 +3.30%`, and the job still
concluded `success`. (Those four are themselves n=5 artifacts — at n=10 interleaved they come back as
`~` with p=0.089 / 0.218 / 0.105 / 0.631, and `EnvFunCallPos`'s point estimate moves the *other* way.
So the gate reported four regressions that aren't real, and would have been unable to fail on them anyway.)

Two separate bugs worth their own issue: the dead regex, and `BENCH_COUNT: 5` being below benchstat's
minimum of 6.


---

sam-at-luther:
## `Lint & Test` went red on `10efb8f` — pre-existing flaky test, not this PR

Recording the diagnosis rather than just re-running until green.

**Failure:** `make race` -> `TestEngine_FunctionBreakpoint` in `lisp/x/debugger`

```
--- Expected
+++ Actual
-(debugger.StopReason) (len=19) "function breakpoint"
+(debugger.StopReason) ""
```

### Why it is not this PR

1. **`lisp/x/` is byte-identical to `main` on this branch.** `git diff origin/main -- lisp/x/` is empty
   — that subtree was reordered during the sweep and then deliberately reverted.
2. **The previous commit on this same branch (`a041e44`) passed `make race`** with every one of the
   `lisp/` struct reorders already in place. The only delta between the green run and the red run is
   `.golangci.yml` plus a four-field reorder of `libhelp.SymbolDoc` — neither is reachable from the
   debugger.
3. **Reproduced on pristine `origin/main`: 2 failures in 25 runs** of `go test -race ./lisp/x/debugger/`
   under CPU contention (8 busy loops on a 4-vCPU box). The flaking test there was
   `TestEngine_RequestPause` — a sibling in the same file with the identical pattern.

### The actual bug (test synchronisation, not a data race in the engine)

`engine_test.go` waits on one condition and asserts on a different one:

```go
require.Eventually(t, func() bool { return e.IsPaused() }, 2*time.Second, 10*time.Millisecond, ...)

mu.Lock()
assert.Equal(t, StopFunctionBreakpoint, stopReason)   // written by the event callback
mu.Unlock()
```

`stopReason` is written by the `WithEventCallback` handler, but the test only establishes
happens-before on `e.IsPaused()`. Nothing orders "engine has entered the paused state" against "the
event callback has run and stored the reason", so the assertion can legitimately observe the zero
value `""`. The mutex makes the access race-free, which is why `-race` reports no data race — it just
reports the wrong value. `t.Parallel()` plus `-race` scheduling on a loaded runner widens the window.

`TestEngine_RequestPause` (line ~965) has the same `stopReason`-via-callback shape.

The fix would be to make `Eventually` wait on the thing actually asserted, e.g. poll until
`stopReason != ""` under the mutex, rather than on `IsPaused()`.

**Not fixed here** — `lisp/x/` is deliberately out of scope for this PR, and this is a pre-existing
defect on `main`. Worth its own issue, alongside the dead benchmark-gate regex reported above. Failed
jobs re-run to confirm the board goes green.


---

sam-at-luther:
## Rebased onto `main` (post-#310), plus the debugger race fixed here

Force-pushed as `2e6ec72`, rebased onto `919ef2d` (#310 merged). Two things changed.

### 1. Rebase conflicts, and how they were resolved

Both were real semantic conflicts with #310, not textual noise:

- **`CallStack`** — #310 added `MaxTailIterations` and documented all three limits; this PR had reordered the fields. Resolved by keeping #310's fields and docs and applying this PR's ordering rule (both slice headers lead, scalars trail), which is what the struct's own doc comment already prescribes.
- **`Runtime`** — #310 added `totalSteps` and `evalDepth` and re-scoped `steps` to per-evaluation. The `a041e44` side had deleted the old block wholesale, so a naive resolution would have silently dropped both new fields and reverted `steps`' doc comment to "Cumulative step counter". Resolved by carrying #310's fields and semantics into the trailing scalar block.

`CallFrame` still measures **72 bytes**, so #310's deliberate `int32`-after-the-bools placement survives the reorder intact.

### 2. The flaky debugger test is fixed here rather than deferred

The earlier revision documented `TestEngine_FunctionBreakpoint` as a pre-existing `main` defect and left it. It's fixed now, and the root cause turned out to reach into shipped code.

`Engine.IsPaused()` goes true **before** the `EventStopped` callback runs — `WaitIfPaused` has to publish the paused state first, because a DAP client answers the stopped event with `stackTrace`/`scopes` requests served on another goroutine and those call `PausedState`. Six engine tests waited on `IsPaused()` and then read a variable the callback writes, with no happens-before between the two. `TestEngine_RequestPause` already carried a hand-rolled fix for exactly this, which is what identified the shape.

Same window was a **real product bug**: dapserver's late-connect path in `onConfigurationDone` fires on `(!stoppedSent && engine.IsPaused())`, and both hold inside the window — so it could send the client a second `stopped` event for a pause the engine was already announcing. It now asks a new `PausedUnannounced()`, which the engine latches in the same critical section that publishes the pause, so the pair is never observed torn.

**One thing worth flagging, because the tests caught me.** My first attempt gated `IsPaused()` on the callback having *returned*. That reads better and fixed the flake — and **deadlocked five dapserver tests**, whose callback writes to a pipe that blocks until the test reads, while the test waits on `IsPaused()`. The callback runs on the eval goroutine and is allowed to block on its consumer, so nothing observable may wait on it finishing. The flag is now set *before* the callback, and the reason is recorded on `IsPaused` so the next person doesn't retry it.

Mutation-verified: never latching the flag, and latching it unconditionally, each turn one of the two new tests red. A third mutation (dropping the reset on resume) did **not** fail — `PausedUnannounced` conjoins the flag with `pausedEnv`, which is nil once the pause ends, so that assertion could never fail. It was removed rather than left in, and the reset is documented as the redundancy it actually is.

### Corrections to the size table above

The rebase moved two baselines, because #310 added fields to both structs. The rest of the table is unchanged.

| struct | GC scan extent before -> after | note |
|---|---|---|
| `CallStack` | **56 -> 32** | was reported 48 -> 32 |
| `Runtime` | **160 -> 112** | was reported 144 -> 112 |
| `Runtime` (size) | **200 -> 200** | was reported 184 -> 184 |

### Verification on `2e6ec72`

| Check | Result |
|---|---|
| `go test ./...` | all packages pass |
| `make race` (whole repo) | clean |
| `golangci-lint run ./...` | **0 issues** — the fieldalignment gate passes against #310's structs |
| `make ci-gates-test` | 53 passed, 0 failed |
| `elps lint` (CI invocation) | 9 diagnostics — **identical to `main`**, verified by building and running `main`'s own binary |
| `elps fmt -l` (CI invocation) · `elps doc -m` | clean · all documented |
| Stress: 3 affected tests, `-race`, `GOMAXPROCS=1` | **0 failures in 120 runs** |

The pre-existing root-permission failure noted in the PR body (`TestWorkspaceTraversalErrorsAreLoggedAndSkipped`) is gone — #310 fixed it.

---

sam-at-luther:
## Static-analysis expansion: 17 linters, plus a correction

Three commits added (`21bd007`, `24bbe35`, `c511fa5`) on top of the alignment work. Combined tree: `go test ./...` clean, `make race` clean, `golangci-lint run` **0 issues**, `make test-elpscheck` clean, `elps lint` **9 diagnostics — identical to `main`**, `ci-gates-test` 53/53.

### Correction: the motivating example in `21bd007`'s commit message is wrong

That message opens by claiming an unhandled `lisp.LType` would be "serialised as nothing" into a ledger via `libjson`. **That is not true, and I am the one who asserted it.**

`lisp/lisplib/libjson/json.go:339` is inside `GoValue`, which carries an explicit `Deprecated: no longer used internally for serialization` marker; its pass-through is documented behaviour. The real serialization path is a table, not a switch — `encode.go:56`:

```go
if fn := encoderFuncs[v.Type]; fn != nil { return fn(enc, v) }
return fmt.Errorf("invalid type encountered: %v", lisp.GetType(v))
```

An unregistered type **errors**. That hole was already closed. The commit message does self-correct further down, but it leads with the overstatement — noting it here so the record is straight.

The genuinely useful consequence: because the encoder is a **table and not a switch**, `exhaustive` is structurally blind to it. A new `LType` would get a nil entry with nothing failing at build time. So `TestEncoderTypeCoverage` was added — every `LType` must have an encoder or a recorded reason it has none — which is the guard the concern actually wanted and which no linter would have provided.

### Counting note that matters for anyone re-measuring

`golangci-lint`'s defaults (`max-same-issues: 3`, `max-issues-per-linter: 50`) silently collapse repeats. Every count in my earlier analysis was truncated:

| linter | truncated | real |
|---|---|---|
| testifylint | 32 | **173** |
| perfsprint | 12 | **47** |
| intrange | 6 | **43** |
| errorlint | 13 | **28** |
| exhaustive (tuned) | 24 | **26** |
| exhaustive (untuned) | 50 | **75** |

~350 findings, not ~110. The config now sets both limits to `0`: a gate that hides findings lies about how much debt is left.

### `exhaustive` (`21bd007`) — 26 sites, no exclusions

Configured `default-signifies-exhaustive: true`; a switch with a `default:` has made an explicit decision, so it is not a silent fallthrough. All 26 resolved in code: 3 rewritten as expressions, 21 explicitly enumerated with behaviour preserved, 1 panic moved into `default:`, and **1 real behaviour change** — `formatter/printer.go` `writeAtom` wrote *nothing* for a non-numeric value, which in a formatter is silent deletion of source. Now falls back to `v.String()`. Unreachable today; `TestWriteAtomUnexpectedType` pins it and fails against the old code.

Negative control: adding an `LDecimal` member to `lisp.LType` turned **8 sites red**, each naming it. Sites resolved by expression or `default:` correctly stayed quiet.

### The 16-linter set (`24bbe35`) — 75 files

All 16 negative-controlled individually (plant → confirm red → remove). One result worth recording: **`wastedassign`'s obvious plant did not fire**, because `ineffassign` reports the same position first and golangci-lint dedupes by position. A second plant with a shape only `wastedassign` catches did fire.

Three exclusion rules, 15 findings, each naming its sites and marked shrink-only: `nilerr` where "error means absent" is the contract (a `filepath.Walk` that must not abort a workspace scan; an LSP formatter that must not pop a dialog on unparseable input), `noctx` on two exported `ServeTCP*` functions where the only non-breaking fix is ceremony with no cancellation, and `testifylint` where `assert.*` is used as a *value* and `require.*` returns nothing.

Not cosmetic, flagged deliberately: **23 `assert.NoError` → `require.NoError`** (tests now abort rather than dereference a value they failed to obtain), and **12 dead `//nolint` directives deleted** — including `//nolint:gosec` on `os.ReadFile`, so a future refactor re-triggering G304 will red CI.

A live latent bug turned up in `lint/lint_test.go`: eight prose comments beginning with the word "nolint" were being parsed as real blanket `//nolint` directives. Reworded rather than accepting the tool's suggested `// nolint` → `//nolint` reformat, which would have made the accident permanent.

### `c511fa5` — linters could not see the tagged build at all

`golangci-lint` analyses only the **default** build. Any file behind a build tag is invisible to every linter, so `golangci-lint run ./...` can report **0 issues** on a tree that has findings.

Not hypothetical: with the default build cleared to zero, `--build-tags elpscheck ./lisp/...` still reported a `testifylint` finding in `lisp/singleton_check_elpscheck_test.go`. `static-checks` now runs a second pass under the tag.

The one-line fix then orphaned the `strings` import and **broke the tagged build** — which `go build ./...` does not catch either. Same blind spot, twice in five minutes; that is why the pass earns its place over a spot fix.

Negative-controlled: reverting the assertion makes the tagged pass report it while the default pass stays at 0.

### Not verified

- Whether `perfsprint`/`intrange` edits to `lisp/env.go`, `macro.go`, `builtins.go` and `parser/token/scanner.go` move the benchmarks. Not measured locally.
- `tree-sitter-elps` — separate module, not reached by `./...`.
- Interactive REPL behaviour after the `repl/repl.go` refactor, beyond `go test ./repl/`.

---

sam-at-luther:
## Native fuzzing: 9 real defects, one 4-byte process kill

The repo had zero fuzzing while parsing untrusted input — phylum source is customer-uploaded and executed as Fabric chaincode in substrate. Three commits: `a888b6f` (targets + fixes), `194adfa` (integration lint), plus the CI wiring.

### The headline, verified against unmodified `main`

`ParseProgram` on the 4-byte input `"abc\xff"`:

| tree | result |
|---|---|
| `origin/main` (`919ef2d`) | **hangs; unbounded memory until the process dies** |
| this branch | returns, 0 expressions |

Built the same probe against both trees rather than trusting the report. `lexer.readToken` fell through when `Accept` failed on an undecodable byte and re-dispatched on the *previous* rune **without consuming input** — an endless stream of zero-width SYMBOL tokens that `ParseProgram` accumulated. The `lex.emitError(...)` whose return value was discarded on the same statement is fixed with it.

### Every defect found

**Fixed (9)**

| # | Defect | Trigger |
|---|---|---|
| 1 | Hang + unbounded memory in `rdparser.ParseProgram` | `"abc\xff"` |
| 2 | Panic `index out of range [-1]` in `token.Scanner.checkRuneError` — reachable on `main` through the exported `Scanner` API alone | `"\xe4\xb8"` |
| 3 | `elps fmt` emits unreadable source (hash-bang hits a whitespace guard that has no business applying to it) | `"#!"` |
| 4 | Zero-length and digit-named symbols after `#'` — `"#'0"` yields a symbol named `0` that prints as an **integer** | `"#'"`, `"#'0"` |
| 5 | `elps fmt` **changes the program** — a dash run tokenises differently before `)`, and dropping the space rewrites one symbol into two | `"(------ )"` |
| 6 | Extra quote level: synthesised `#^`/`#'` forms carry no `BracketType`, so the fallback quotes them again | `"'#^0"` |
| 7 | `Format` not idempotent — blank-line bookkeeping read from the inner token instead of the prefix token | `";\n'\n\n0"` |
| 8 | `elps fmt` **destroys the file**: `(lisp:expr (+ 1 (f x)))` re-sugars to `#^(+ 1 (f x))` → `unbound-expression-error` | as shown |
| 9 | **Minifier non-deterministic** — `applyAssignments` ranged over a map, writing through AST nodes two symbol records can share | `"(())(defun 2(e'e))"` |

Five of these are `elps fmt` corrupting source it was asked to format, and #9 matters because the minifier's output *is* the chaincode — non-determinism there breaks reproducible builds and makes the symbol map useless for decoding a trace. 300/300 identical after the fix.

**Found, NOT fixed (1)**

Catastrophic backtracking in `regexparser.ParseLVal`. Measured: 12 unmatched `(` → 68 ms, 14 → 292 ms, 16 → 1.1 s, 18 → 4.8 s, 20 → no result in 10 s. A ~24-byte input never returns. Inherent to the `newParsecParser` grammar — `sexpr`/`sexprOUnmatched`/`vector`/`qexprOUnmatched` are `OrdChoice` alternatives each re-running `Kleene(&expr)` over the same suffix, unmemoised. Making it linear is a redesign, not a fix.

`FuzzParseLVal` is registered but `f.Skip`ped with the measurement table and a `TODO(fuzz)`; re-enabling is a one-line delete. Small-input coverage stays via `TestParseLValHandlesSmallInput`, and a companion test **logs without asserting** which side of a deadline the parse lands on, so a future fix gets noticed. Nothing in the repo selects this reader.

### Targets and CI

11 targets across `rdparser` (strict, fault-tolerant, format-preserving), `lexer`, `token` (driven by a fuzzer-supplied operation script so orderings are explored, not fixed), `formatter`, `minifier`, `libjson`, `regexparser`. Invariants are round-trip properties where the repo already guarantees them — AST preservation, idempotence, strict/fault-tolerant agreement, minifier determinism — each justified against an existing test rather than assumed.

Deliberately **not** asserted: JSON numeric-text preservation. Decoding through `float64` rounds above 2^53; comparing decode-of-re-encode against the *first* decode sidesteps that known out-of-scope defect so it generates no noise.

- **PR path: 4m56s measured**, separate job, parallel with the rest.
- Nightly 10m/target; manual dispatch 60s.
- Seed corpus under plain `go test`: **1.3s**.
- Triple time bound: `fuzz.sh` refuses an unparsable `FUZZTIME` (exit 2), computes a hard `-timeout` per invocation, and the job carries `timeout-minutes: 120`.
- `ci-gates-test` gains **`FuzzGateSelfTest`** — a deliberately-failing target, inert unless armed by that suite — proving end-to-end that a crashing target reddens the board. 53 → **75 assertions**.

### An integration finding neither agent could have seen

The combined tree failed `golangci-lint` with 3 issues. The fuzz targets were written against `05cdedc`, before the expanded linter set landed, so nothing had ever checked them against `intrange`/`nolintlint`. Fixed in `194adfa` and called out as a property of parallel work rather than anyone's mistake.

### Verification on `194adfa`

`go build` · `go test ./...` · `make race` · `make test-elpscheck` — all clean. `golangci-lint run` **0 issues** on the default *and* `elpscheck` builds. `elps lint` **9 diagnostics, identical to `main`**. `elps doc -m`, `elps fmt -l` clean. `ci-gates-test` **75 passed, 0 failed**.

### Carried forward, not buried

- One `FuzzFormat` failure in an early 30s sweep wrote **no** crasher to `testdata` and has not recurred in 11 subsequent runs. Most consistent with a worker-process death rather than an input-attributable crash. `fuzz.sh` does not filter `go test` output, so a recurrence is fully diagnosable from the job log.
- No `FuzzEval`. Evaluating fuzzer input means non-terminating programs with no per-input deadline in native fuzzing; it needs a step/alloc-budgeted harness. Out of scope.
- No targets for `analysis`, `lint`, `lsp`, `repl`, `tree-sitter-elps`. All targets take raw `[]byte` — no grammar-aware fuzzing.

---

PR #321, opened 2026-08-08
refactor(parser): remove regexparser and the goparsec dependency

### LINKED ISSUE #315: goparsec scanString panics on an unterminated string literal (2-byte input, upstream)
Found by differential fuzzing of `parser/regexparser` (PR #311). Mitigated, not fixed — the defect is in a third-party dependency.

## Reproduction

```go
regexparser.ParseLVal("\"(")   // panic: index out of range
```

Two bytes.

## Root cause

Upstream in goparsec's `scanString`:

- it walks the input with an unbounded `for txt[e] != '"'`, running off the end when the literal is never closed
- it also indexes `txt[e+1]` on a trailing backslash

## Why it is not fixed here

Both available fixes are bad:

1. **Replace `parsec.String()`** — goparsec's escape table is not Go's, so substituting our own changes the accepted language. That difference is real and now pinned in `TestParserDivergence`: `"\x41"` yields `"A"` from rdparser and `"\x0041"` from regexparser.
2. **Transcribe the ~80 lines of dependency internals** — rots on the next bump, and we would own a fork of a parser we do not otherwise maintain.

## Current mitigation

The fuzz target excludes input containing `"`, on the argument that `parsec.String()` does nothing unless the byte at the cursor is `"`, so input with no `"` cannot reach `scanString` from any cursor.

`TestParseLValPanicsOnUnterminatedStringLiteral` asserts the **current broken behaviour**, so a goparsec bump that fixes it upstream turns the board red and prompts deleting both the test and the exclusion. That is deliberate: it makes the workaround self-retiring instead of permanent.

## Cost of the mitigation

**`regexparser` string literals are entirely unfuzzed**, in both the differential and standalone targets. `newAST`'s `nodeTerm` string path has no coverage.

## Blast radius

`regexparser` is dead code in this repository — nothing selects it (`parser.NewReader` returns `rdparser`). This matters only to an external embedder that chose it explicitly.

## Next steps

- Check whether upstream has already fixed it; if so, bump and delete the workaround.
- If not, consider filing it upstream with this reproduction.
- If neither, consider whether `regexparser` should be deprecated outright given nothing selects it.

sam-at-luther:
Removes `parser/regexparser`, the alternative goparsec-based reader, and drops
`github.com/prataprc/goparsec` from the module.

Closes #315

## What was removed

| File | Lines |
|---|---|
| `parser/regexparser/parser.go` | 407 |
| `parser/regexparser/bench_test.go` | 25 |
| `go.mod` — `github.com/prataprc/goparsec` require | 1 |
| `go.sum` — its two hash lines | 2 |
| `CLAUDE.md` — the "two parser implementations" line, reworded | 1 |

Net: 5 files changed, 1 insertion, 436 deletions.

`goparsec` is gone from `go.sum` and from `go list -m all`. `go mod tidy` is a
no-op afterwards.

## Why

- **Nothing selects it.** `parser.NewReader` returns an `rdparser` reader (or the
  format-preserving `rdparser.NewFormatting` variant). No other elps package
  imports `regexparser`, and `luthersystems/substrate` does not reference it
  either — substrate consumes `minifier` and the default reader only.
- **It was our only third-party parsing dependency**, pinned to an untagged
  pseudo-version from December 2021 (`v0.0.0-20211219142520-daac0e635e7e`).
- **It is the direct source of #315.** Reproduced on `main` before deleting:

  ```
  regexparser.ParseLVal([]byte("\"(" ))
  → panic: runtime error: index out of range [2] with length 2
  ```

  The fault is inside the vendored combinator library, so it could only ever be
  worked around, not fixed. The same 2-byte input through the surviving
  `rdparser` is handled cleanly — `scan-error: unexpected rune '\x00'`, non-zero
  exit, no panic — via `elps run`, `elps fmt`, and `elps lint`.

## Formatter / minifier equivalence

`elps fmt` output and `elps minify` output are load-bearing for substrate:
`minify`'s output *is* the chaincode substrate ships. Binaries were built before
and after the deletion and run over both corpora; outputs diffed byte-for-byte.

| Corpus | Files | Captured bytes | Result |
|---|---|---|---|
| `elps fmt` — elps repo `.lisp` (excl. `tree-sitter-elps/`) | 33 | 90,848 | byte-identical |
| `elps fmt` — `substrate/internal/substrate/shirocore/` | 20 | 555,422 | byte-identical |
| `elps fmt` — entire substrate `.lisp` corpus | 93 | 782,545 | byte-identical |
| `elps minify` — shirocore, per file | 20 | 410,640 | 19/20 byte-identical, see note |
| `elps minify --workspace -w` — whole shirocore dir | 20 | — | 19/20 byte-identical, same note |

### The shipped artifact is byte-identical

The strongest check reproduces substrate's actual `genshirocore` pipeline (the
11 semantic source files + `.spec` exclusion sets through `minifier.Minify`,
`PreserveParams: true`) against this worktree's `minifier` package:

```
shirocore.lisp  158,839 bytes  md5 0cd447271419991d1a2e765b0d4ea7ea
map.json        183,561 bytes  md5 43e0dab9c496bbe268748e838ef153c5
```

Identical before and after — **and identical to the `shirocore.lisp` currently
committed in substrate**. Re-ran 5x on the post-deletion tree: same hash every
time.

### Note on `utils_test.lisp`

One file in the shirocore directory, `utils_test.lisp`, minifies to different
output before vs. after. This is **pre-existing run-to-run nondeterminism on
`main`, not a behaviour change**, and it does not touch the shipped artifact
(`utils_test.lisp` is a test file, not one of the 11 files `genshirocore`
consumes).

Evidence — the *same* `main` binary produces different output on consecutive
runs:

```
before-run1 e59cb0a6   before-run2 24c8d26c   before-run3 15bd583e
before-run4 cd1f15d6   before-run5 e59cb0a6   before-run6 e7ea783e
```

At n=60 runs per binary the output populations converge to the identical
8-element set:

```
before distinct(8): 15bd583e 24c8d26c 58f2d93d 91feb7f2 cd1f15d6 e59cb0a6 e7ea783e f00e18b2
after  distinct(8): 15bd583e 24c8d26c 58f2d93d 91feb7f2 cd1f15d6 e59cb0a6 e7ea783e f00e18b2
```

Every other file in the corpus is stable at 1 distinct hash over 12 runs on both
binaries. Worth a follow-up issue against the minifier; out of scope here.

## `libregexp` is untouched

`parser/regexparser` had nothing to do with ELPS's regular-expression *language*
support. That lives in `lisp/lisplib/libregexp` (the `regexp:` package wrapping
Go's stdlib `regexp`) and does not appear anywhere in this diff.

- `go test ./lisp/lisplib/libregexp/` — PASS (`$load`, `regexp-compile`,
  `regexp-match?`).
- A 40-assertion probe driving all four builtins (`regexp?`, `regexp-compile`,
  `regexp-pattern`, `regexp-match?`) — compiled and raw-pattern forms, string and
  bytes subjects, `(?i)` flags, `\p{Greek}`, raw-string patterns, and all five
  error paths — produces byte-identical output under both binaries.
- `elps doc -p regexp` and `elps doc regexp-match?` byte-identical.

## Verification

| Gate | Result |
|---|---|
| `go build ./...` | ok |
| `go test ./...` | ok (36 packages) |
| `make test` (go tests + `_examples`) | ok |
| `make race` | ok |
| `make test-elpscheck` | ok |
| `./elps doc -m` | All functions and exports are documented |
| `./elps fmt -l --exclude 'grammar' ./...` | empty |
| `make ci-gates-test` | 53 passed, 0 failed |
| `./elps lint --workspace=. --exclude 'grammar' --include '_examples' ./...` | 9 diagnostics — byte-identical to a binary built from `origin/main` |
| gofmt-unclean files | 32 on `main`, 32 here; no new ones |

## Deliberately not touched

`scripts/testdata/benchstat-{clean-ci,noisy-sandbox,bps-dip-only}.txt` each carry
a `pkg: github.com/luthersystems/elps/parser/regexparser` line. These are frozen
captures of real benchstat output used as fixtures for the benchmark-gate
self-test — their value is provenance, so editing them to erase a package would
falsify the record and perturb the gate's expected results. `make ci-gates-test`
is green with them unchanged.

---


sam-at-luther:
## The `utils_test.lisp` minify caveat is already fixed — on #311

This PR's description flags one file whose `elps minify` output differs between the before and after binaries, correctly diagnosed as **pre-existing run-to-run nondeterminism on `main`** rather than a behaviour change from this deletion. That diagnosis is right, and the cause is already known and fixed elsewhere.

Verified just now, same input file, 8 consecutive runs each:

| tree | distinct outputs |
|---|---|
| `main` (`919ef2d`) | **5** — `15bd583e`, `24c8d26c`, `58f2d93d`, `91feb7f2`, `e7ea783e` |
| #311 (`fde78f3`) | **1** — `91feb7f2`, stable across all 8 |

Root cause, found by round-1 fuzzing on #311 from the input `(())(defun 2(e'e))`: `applyAssignments` ranged over a **map** and wrote through AST nodes that two symbol records can share, so the renaming depended on Go's randomised map iteration order. Fixed there, with 300/300 identical runs after.

So no issue needs filing for it, and nothing here needs changing. Two consequences worth noting:

1. **This PR is clean.** The one non-identical file is explained entirely by a pre-existing defect on the base, not by removing `regexparser` — and the 11 files `genshirocore` actually consumes were byte-identical, md5 `0cd4472714…`, stable across 5 re-runs and matching the `shirocore.lisp` currently committed in substrate.

2. **It raises the priority of #311.** The minifier's output *is* the chaincode substrate ships. Until that fix lands on `main`, `elps minify` is nondeterministic for at least some inputs, which means reproducible builds are not reproducible and a symbol map cannot be trusted to decode a trace.

On the unresolved question in the report: I could not confirm whether the "1 high vulnerability on the default branch" push warning refers to the `goparsec` pin either — the Dependabot alerts API isn't reachable from this session. Worth a look when someone with repo access reviews this, since removing an untagged 2021 dependency is a plausible way to clear it.

---

sam-at-luther:
## `Benchmark Comparison` is red for a structural reason, not a defect in this PR

The job failed with **exit 2 — "the benchmark comparison could not be interpreted"**, not exit 1 ("regressions detected"). Those are different things: exit 2 means the gate could not read its input at all.

Cause: this PR deletes `parser/regexparser/bench_test.go` along with the package. On `main` the benchmark gate compares against a **baseline restored from `actions/cache`**, recorded on an earlier `main` push — and that cached baseline contains `pkg: github.com/luthersystems/elps/parser/regexparser` rows that the PR arm no longer produces. benchstat is handed two arms with different benchmark sets and emits nothing the gate can adjudicate.

So the gate is behaving correctly given its input. The input is the problem.

### This is exactly the design flaw #311 fixes

#311 replaces the cached baseline with **both arms measured in one job on one runner** — PR head and PR base checked out side by side and benchmarked alternately. Under that design, deleting a benchmarked package removes it from *both* arms simultaneously and there is nothing to mismatch.

That change was made for a different symptom (consecutive runs of identical code landing on an EPYC 7763 and an EPYC 9V74, benchstat refusing to pair them, gate flapping red then green with no code change). Deleting a package is a second, independent way the same cached-baseline design produces a false red. Worth recording, because it means the fix is load-bearing for more than the one case it was written for.

### Recommended order

**Merge #311 first, then re-run this board.** That gets the dual-arm gate onto `main`, after which this PR compares like-for-like and should go green without any change here.

Failing that, the three non-blocking options are: re-run after #311 lands, temporarily rebase this onto #311, or merge on the strength of the other six green checks with this noted. I would not weaken the gate to accommodate it — a gate that cannot read its input must never report success, and that property is why the failure is legible at all.

### Everything else on this board is green

`Lint & Test`, `Build (Windows)`, `CI Gate Self-Test`, `govulncheck`, and both Socket checks pass. The corpus evidence in the PR body stands: `elps fmt` byte-identical across 93 substrate `.lisp` files, and the 11 files `genshirocore` consumes byte-identical and matching the `shirocore.lisp` currently committed in substrate.

---

PR #322, opened 2026-08-08
fix(debugger): make the pause tests actually assert, and ack the DAP pause after arming

sam-at-luther:
## The finding: these two tests have never asserted anything

`TestEngine_RequestPause` and `TestDAPServer_PauseRequest` fail intermittently under `make race`. Six agents have now burned effort proving it wasn't their change.

The headline is not the flake. It is that **both tests have been asserting nothing about a running program for their entire life.**

Both ran `(countdown 100000)`. With a debugger attached, tail-call optimisation is disabled globally (`lisp/env.go:1307`: `if env.Runtime.Debugger == nil { npop = env.Runtime.Stack.TerminalFID(...) }`), so every turn of the loop costs a *physical* stack frame and the program aborts at the 25,001-frame cap instead of counting down. Measured on this branch (`go test -race`, `lisp/x/debugger`):

| n | elapsed | result |
|---|---|---|
| 100000 | 1.78s | `physical stack height exceeded maximum: 25001` |
| 25000 | 0.93s | `physical stack height exceeded maximum: 25001` |
| 2000 | 0.21s | `0` (completes) |
| 200 | 0.01s | `0` (completes) |

`n=100000` and `n=25000` are indistinguishable — both just hit the cap. **Both tests discarded the result**, so this passed silently. The "flake" was a 5-second deadline racing a program that was never going to finish.

The failure mode is unanimous with that reading. Across **214 reproduced failures** in two independent stress runs under deliberate CPU load, every single one landed on the final `select` deadline:

```
  84  engine_test.go:1049: timeout waiting for eval result
 130  handler_test.go:1724: timeout waiting for program to finish
```

**Zero** landed on any pause assertion or stop-reason assertion. `DATA RACE` / `panic:` count across ~1600 race-instrumented runs: **0**. It is a wall-clock problem, not a synchronisation defect.

## The fix

1. **Bound the work, don't widen the wait.** ~200 turns that actually complete. The deadlines are untouched at 5s.
2. **Add a blocking gate.** Shrinking alone trades one flake for a rarer one — a short program can finish before a starved test goroutine reaches `RequestPause`. A `(debug-test-gate)` builtin (test-only, installed via `env.AddBuiltins`) parks the eval goroutine until the test releases it, so the pause is armed while the program is *provably* mid-flight rather than probably mid-flight. `EvalCount()` polling and the `time`-based resume loop are gone.
3. **Assert the result.** `require.Equal(t, lisp.LInt, res.Type)` + `assert.Equal(t, 0, res.Int)`, so a future bump back over the frame cap fails loudly instead of silently.

## Separate, user-visible: `onPause` acknowledged before arming

`handler.onPause` sent the DAP success response **before** calling `engine.RequestPause()`:

```go
h.send(resp)
h.engine.RequestPause()
```

This is a real product bug independent of the tests. A DAP client (VS Code) that receives a successful `pause` response is entitled to assume the request has taken effect. With the old ordering, a short-lived program can run to completion and terminate in the window between the reply and the arming — the client is told the pause succeeded and then watches the program finish without ever stopping. Arm first, acknowledge second.

It also restores the happens-before the DAP test now relies on: the test releases the gate only after reading a successful `PauseResponse`. Outbound writes are already serialised by `Server.send`'s mutex, so reordering is safe.

## Stress numbers

Interleaved A/B — old and new binaries alternating inside **one** load window, arm order swapped on alternate iterations, so neither arm gets a quieter machine. 200 iterations per arm, `-race`, busy-spin load on a 4-core box (raised from 8 to 16 loaders early in the run; because the arms interleave, both saw it equally). Peak load average 22. Total wall 2910s.

| arm | failures | mean wall per invocation |
|---|---|---|
| `TestEngine_RequestPause` — **before** | **84 / 200 (42%)** | 5312 ms |
| `TestEngine_RequestPause` — **after** | **0 / 200** | 1554 ms |
| `TestDAPServer_PauseRequest` — **before** | **130 / 200 (65%)** | 5849 ms |
| `TestDAPServer_PauseRequest` — **after** | **0 / 200** | 1531 ms |

(The "after" means are dominated by race-binary process startup under saturation, not by the tests.) In-test time, same binaries, three consecutive runs each:

| test | before | after |
|---|---|---|
| `TestEngine_RequestPause` | 1.24s / 1.18s / 1.28s | 0.02s / 0.03s / 0.03s |
| `TestDAPServer_PauseRequest` | 1.26s / 1.17s / 1.24s | 0.05s / 0.05s / 0.03s |

Margin against the untouched 5s deadline goes from ~4× to ~150×.

A prior non-interleaved run at 10-way load gave the same picture: 40 iterations → 9/40 engine and 11/40 DAP failures, again all at the final deadline.

## Sibling tests with the same shape

Reported, not fixed here (keeping this PR small):

- `TestEngine_ExceptionBreakpoint` (`engine_test.go` ~L373) and `TestEngine_FunctionBreakpoint` (`engine_test.go` ~L1137) both wait on `e.IsPaused()` and then **immediately sample `stopReason`**, which is written by the event callback. `WaitIfPaused` publishes `pausedEnv` (what `IsPaused` reads) *before* invoking the callback, so the reason can lag the paused flag by a scheduling quantum. `TestEngine_RequestPause` already carried a comment describing exactly this and worked around it with `require.Eventually`; these two did not get the workaround. Same class: waiting on one signal while asserting on another.
- No other test in `lisp/x/**` pairs a hard wall-clock deadline with a heavy program — every other `countdown` in these suites is 3–5 turns. The two `time.Sleep` sites (`engine_test.go` ReadyCh delay, `handler_test.go` `dialWithRetry`) are benign.
- `Engine.EvalCount()` now has no callers. It is exported API on an experimental package, so it is left in place; worth a look if `lisp/x/debugger` ever gets an API sweep.

## Verification

`go build ./...` · `go test ./...` · `make race` · `make test-elpscheck` · `golangci-lint run ./...` **0 issues** · `./elps doc -m` · `./elps fmt -l --exclude 'grammar' ./...` · `make ci-gates-test` (53 passed) · `./elps lint --workspace=. --exclude 'grammar' --include '_examples' ./...` byte-identical to `919ef2d` · no new gofmt-unclean files (34 pre-existing on main, unchanged). No struct fields reordered in `lisp/x/` (`fieldalignment` is exempt there by `.golangci.yml`).

### DIFF
--- a/lisp/x/debugger/dapserver/handler.go
+++ b/lisp/x/debugger/dapserver/handler.go
@@ -1039,10 +1039,16 @@ func (h *handler) applyLaunchConfig(cfg launchConfig) {
 }
 
 func (h *handler) onPause(req *dap.PauseRequest) {
+	// Arm the engine BEFORE acknowledging. A client that receives a
+	// successful pause response is entitled to assume the request has taken
+	// effect; replying first leaves a window in which a short-lived program
+	// can run to completion and terminate without ever stopping, even though
+	// the client was told the pause succeeded.
+	h.engine.RequestPause()
+
 	resp := &dap.PauseResponse{}
 	resp.Response = h.newResponse(req.Seq, req.Command)
 	h.send(resp)
-	h.engine.RequestPause()
 }
 
 func (h *handler) onSource(req *dap.SourceRequest) {
--- a/lisp/x/debugger/dapserver/handler_test.go
+++ b/lisp/x/debugger/dapserver/handler_test.go
@@ -9,11 +9,13 @@ import (
 	"net"
 	"os"
 	"strings"
+	"sync"
 	"testing"
 	"testing/fstest"
 	"time"
 
 	"github.com/google/go-dap"
+	"github.com/luthersystems/elps/elpsutil"
 	"github.com/luthersystems/elps/lisp"
 	"github.com/luthersystems/elps/lisp/lisplib"
 	"github.com/luthersystems/elps/lisp/x/debugger"
@@ -452,6 +454,54 @@ func newDAPTestEnv(t *testing.T, dbg *debugger.Engine) *lisp.LEnv {
 	rc = env.InPackage(lisp.String(lisp.DefaultUserPackage))
 	require.True(t, rc.IsNil(), "InPackage failed: %v", rc)
 	return env
+}
+
+// pauseGate is a rendezvous between the test goroutine and the eval
+// goroutine. newPauseGate installs a `(debug-test-gate)` builtin in env; the
+// program under test calls it, which parks the eval goroutine until the test
+// calls Release. That lets a test drive a DAP pause while the program is
+// provably mid-flight, instead of racing a short program with a sleep or a
+// poll on EvalCount. (Mirrors the helper in the debugger package — test
+// helpers do not cross package boundaries.)
+type pauseGate struct {
+	entered     chan struct{}
+	release     chan struct{}
+	enteredOnce sync.Once
+	releaseOnce sync.Once
+}
+
+// newPauseGate installs the gate builtin into env's current package.
+func newPauseGate(t *testing.T, env *lisp.LEnv) *pauseGate {
+	t.Helper()
+	g := &pauseGate{
+		entered: make(chan struct{}),
+		release: make(chan struct{}),
+	}
+	// Never leave the eval goroutine parked if the test fails before it
+	// reaches its own Release call.
+	t.Cleanup(g.Release)
+	env.AddBuiltins(true, elpsutil.Function("debug-test-gate", lisp.Formals(),
+		func(env *lisp.LEnv, args *lisp.LVal) *lisp.LVal {
+			g.enteredOnce.Do(func() { close(g.entered) })
+			<-g.release
+			return lisp.Nil()
+		}))
+	return g
+}
+
+// Release unparks the eval goroutine. Safe to call more than once.
+func (g *pauseGate) Release() {
+	g.releaseOnce.Do(func() { close(g.release) })
+}
+
+// WaitEntered blocks until the program reaches the gate.
+func (g *pauseGate) WaitEntered(t *testing.T) {
+	t.Helper()
+	select {
+	case <-g.entered:
+	case <-time.After(5 * time.Second):
+		t.Fatal("program never reached debug-test-gate")
+	}
 }
 
 func sendDAPRequest(t *testing.T, w io.Writer, msg dap.Message) {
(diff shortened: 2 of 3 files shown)


sam-at-luther:
## The two sibling tests flagged here are already fixed — on #311

This PR reports `TestEngine_ExceptionBreakpoint` and `TestEngine_FunctionBreakpoint` as siblings with the same shape (wait on `IsPaused()`, then immediately sample `stopReason`). That is **correct against `main`**, which is this PR's base — but both were already fixed on #311, so no separate follow-up is needed.

Verified by counting `return e.IsPaused()` inside each function body on both trees:

| test | `main` (`919ef2d`) | #311 (`fde78f3`) |
|---|---|---|
| `TestEngine_ExceptionBreakpoint` | 1 | **0** |
| `TestEngine_FunctionBreakpoint` | 1 | **0** |

#311's commit `05cdedc` converted six tests with that shape to wait on the state they actually assert, and — more to the point — fixed the underlying cause: `Engine.IsPaused()` reported a pause *before* the `EventStopped` callback had run, so any caller polling it and then reading callback-written state had no happens-before with the write. `PausedUnannounced()` was added there for the one consumer that genuinely needs to distinguish "paused" from "paused and announced" (dapserver's late-connect path, which could otherwise emit a duplicate stopped event).

### Merge interaction to expect

This PR is based on `main` and #311 also edits `lisp/x/debugger/engine_test.go` and `dapserver/handler.go`, so **whichever lands second will conflict**. The conflicts should be small and non-semantic — #311 rewrites the `Eventually` predicates in the six converted tests; this PR rewrites the *program* those tests run and reorders `onPause`. Different lines, same files. Worth knowing before someone hits it cold.

Note the two changes are complementary rather than overlapping: #311 fixed a **synchronisation** defect (a signal observed too early), while this PR fixes a **wall-clock** one (a program that could never finish inside the deadline). The evidence here is what separates them — 20/20 captured failures at the final `select` deadline, **zero** at any pause or stop-reason assertion. If this had been the same bug as #311's, the failures would have landed on the assertions.

### Also noted

`Engine.EvalCount()` now has no callers. Genuinely dead once the racy idiom it served is gone — reasonable to remove, though not in this PR's scope.

---

sam-at-luther:
## `Benchmark Comparison` red — same structural cause, this PR is clean

Exit **2 — "could not be interpreted"**, not exit 1 "regressions detected". The gate could not read its input; it is not reporting a slowdown.

This PR touches two debugger test files and one handler function. It adds, removes and renames no benchmarks. The benchstat output in the job log ends with a geomean and `need >= 6 samples for confidence interval at level 0.95` and **no `vs base` column** — the signature of benchstat rendering a single arm because it declined to pair the two, which is the CPU-model mismatch already documented on #311: the cached baseline and the PR arm land on different models from GitHub's heterogeneous `ubuntu-latest` fleet (EPYC 7763 vs EPYC 9V74 observed), benchstat keys its configuration off the `cpu:` header, and refuses to compare.

### Two PRs, two different non-code failures, same root design

| PR | Change | Why the gate failed |
|---|---|---|
| #321 | deletes `parser/regexparser` (and its `bench_test.go`) | cached baseline still contains rows the PR arm cannot produce |
| **#322** (this) | debugger tests + one handler | cached baseline recorded on a different CPU model |

Neither is a code defect, and neither is fixable inside its own PR. Both come from the same design: **comparing against a baseline recorded on a different machine at a different time**, which attributes hardware and repo-shape deltas to the code delta.

### #311 fixes both

It replaces the cached baseline with **both arms measured in one job on one runner** — PR head and PR base checked out side by side, benchmarked alternately at n=10. Same silicon for both arms kills the mismatch here; a deleted package vanishing from both arms simultaneously kills #321's.

**Recommended: merge #311 first, then re-run this board.** Expected to go green with no change to this PR.

### Everything else here is green

`Lint & Test`, `Build (Windows)`, `CI Gate Self-Test`, `govulncheck`, both Socket checks.

---

PR #323, opened 2026-08-08
fix(lisp): fuzz the evaluator and repair 5 defects it found

sam-at-luther:
## Base

**This PR targets `claude/elps-fieldalignment` (PR #311), not `main`** — it is stacked on the round-1 fuzzing infrastructure (`scripts/fuzz.sh`, `internal/fuzzseed`, `.github/workflows/fuzz.yml`, `make fuzz`) that #311 introduces.

Scope is the **evaluator only**. The builtins/stdlib and parser fuzzing work are separate PRs.

## Why

The existing suite stops at parsing. Nothing fuzzed `lisp` evaluation — the interpreter that runs customer-uploaded phylum as Fabric chaincode in `luthersystems/substrate`, where a panic is a chaincode crash and a non-terminating evaluation is a wedged transaction.

## The harness (`lisp/eval_fuzz_test.go`)

Native fuzzing has no per-input deadline and evaluation is Turing-complete, so a naive target hangs on the first `(defun f () (f))` the mutator produces. Three things make the target work:

1. **A budget.** Every evaluation runs under `WithMaxSteps`, `WithMaxTailIterations`, `WithMaximumPhysicalStackHeight`, `WithMaxAlloc`, `WithMaxMacroExpansionDepth` and a context deadline. `MaxHeightLogical` is deliberately left off: its unit is *elided tail frames*, a running total that varies with the shape of a loop body, so it fires data-dependently on correct constant-space loops. `MaxTailIterations` bounds the same runaway shape in honest units.

2. **A watchdog.** Each evaluation runs on its own goroutine; the test fails if it has not returned in 30s. The budgets can only stop code that reaches a limit check, so the watchdog is what turns "this terminates" into an assertion. It is what caught defect A5.

3. **`lisp.IsInternalPanic(result) == false`.** This is the single most important line. `env.eval` recovers every Go panic into an ordinary-looking `*LVal`, so "the process survived" proves nothing whatsoever — without this assertion, A1–A3 all come back as values and the target passes. The marker is the recovered Go-stack snapshot, not the condition name. `TestInternalPanicMarkerIsNotForgeable` pins both directions next to the harness that depends on them: `(error 'internal-panic "forged")` must **not** satisfy it, and a genuine recovered panic must. An explicitly-named `handler-bind ([internal-panic ...])` *does* intercept a real panic by design, so fuzzer input containing such a handler can legitimately mask a defect — documented as a known gap.

Environments are rebuilt per input (a crasher that depends on prior inputs is worthless) and `Runtime.Library` is left nil, so `load-file` errors instead of reading the test machine's disk. "Did this parse?" is decided by running the reader, not by pattern-matching the error text — a heuristic would silently discard an *evaluation* result whose message happens to carry the reader's wording (e.g. a program that evaluates `(load-string "(")`), and discarding inputs is how a fuzz target goes quiet.

## Seeds (`internal/fuzzseed/evalseed.go`)

Hand-written, and deliberately **not** seeded from `LispSources()` — eval seeds are *executed*, so handing the mutator the repository's `.lisp` test suite throttles every generation descended from it. Split by expected outcome:

- `EvalRunaway` — must be stopped by a budget.
- `EvalTerminating` — must complete **without error** under the same budget. This is the false-positive guard: an infinite ELPS loop the budget stops is CORRECT; a bounded program that trips a budget is a defect.
- `EvalAdversarial` — no expected outcome; good starting points for the mutator.

Both previously-known seed bugs are fixed and documented in place:

- `dotimes-huge` was written `(dotimes ([i 1000000000]) i)`. `[` and `(` are interchangeable, so the control sequence read as `((i 1000000000))` and bailed on a **syntax error** before consulting any budget — it passed for the wrong reason. Written correctly as `(dotimes (i 1000000000))` it immediately found A5.
- `deep-nested-progn` was misclassified in `EvalTerminating` (it actually returns "physical stack height exceeded maximum"). The terminating assertion now checks **both** that evaluation finished *and* that `result.Type != LError`; previously only the first was enforced, so the split held in one direction only.

Three new guards in `internal/fuzzseed/fuzzseed_test.go` keep the corpus honest: the two corpora must be disjoint (a program cannot be required both to complete and to be stopped), runaway seeds must not be blank, and adversarial seeds must be distinct.

### A1 — `#^(%555555591)`: unbounded allocation, unrecoverable OOM abort

`opExpr` read the positional-placeholder index with `strconv.Atoi` straight from source and used it as the formals-slice length. Measured: **136 bytes per formal**, so 14 bytes of source asks for ~75GB. Demonstrated under a 2GB address-space cap:

```
runtime: out of memory: cannot allocate 801112064-byte block
fatal error: out of memory
```

A `runtime.throw` — `recover()` cannot intercept it and no `handler-bind` can contain it.

**Not fixed with `Runtime.CheckAlloc`.** Measured on this tree, that knob is ~130x miscalibrated here and unfixably so:

| measurement | result |
|---|---|
| `#^(list %1000000)` | 129.7 MB (136 B/formal) |
| `#^(list %10485760)` under the default `MaxAlloc` (10,000,000) | **accepted**, ~1.4 GB |
| 1 / 2 / 4 / 6 retained `#^(list %2000000)` forms | 259.4 / 518.8 / 1037.6 / **1556.4 MB** — `CheckAlloc` is per-operation, not cumulative, so the OOM stays reachable at ~6 tokens |
| `WithMaxAlloc(8)` on `#^(list %9)` (an ordinary nine-argument lambda) | a `CheckAlloc` guard rejects it — the knob spans incompatible units |
| `WithMaxSteps(10)` + 100ms deadline on `#^(list %8000000)` | returned after **2.45s** of uninterruptible work |

Fixed instead with a **dedicated bound, `lisp.MaxExprFormals = 1024`**, checked where the integer is parsed. 1024 is justified against real programs rather than memory: C guarantees only 127 parameters (C99 5.2.4.1) and the JVM caps a method at 255, so 4–8x the most permissive mainstream limit cannot plausibly reject a program a human wrote, while capping this site at ~136KB. It is a constant, not a `Runtime` knob, because it bounds a syntactic construct. **Plus** a `checkLimits` inside the formals loop, so the loop is step- and context-aware regardless of what the bound is later set to.

### A2 — `#^%-1`: `make([]*LVal, -1)`

Same site, **separate guard** — a cap only rejects values *above* it. Reached through `countExprArgs`' bare-symbol branch; surfaced as an uncatchable `internal-panic: makeslice: len out of range`. Both checks live in one `parseExprArgIndex` helper so a third placeholder syntax cannot be added without them. (The s-expression branch did not panic on a negative index — it silently ignored it; that is now rejected too.)

### A3 — `(append 'bytes 0)`: `LVal.Bytes()` panic

`builtinAppend` validated the type *specifier*, then handed the arg list to `builtinAppend_Bytes`, whose comment claimed the sequence "has already been validated". It had not. Mirrors the guard `append-bytes` already had; the lying comment is replaced with the reason it is there.

### A4 — `(debug-print)` with zero args wrote to `os.Stdout`

Via `fmt.Println`, ignoring `Runtime.Stderr` — which its docstring promises and its non-empty branch honours. For a host whose stdout is a protocol stream (the LSP server, a JSON-RPC chaincode process) that is a corrupted stream, not cosmetic noise. The zero-argument branch was entirely redundant (`fmt.Fprintln(w)` already writes exactly the newline), so removing it is the whole fix. Confirmed that reverting it leaves everything outside `./lisp` green — hence the explicit `{"(debug-print)", "()", "\n"}` row.

### A5 — `(dotimes (i 1000000000))` reached no limit check at all — **new, found by the watchdog**

`opDoTimes` drives its own Go loop over a program-supplied count. With a body, the body's `Eval` reaches `checkLimits` and the loop is bounded as a side effect. With **no body** there is nothing to evaluate, so nothing checked anything: it ignored a 2s context deadline and a 2,000,000-step budget alike and was still running 30 seconds later. This is exactly the seed that "passed for the wrong reason" before. Fixed with a `checkLimits` at the top of each iteration.

A sweep of the other Go-level loops in `lisp/op.go` found no further instances: every other loop iterates over `.Cells`, bounded by parsed source size. (`libstring`'s `string:repeat` takes a program-supplied count but is already correctly guarded.)

## Mutation verification

Each fix reverted individually; its test must fail; then restored. Tree confirmed clean afterwards.

| Defect | Mutation | Test | Result |
|---|---|---|---|
| A1 | drop the `MaxExprFormals` upper bound | `TestSpecialOp` (`#^ argument index bounds`), `TestExprFormalsBoundIsInclusive` | **FAIL** as expected |
| A2 | drop the negative-index check | `TestSpecialOp`, `FuzzEval/panic-expr-negative-placeholder`, `TestEvalAdversarialSeedsSurvive` | **FAIL** — `internal-panic: makeslice: len out of range` |
| A3 | drop the `LBytes` guard | `TestBytes` (`append 'bytes rejects a non-bytes sequence`), `FuzzEval/panic-append-bytes-non-bytes` | **FAIL** — `internal-panic: not bytes: int` |
| A4 | restore the `fmt.Println()` branch | `TestDebugPrint` | **FAIL** — `expected debug output "\n" (got "")`; `go test` for every package except `./lisp` stays **green**, confirming nothing else covers it |
| A5 | drop the `checkLimits` in `opDoTimes` | `TestDoTimesHonoursStepBudget`, `TestDoTimesHonoursContextDeadline` | **FAIL** as expected |

## Gates

All run locally on a 4-core linux/amd64 box, on the final tree.

| Gate | Result |
|---|---|
| `go build ./...` | clean |
| `go test ./...` | pass (28.8s) |
| `make race` | pass — exit 0, 30 packages ok, **0 data races** (1m36s) |
| `make test-elpscheck` | pass (33.6s) |
| `golangci-lint run --build-tags elpscheck ./lisp/...` | **0 issues** |
| `make ci-gates-test` | **75 passed, 0 failed** |
| `./elps lint --workspace=. --exclude 'grammar' --include '_examples' ./...` | **9 findings — identical to `main`** (verified by running the same binary against a `main` worktree; zero `.lisp` files are touched by this PR) |
| `gofmt -l` | **no new unclean files** (diffed against the PR base — the 17 unclean files are all pre-existing) |

## Measured CI cost

Fuzz targets go from 11 to 12; discovery in `scripts/fuzz.sh` is dynamic, so `FuzzEval` is picked up with no registration.

- **PR path** (`FUZZTIME=30s`, all 12 targets): **5m44s wall**, of which **FuzzEval is 32s** — that is the marginal cost added to every PR.
- **Throughput**: ~3,500 exec/sec (fresh env + full stdlib per input).
- **8-minute discovery run**: **1,700,201 execs, 1,071 corpus entries, 0 failures.**
- **Nightly** (`10m`/target): 10 targets actually fuzz (`FuzzGateSelfTest` is inert unless armed, `FuzzParseLVal` skips), so ~100 min + build. `timeout-minutes` raised **120 → 140** to keep the backstop above the new target count; the comment now records the measured numbers and says to re-check it when a target is added.

## Regression corpus

Five minimised inputs committed to `lisp/testdata/fuzz/FuzzEval/`, replayed by plain `go test`:
`panic-expr-negative-placeholder`, `oom-expr-huge-placeholder`, `oom-expr-huge-placeholder-call`, `panic-append-bytes-non-bytes`, `hang-dotimes-empty-body`.

## Not verified

- The *mutator itself* found no new crash in the 8-minute run. Every defect above came from the hand-written seed corpus and the classification tests. The nightly 10m budget plus corpus carry-over is what will exercise the mutator properly over time.
- The OOM abort for A1 was demonstrated at `#^(%100000000)` under a 2GB address-space cap; the literal `#^(%555555591)` was never run unbounded, since it would abort the machine. The 136 B/formal figure it extrapolates from is measured.
- Timing figures are from this sandbox (4 cores, 16GB); GitHub runners will differ, which is why `timeout-minutes` keeps a wide margin.
- `MaxExprFormals = 1024` is argued from language precedent (C99, JVM) and from the measured per-formal cost, not from a survey of real ELPS corpora — no phylum source was available here to confirm nothing in production declares more than 1024 parameters.


sam-at-luther:

```yaml
on:
  pull_request:
    branches:
      - main
```

This PR targets `claude/elps-fieldalignment` (stacked on #311 for the fuzzing infrastructure), so `Lint & Test`, `Fuzz targets`, `CI Gate Self-Test`, `govulncheck`, `Build (Windows)` and `Benchmark Comparison` **never fired**. The only checks present are the two Socket ones, which are a GitHub App and not branch-filtered.

---

Issue #314, opened 2026-08-08
time:sleep is unbounded by every containment budget (292-year block reachable from phylum source)


sam-at-luther:
Found while building the stdlib fuzz targets (PR #311). Excluded from fuzzing by name rather than fixed, because it is a design decision rather than a bug in a call site.

## The problem

`time:sleep` blocks for a caller-supplied duration **inside a single evaluation step**. None of the interpreter's containment mechanisms can bound it:

| Mechanism | Why it does not apply |
|---|---|
| `Runtime.MaxSteps` | The sleep happens within one step; no step boundary is crossed |
| `Runtime.MaxAlloc` | Allocates nothing |
| `CallStack.MaxHeightPhysical` / `MaxTailIterations` | No frames, no tail calls |
| `LEnv.EvalContext` deadline | Only observed *between* steps |

A phylum can reach `time.Duration(math.MaxInt64)` by parsing `"9223372036854775807ns"`, which blocks for roughly **292 years**.

## Why it matters here

In `luthersystems/substrate`, phylum source is customer-uploaded and executed as Hyperledger Fabric chaincode. Every other unbounded-execution shape found by fuzzing was containable by one of the budgets above; this one is not, which makes it the only known way to wedge an evaluation that no embedder configuration can stop.

## What is NOT the argument

"A phylum can equally write an infinite loop." It can — and the step budget stops it. That is the difference: every other runaway shape is bounded by a knob the embedder controls, and this one is bounded by nothing.

## Possible directions (not a recommendation)

1. Cap the sleep at the remaining context deadline — sleep on `min(d, time.Until(deadline))` and return a budget error if truncated. Preserves the API, makes the context deadline actually authoritative.
2. Make the sleep interruptible: `select` on a timer and `ctx.Done()`.
3. Cap at a configurable `Runtime.MaxSleep`, defaulting to something small, with `0` meaning "unbounded" for embedders who want today's behaviour.
4. Decide that embedders who expose `libtime` to untrusted code must set an external deadline, and document that explicitly.

(1) and (2) compose and look cheapest, but any of these is language-visible for a phylum that currently sleeps legitimately, so it wants a deliberate call rather than a drive-by fix.

## Notes

- The exclusion is recorded in the fuzz target source with this rationale, so re-including it is a one-line change once a direction is chosen.
[...]

Issue #320, opened 2026-08-08
Round-2 fuzzing: 13 defects found and fixed, work lost before push — durable record for redo


sam-at-luther:
Round-2 fuzzing (evaluator, builtins/stdlib, grammar-aware) found and fixed **13 defects**, and two adversarial reviews found **2 more defects in those fixes**. That work was completed and verified but lived only in a sandbox that was reset before it was pushed. It is lost as code.

This issue is the durable record so it can be redone directly rather than rediscovered. Every item below was reproduced and verified; minimised inputs are exact.

Tracking PRs will be small and focused — see the checklist at the end.

---

## A. Evaluator defects (all fixed, all lost)

**A1. `#^(%555555591)` — unbounded allocation, unrecoverable OOM abort.**
`opExpr` reads the positional-placeholder index with `strconv.Atoi` straight from source and uses it as the formals-slice length. 555M `*LVal` → gigabytes → process abort no handler can catch.

**A2. `#^%-1` — `make([]*LVal, -1)`.** Same site, separate guard needed: a cap only rejects values *above* it. Reaches the allocation via `countExprArgs`' bare-symbol branch. Surfaces as an uncatchable `internal-panic`.

⚠️ **The first fix for A1/A2 was judged NEEDS-WORK on review — do not reimplement it as written.** It used `Runtime.CheckAlloc`, which is ~130× miscalibrated for this site: `#^(list %10485760)` was still **accepted**, allocating **1363.8 MB in 3.22 s** from 20 bytes of source. Every other `CheckAlloc` site bounds a byte buffer (10 MB at the cap) or a slice of pointers you already hold; this site materialises a fresh `LVal` + `fmt.Sprintf` string per unit (~136 B each) from one integer literal. Also, the OOM stays reachable at ~6 tokens because `CheckAlloc` is explicitly per-operation, not cumulative (measured: 259.7 → 519.1 → 1037.9 → 1556.8 MB across 1/2/4/6 retained `#^(list %2000000)` forms). And `opExpr`'s formals loop is step- and context-blind — `WithMaxSteps(10)` + `#^(list %8000000)` returns after **2.742 s** of uninterruptible work.
**Do instead:** a dedicated bound sized for this site (≈1024 formals, justify the number) plus a `checkLimits` inside the loop. Note `WithMaxAlloc(8)` currently refuses `#^(list %9)` — a nine-argument lambda — which shows the shared knob spans incompatible units.

[...]

PR #324, opened 2026-08-08
Fuzz the builtins and stdlib: a non-terminating pow, an uninterruptible dotimes, and a whole class of libschema crashes

sam-at-luther:
Stacked on #311 — this branch is based on `claude/elps-fieldalignment` and uses the round-1 fuzzing infrastructure it carries (`scripts/fuzz.sh`, `internal/fuzzseed`, `.github/workflows/fuzz.yml`, `make fuzz`). Review #311 first; this PR targets that branch, not `main`.

Round 1 fuzzed **source**: the parser, lexer, formatter, minifier, plus libjson's decode path. That left the whole callable surface unfuzzed against its **arguments** — every builtin, special operator, macro and stdlib export. Downstream in `luthersystems/substrate` a phylum is customer-uploaded and runs as Hyperledger Fabric chaincode, so a panic or a hang in any of them is remotely triggerable from untrusted input, exactly like a parser crash.

## What was added

`internal/fuzzval` turns a fuzzer byte string into typed `LVal`s, including shapes no ELPS source can spell: `LNative` wrapping a Go value, multi-dimensional `LArray`, symbol-keyed sorted-maps, `LError` in argument position, `LQuote`, user lambdas (`Builtin == nil`) alongside Go builtins, and the shared `Nil()`/`Bool()` singletons.

Three targets (11 → 14):

| Target | Surface |
|---|---|
| `lisp/lisplib.FuzzApplyStdlib` | All 240 callables the library registers, each through its real calling convention — `FunCall`, `SpecialOpCall` or `MacroCall`, so special operators and macros get the unevaluated fragments they are actually handed |
| `libjson.FuzzDumpJSON` | The **encode** direction. `FuzzLoadJSON` only ever hands `Dump` a value `Load` produced, and `Load` emits six shapes; tagged values, natives, multi-dimensional arrays, integers, byte slices and error values had never been encoded |
| `libschema.FuzzSchemaValidate` | Schema **composition** — nested constraint slots, with the subject bound as a Go value so unspellable shapes reach the validators |

Invariants: no panic escapes, non-nil result, `String()` terminates, the three shared singletons are bit-identical afterwards, and the call terminates. Bounded three ways because no single bound sees every loop — a context deadline, a step limit, and an out-of-band watchdog for builtins that loop in Go without evaluating anything (`checkLimits` is per-evaluation, so those consult nothing). `MaxAlloc` is set deliberately low (4096) so the fuzzer explores the allocation guard rather than OOM-ing the runner; a builtin refusing an oversized allocation is correct, not a crash.

Deliberately not asserted: reflexivity (`LVal.Equal` is false for `LError`/`LFun`/`LQuote`/`LBytes`/`LNative` even against the same object, and `(= NaN NaN)` is false), and byte-identity of a JSON re-encode against the first encode (the known >2^53 rounding, plus `encoding/json` replacing invalid UTF-8 with U+FFFD — both are lossy on the way *in*, so the comparison is made on the far side, second decode against first).

`time:sleep` is excluded by name with a source comment: it blocks for a caller-supplied duration inside one step and no interpreter budget can bound it (#314). Out of scope here.

All three targets survived a 240s fuzz run each after the fixes (`fuzz.sh: 3 target(s), 0 failure(s)`); `FuzzApplyStdlib` reached 678k execs.

### B1 — `(pow -128 9223372036854775807)` never terminates

`powInt` doubled an exponent accumulator: `n` goes 1, 2, …, 2^62, then 2^63 wraps to `MinInt`, then `MinInt*2` wraps to exactly **0**, and `2*0 < b` is true forever. Zero is a true fixed point. The loop allocates nothing and evaluates nothing, so `MaxAlloc`, `MaxSteps` **and** a context deadline are all blind to it — none is consulted inside a builtin.

Found independently by `FuzzApplyStdlib`, minimised to `(pow -9223372036854775808 9223372036854775806)`.

Replaced with binary exponentiation (≤63 iterations). **No answer changes**: Go's `int` multiplication is arithmetic mod 2^64, a commutative ring, so association order cannot change the product — square-and-multiply and multiply-b-times agree bit for bit, wrapped answers included. `(pow 2 64)` still returns `0`. Evidence: a **3,861,000-pair** sweep against the naive definition with zero mismatches, plus a re-check against the old implementation over every pair it could actually compute, plus pinned wrapped answers.

`TestPowIntTerminates` runs the call on its own goroutine with a **500ms** deadline. Reverting the fix produces no answer rather than a wrong one, so a plain regression test would hang until the package's 10-minute timeout and take the whole test binary down with it.

### B2 — `(dotimes (i 2147483647))` with an empty body is uninterruptible

An empty body evaluates nothing, so the loop incremented no counter and consulted no context. **Measured before: 2e8 turns ran 44.2s of uninterruptible CPU against a 2s context deadline**, and the cancellation was only observed after the loop finished on its own. `count.Int` can be `MaxInt`, so there was no upper bound. After: the same program returns at **2.003s**.

`opDoTimes` now calls `checkLimits` once per turn.

**Cost, stated plainly.** A turn is one step more expensive. Measured at 1500 turns:

| body forms | before | after | per turn | delta |
|---|---|---|---|---|
| 0 | 4 | 1504 | 0 → 1 | the point |
| 1 | 6004 | 7504 | 4 → 5 | **+25.0%** |
| 2 | 12004 | 13504 | 8 → 9 | +12.5% |

A one-form loop of 1500 turns used to fit inside `WithMaxSteps(7000)` and no longer does. That is ordinary code, not pathological input — callers pinning a tight budget will need to raise it. `TestDoTimesStepCostPerTurn` pins these figures.

`Runtime.Steps` and `WithMaxSteps` both enumerated exactly three step sources; a dotimes turn is a fourth, and both docs are updated.

**Not claimed:** `checkLimits` short-circuits when the runtime has neither a context nor a step limit, so an embedding that configures neither still has an uninterruptible `dotimes`. Substrate sets a context but no `MaxSteps`, so what this buys there is *cancellability, not a bound*. `TestDoTimesUnboundedWithNeitherLimit` pins that honestly rather than letting the step counting read as a guarantee it is not.

### B3 — libschema invokes caller-supplied functions through a private calling convention

A validator's Go closure takes the value under test **directly**, not an argument list, so constraints are called by reaching into `FunData().Builtin` rather than through `LEnv.FunCall`. Nothing enforced that the value in a constraint slot was built by libschema:

```lisp
(s:validate identity 1)        ; index out of range [0] with length 0
(s:validate (lambda (x) x) 1)  ; nil pointer dereference (Builtin is nil)
```

A class, not two instances: **16 call sites** across `s:validate`, `s:deftype`, `s:make-validator`, `s:has-key`, `s:may-have-key`, `s:of`, `s:no-other-keys`, `s:not`, `s:when` and every `s:check-*` handler.

Every validator now carries an **unforgeable marker cell** — the address of a package-private zero-size type, which no lisp-visible path can produce. Identity, not metadata: a user function defined inside a package named `s` has the same package label and the same shape and is still refused. All 16 sites route through one `applyConstraint`, and `getHandler` rejects a foreign function at **construction** so composite constraints never build around one.

#### B3a — the inversion hazard, which is why the placement matters

`s:when` reads an error from its guard as *"condition not met, skip these checks"*. Routing the guard through the marker check at **application** time — the obvious fix — turns a refusal into a skip:

```lisp
(s:deftype "M" s:sorted-map (s:has-key "a" s:int) (s:has-key "b" s:int)
                            (s:when "a" identity "b" (s:gt 100)))
(s:validate M (sorted-map "a" 1 "b" 1))
```

| | result |
|---|---|
| before | `[INTERNAL-PANIC]` — loud, obviously broken |
| naive fix | `[list] ()` — **validation PASSES**, `b` = 1 against a constraint demanding > 100 |
| this PR | `bad-arguments` at construction |

A crash in a validator is bad; a validator that quietly approves invalid data is worse. `s:when` and `s:not` reject at construction, where no inversion can reach the refusal.

Every other error→skip inversion in the file was audited: `builtinIsFalsy` inverts only a libschema-internal constraint; `has-key`/`may-have-key`/`of` read an error as "did not match", which fails safe, and now also propagate a construction error rather than reporting a misconfigured schema as bad data.

Also fixed while auditing: `s:when` called `input.Map()` with no type check, which panics on any non-map. Reachable under `s:any`, where nothing has checked the type first.

#### B3b — the drift guard

`TestNoUnroutedConstraintInvocation` fails if a second raw invocation appears. It matches **both** spellings this file has used, `FunData().Builtin(env,` and `.Builtin()(` — the second was already sitting in `builtinIsFalsy`, so a guard grepping only the first would have declared the file clean.

#### B3c — the extension point

Before the marker, a Go embedder could pass any `lisp.LFun` where libschema expected a constraint and it worked by accident. `libschema.NewValidator` restores that capability **deliberately, with a documented contract**, rather than dropping it silently. Nothing here or in substrate uses it; it exists so closing the crash does not also remove something someone might rely on. `TestNewValidatorIsAcceptedAsAConstraint` proves the exported constructor really produces something the package accepts.

## Mutation verification

Every fix was reverted, the test run, and restored.

| # | Mutation | Result |
|---|---|---|
| 1 | `powInt` restored to exponent doubling | `TestPowIntTerminates` FAILS in **500ms** (vs. a 10-min package timeout without the goroutine bound) |
| 2 | `checkLimits` call removed from `opDoTimes` | 4 tests FAIL — context, step limit, and all three step-cost cases |
| 3 | marker check removed from `isValidator` | **40** subtests FAIL |
| 4 | `s:when` guard rejected at application instead of construction (the naive fix) | 3 tests FAIL, reporting `was ACCEPTED (result ())` |
| 5 | `.Builtin()(` spelling reintroduced in `builtinIsFalsy` | drift guard FAILS, naming both offenders |
| 6 | `s:when` non-map type check removed | FAILS with the recovered panic |

## Gates

| Gate | Result |
|---|---|
| `go build ./...` | clean |
| `go test ./...` | pass |
| `make test-elpscheck` | pass (the targets feed the singletons in deliberately, so this is the strong check) |
| `golangci-lint run --build-tags elpscheck ./lisp/...` | **0 issues** |
| `make ci-gates-test` | 75 passed, 0 failed |
| `./elps lint --workspace=. --exclude 'grammar' --include '_examples' ./...` (the CI invocation) | **9 findings, byte-identical to the base commit** |
| new gofmt-unclean files | none (17 pre-existing, unchanged set) |
| `make race` | `TestEngine_RequestPause` / `TestDAPServer_PauseRequest` — **A/B'd against a pristine `fde78f3` worktree, which fails both the same way**. Pre-existing flake, #322. `-race` over every other package passes |

## Measured CI cost

30s budget each on this machine: `FuzzApplyStdlib` 33s, `FuzzDumpJSON` 33s, `FuzzSchemaValidate` 34s — **100s added** to a job that already ran 11 targets. Seed-corpus cost in `make test` (where regressions are actually caught, in milliseconds) is ~1.5s; seed corpora are sampled rather than fully crossed, because the full cross measured ~15s and ~3s respectively.

`fuzz.yml`'s job backstop is raised 120 → 165 minutes. Discovery is dynamic, which is what makes a new target covered the day it lands *and* what makes it silently cost another `FUZZTIME` on the nightly: at 120 minutes the 14 targets would have been cut off mid-sweep and the last ones would never have run. The comment now says out loud that the number has to move whenever a target is added.

---

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -7,7 +7,11 @@ name: Fuzz
 # a phylum.  This workflow keeps a continuous fuzzing budget pointed at that
 # surface: the recursive-descent parser (strict, fault-tolerant and
 # format-preserving modes), the lexer and byte scanner, the formatter and
-# minifier round-trips, and the JSON decoder that reads chaincode state.
+# minifier round-trips, the JSON decoder that reads chaincode state and the
+# JSON encoder that writes it, and -- the widest surface of all -- every
+# builtin, special operator, macro and stdlib export, applied to generated
+# typed LVals.  A phylum calls those with arguments of its own choosing, so a
+# panic or a hang in any of them is as reachable as a parser crash.
 #
 # TIME BOUNDING IS THE LOAD-BEARING PART.  `go test -fuzz` has no default limit;
 # without -fuzztime it runs until something kills it.  All logic, including the
@@ -60,9 +64,16 @@ jobs:
     # over; this catches a hang OUTSIDE any single `go test` (a wedged build,
     # a stuck module download) that neither of those can see.
     #
-    # Sized for the scheduled run: 10 targets x 10m plus build. The PR path
-    # measures far below it -- see the measured cost in the run summary.
-    timeout-minutes: 120
+    # MUST BE RAISED WHENEVER A TARGET IS ADDED. Discovery is dynamic, so a
+    # new target silently costs another FUZZTIME on the nightly run; at 120
+    # minutes the 14 targets present today would be cut off mid-sweep and the
+    # last ones would never run at all. Sized for the scheduled run: 14
+    # targets x 10m plus build and the corpus cache restore. The PR path
+    # measures far below it: each of the three targets added with this comment
+    # measured 33-34s wall for a 30s budget (FuzzApplyStdlib 33s,
+    # FuzzDumpJSON 33s, FuzzSchemaValidate 34s), so 14 x ~33s is ~8 minutes
+    # end to end.
+    timeout-minutes: 165
     env:
       # 30s on a PR, 10m on the nightly schedule, whatever was asked for on a
       # manual dispatch. The PR budget is intentionally small: this job is a
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -8,10 +8,13 @@ name: Fuzz
 # surface: the recursive-descent parser (strict, fault-tolerant and
 # format-preserving modes), the lexer and byte scanner, the formatter and
 # minifier round-trips, the JSON decoder that reads chaincode state and the
-# JSON encoder that writes it, and -- the widest surface of all -- every
-# builtin, special operator, macro and stdlib export, applied to generated
-# typed LVals.  A phylum calls those with arguments of its own choosing, so a
-# panic or a hang in any of them is as reachable as a parser crash.
+# JSON encoder that writes it, the EVALUATOR itself -- where that same
+# untrusted source is actually run, so a panic is a crashed chaincode process
+# and a non-terminating evaluation is a wedged transaction -- and, the widest
+# surface of all, every builtin, special operator, macro and stdlib export,
+# applied to generated typed LVals.  A phylum calls those with arguments of
+# its own choosing, so a panic or a hang in any of them is as reachable as a
+# parser crash.
 #
 # TIME BOUNDING IS THE LOAD-BEARING PART.  `go test -fuzz` has no default limit;
 # without -fuzztime it runs until something kills it.  All logic, including the
@@ -60,20 +63,32 @@ jobs:
   fuzz:
     name: Fuzz targets
     runs-on: ubuntu-latest
-    # Backstop for the whole job. fuzz.sh already bounds each target twice
-    # over; this catches a hang OUTSIDE any single `go test` (a wedged build,
-    # a stuck module download) that neither of those can see.
+    # A BACKSTOP, NOT A BUDGET -- and the number must not be read as an
+    # accurate target count.  Discovery is dynamic (scripts/fuzz.sh enumerates
+    # targets at run time), so a new target silently costs another FUZZTIME on
+    # the nightly sweep, and a value sized for yesterday's count truncates the
+    # sweep with the last targets never executing at all.
     #
-    # MUST BE RAISED WHENEVER A TARGET IS ADDED. Discovery is dynamic, so a
-    # new target silently costs another FUZZTIME on the nightly run; at 120
-    # minutes the 14 targets present today would be cut off mid-sweep and the
-    # last ones would never run at all. Sized for the scheduled run: 14
-    # targets x 10m plus build and the corpus cache restore. The PR path
-    # measures far below it: each of the three targets added with this comment
-    # measured 33-34s wall for a 30s budget (FuzzApplyStdlib 33s,
-    # FuzzDumpJSON 33s, FuzzSchemaValidate 34s), so 14 x ~33s is ~8 minutes
-    # end to end.
-    timeout-minutes: 165
+    # Re-derive rather than trusting this literal: `make fuzz-list` prints the
+    # targets actually discovered on the current tree.  The arithmetic is
+    # (targets that really fuzz) x FUZZTIME + build + corpus cache restore.
+    # FuzzGateSelfTest is inert unless armed and does not count.
+    #
+    # 180 is sized for the fully-landed round-2 state: main carries 11
+    # targets, the evaluator work adds FuzzEval, the stdlib work adds
+    # FuzzApplyStdlib / FuzzDumpJSON / FuzzSchemaValidate, and retiring
+    # regexparser removes FuzzParseLVal -- 14 that really fuzz, so 14 x 10m =
+    # 140 minutes plus overhead.  It is deliberately above every intermediate
+    # merge order as well, because over-provisioning a backstop costs nothing
+    # while under-provisioning drops coverage silently.
+    #
+    # Measured PR-path cost (FUZZTIME=30s, 4-core linux/amd64): the whole
+    # sweep is ~6 minutes; FuzzEval 32s, FuzzApplyStdlib 33s, FuzzDumpJSON
+    # 33s, FuzzSchemaValidate 34s.
+    #
+    # The durable fix is a derived gate asserting
+    # (target count x nightly budget) <= timeout-minutes -- tracked in #320.
+    timeout-minutes: 180
     env:
       # 30s on a PR, 10m on the nightly schedule, whatever was asked for on a
       # manual dispatch. The PR budget is intentionally small: this job is a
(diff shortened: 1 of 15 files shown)


sam-at-luther:

```yaml
on:
  pull_request:
    branches:
      - main
```

This PR targets `claude/elps-fieldalignment` (stacked on #311 for the fuzzing infrastructure), so `Lint & Test`, `Fuzz targets`, `CI Gate Self-Test`, `govulncheck`, `Build (Windows)` and `Benchmark Comparison` **never fired**. The only checks present are the two Socket ones, which are a GitHub App and not branch-filtered.

---

sam-at-luther:
Pre-resolved the collision with #323 so it isn't a surprise at merge time. **Either merge order now works with no conflict** — verified by simulating both squash-merges locally, then building and testing the combined tree (all packages pass; all 7 `dotimes` tests from both PRs pass together).

Two things collided.

**The `dotimes` fix.** Both PRs found issue #320's B2 independently and fixed it in `opDoTimes`. The fixes turned out to be equivalent — each one passes the other's tests, which is how I checked rather than reasoning about it — but they differed in placement relative to `n++`, in passing `loopenv.evalCtx` vs `env.evalCtx`, and in that only this PR updated the `WithMaxSteps` / `Runtime.Steps` docs. Both hunks and both doc blocks are now byte-identical across the branches. `env.evalCtx` won: `newEnvN` copies the parent's ctx at construction, so the two are equal today, but a snapshot could go stale if `loopenv` were ever hoisted.

**A correction while doing it: the per-turn cost was misreported in both places.** This PR's comment said "4 to 5 steps per turn"; issue #320 said "6 → 7 steps (+16.6%)". Both are the same +1 seen through different loop bodies, stated as though the percentage were a property of the fix. Measured at n=1000 on this tree:

| body | before | after |
|---|---:|---:|
| *(empty)* | 0 | 1 |
| `1` | 1 | 2 |
| `(+ i 1)` | 4 | 5 |
| `(+ (* i 2) (- i 1))` | 10 | 11 |
| `(+ i 1) (+ i 2) (+ i 3)` | 12 | 13 |

So it is exactly **one extra step per turn**, and the fraction ranges from +8% to +100% purely with body size. The docs now give the absolute figure and point at the table instead of quoting a percentage.

**`.github/workflows/fuzz.yml`.** Each PR rewrote the header prose for its own targets and sized `timeout-minutes` off its own count (140 vs 165). Both regions are now identical, and the backstop is sized for the fully-landed set: main has 11 targets, +`FuzzEval`, +the three added here, −`FuzzParseLVal` once #321 retires regexparser = **14 that really fuzz** (`FuzzGateSelfTest` is inert unless armed), so 14 × 10m = 140 plus overhead → **180**, which also clears every intermediate merge order.

The comment now says to re-derive with `make fuzz-list` rather than trust the literal, because a stale count truncates the nightly sweep with the last targets never running — and a truncated sweep looks exactly like a finished one. The durable fix is the derived gate (count × budget ≤ timeout) already tracked as a separate item in #320.

No functional change to what this PR does; the reconciliation is mechanical plus the corrected measurement.

---

PR #326, opened 2026-08-08
Grammar-aware fuzzing of the parse -> format -> minify pipeline, and the 9 defects it found

sam-at-luther:
Stacked on #311 — this PR targets `claude/elps-fieldalignment`, not `main`.

## What this adds

`internal/fuzzgen` turns the fuzzer's byte script into **syntactically valid but semantically hostile** ELPS, and `FuzzGeneratedPipeline` drives the whole source-rewriting chain over it:

```
parse -> format -> re-parse -> format again -> minify -> re-parse
      -> format the minified output -> re-parse
```

Every existing target takes raw bytes, so the mutator spends most of its budget on input that dies at the first token. The states *behind* a successful parse — comment attachment, blank-line bookkeeping, prefix-form re-sugaring, scope-aware renaming — are exactly where `elps fmt` and `elps minify` silently rewrite a customer's program, and raw bytes hardly ever get there.

Generation is a pure function of `(script, limits)`, bounded three ways (byte budget, nesting cap, script length). Corpora are committed as **scripts**, not `.lisp` source. Coverage: pathological nesting, quasiquote / unquote-splicing, `#'` and `#^`, both bracket kinds, keyword and package-qualified symbols, huge literals, unicode identifiers, and comments in every position the grammar permits one — before a form, between siblings, before a closing bracket, trailing on the same line, between a prefix macro and its operand, and at EOF.

**No dialects.** The previous iteration carried a `Core` subset defined as `rdparser ∩ regexparser`; regexparser is being deleted in #321, so that definition names nothing. One grammar, and the package comment says why. Nothing in this PR touches `parser/regexparser`.

## The invariants, and why each is safe to assert

| # | Invariant | Justified by |
|---|---|---|
| 1 | the generated program parses | fuzzgen's own contract, pinned by `TestGeneratedProgramsAlwaysParse` over 24k scripts × 4 limit profiles |
| 2 | `Format` accepts everything the parser accepts | `FuzzFormat` asserts the converse; formatting mode only *attaches* metadata to the same tree (`FuzzParseFormatting`) |
| 3 | `Format` preserves the AST | `formatter_test.go: roundTripEqual`, `FuzzFormat` |
| 4 | `Format` drops no comment | `TestCommentPreservesText` and the comment tests in `formatter_test.go`, generalised |
| 5 | `Format` invents no interior blank line | every blank line the printer writes comes from a source-measured count clamped to `MaxBlankLines`, so it can shrink but not grow |
| 6 | `Format` is idempotent | `runFormatTests`, `TestEdgeIdempotency`, `FuzzFormat` |
| 7 | minify accepts formatted source, is deterministic | `TestMinifySource_DeterministicAndScopeAware`, `FuzzMinifySource` |
| 8 | minified output parses, same type skeleton | `FuzzMinifySource` |
| 9 | formatting minified output round-trips and is idempotent | 3+6 applied to the minifier's output |

(4) and (5) are the ones that earn their keep. A formatter that **deletes a comment** or **inserts a blank line** leaves the AST identical, leaves the output a stable fixed point, and passes every other check — six of the nine defects below were fixed points that had simply lost, or gained, a line. Nothing is asserted about layout, chosen names, or semantics.

## Defects fixed

The four from #320 §C, plus C6 (which the C4 fix introduces — shipped together, never apart), plus four more of the same class the comment invariant turned up.

| | defect | effect |
|---|---|---|
| **C2** | float exponents rejected | `1e5`, `2e3`, `1E5`, `1.5e2` were scan errors while `1e55` and `1e+5` parsed — both call sites consumed one unconditional rune before `readFloatExponent`, so an exponent needed *two* characters. `1e`, `1e+`, `1e-`, `1.`, `1.5e` still rejected; `1e5e5` now splits into `1e5` + `e5`, matching how `1abc` has always split. |
| **C3** | no string could end in a backslash | escape parity was read off the *last rune* of a maximal run, so `"a\\"` consumed its own closing quote. Parity is now counted. Genuinely unterminated strings still rejected. |
| **C4** | comments before a prefix form deleted | `; c\n#'a` → `#'a`. And `#!/usr/bin/env elps\n#':%` printed the comment *inside* the longhand form, producing output the parser then rejected. |
| **C5** | signed literal swallowed by a comment | `(a ; c\n-1)` → `(a ; c -1)`, which re-parses as `(a)`. `ParseNegative` merged the sign token without its whitespace bookkeeping. |
| **C6** | blank line inserted before a comment | `(a)\n; c\n\n#'f\n` → `(a)\n\n; c\n\n#'f\n`. `applyPrefixNewlines` carried a premise the C4 fix invalidated. Only `#'` is affected, because `readFunRef` emits FUN_REF and its symbol from one `readToken` so the operand inherits the prefix's `PrecedingNewlines`; `'` and `#^` scan separately and report 0. **It hits a real file in two repos** — `editors/vscode/test/grammar/basics.lisp`, where the `; <-` markers are syntax-test assertions *about the line above*. |
| **N1** | comment in an empty form deleted | `(\n; c\n)` → `()`. `writeSExpr`'s empty-cells branch skipped both `InnerTrailingComments` and `ClosingBracketNewline`; `[...]`, `'(...)` and non-empty forms all reach `writeListInner`, which handles both. Now at parity. |
| **N2** | comment before a form's head deleted | `(\n; c\n f x)` → `(f x)`. `writeSExpr` never wrote the head child's leading comments. |
| **N3** | comment between a prefix and its operand deleted | `'\n; c\n[a]` → `'[a]`, `#^; c\na` → `#^a`. Now hoisted onto the prefix node. |
| **N4** | re-sugaring longhand deleted comments | `(lisp:function ; c\n f)` → `#'f`. The shorthand has nowhere to put them, so those forms stay longhand — except under `StripComments`, where nothing is lost. |

`takeLeadingComments`/`attachLeadingComments` and `hoistOperandComments` turned out to be the same fix from opposite ends: neutering the first left `Format` and `MinifySource` **byte-identical over 32,000 generated programs** with every test still passing. Collapsed onto one mechanism rather than shipping a fix no test can distinguish from its own absence.

## Evidence

**Mutation table** — each fix reverted individually; both the unit test and the committed fuzz corpus (seeds + `testdata/`) re-run at plain `go test`:

| defect | unit test | unit | fuzz corpus |
|---|---|---|---|
| C2 | `parser/lexer TestFloatExponentSingleDigit` | FAILS | FAILS |
| C3 | `parser/lexer TestStringTrailingBackslash` | FAILS | FAILS |
| C4 | `formatter TestCommentBeforePrefixForm` | FAILS | FAILS |
| C5 | `formatter TestNegativeLiteralAfterComment` | FAILS | FAILS |
| C6 | `formatter TestCommentBeforePrefixForm` | FAILS | FAILS |
| N1 | `formatter TestCommentInEmptySExpr` | FAILS | FAILS |
| N2 | `formatter TestCommentBeforeSExprHead` | FAILS | FAILS |
| N3 | `formatter TestCommentBetweenPrefixAndOperand` | FAILS | FAILS |
| N4 | `formatter TestLonghandPrefixFormKeepsComments` | FAILS | passes |

N4 is unit-test-only by construction: the generator has no reason to write `(lisp:function f)` in longhand, so the fuzz corpus cannot reach it.

The test-gap C6 exploited is closed: `TestCommentBeforePrefixForm` is now a matrix over {`'`, `#'`, `#^`, plain} × {first form, after a form} × {adjacent, blank line} — the original had neither a blank-line row nor a preceding-form row — with `'`/`#^`/plain as controls so an over-broad fix fails.

**Real-corpus round trip** — every `.lisp` in this repo (excluding `tree-sitter-elps/`) and in `luthersystems/substrate`, 126 files, through `elps fmt` and `elps minify`, before and after:

- **`minify`: byte-identical.** Zero diff.
- **`fmt`: 6 added lines, all restorations of content the old build deleted** — two comments and one blank line, in `editors/vscode/test/grammar/basics.lisp` in each repo. No file changed in any other way; no file gained or lost an error.
- `basics.lisp` still differs from its formatted form in 3 lines: `elps fmt` indents `;   ^^^^` markers inside a `defun` body. That behaviour is **unchanged from before this PR** (base produced the same 3), it is pre-existing and out of scope, and the CI gate already excludes these fixtures (`elps fmt -l --exclude 'grammar'`).

**Gates** (all green): `go build ./...` · `go test ./...` · `make race` · `make test-elpscheck` · `golangci-lint run ./...` **0 issues** · `golangci-lint run --build-tags elpscheck ./lisp/...` **0 issues** · `./elps doc -m` · `./elps fmt -l --exclude 'grammar' ./...` clean · `make ci-gates-test` **75 passed, 0 failed** · `./elps lint` **9 warnings, identical to base** · no new gofmt-unclean files (15 pre-existing, unchanged).

**Measured CI cost.** No workflow change: `scripts/fuzz.sh` discovers targets dynamically, so the new one is picked up automatically (12 targets, was 11).

- Fuzz workflow, PR path (`FUZZTIME=30s`): **+32s wall** for the new target, measured. Nightly (`10m`): +~10m, inside the existing 120-minute job backstop.
- Tests workflow: **+4s** for `./internal/fuzzgen` (the deterministic sweeps); `./formatter` still ~1s.
- Target throughput: ~2,000 exec/sec, 192 new interesting inputs in a 45s run.

## Not verified

- The blank-line invariant (5) is argued from the printer's structure and measured over 72,000 generated programs plus all 126 real files; it is not proven.
- `make race` passes here, so the `TestEngine_RequestPause` flake from #322 did not reproduce; I did not A/B it against a pristine checkout.
- The three `;   ^^^^` indentation differences in `basics.lisp` are unchanged from base and left alone.

---

### DIFF
--- a/formatter/comment_preservation_test.go
+++ b/formatter/comment_preservation_test.go
@@ -156,6 +156,49 @@ func TestCommentBetweenPrefixAndOperand(t *testing.T) {
 	}
 }
 
+// TestLonghandPrefixFormKeepsComments covers the other direction: longhand a
+// human wrote out, with a comment somewhere inside it.
+//
+// The printer re-sugars "(lisp:function f)" to "#'f", and the shorthand has
+// nowhere to put a comment that sat between the brackets -- so re-sugaring
+// DELETED it.  Those forms now stay longhand.  The comment-free rows are the
+// controls: re-sugaring is the normal, desirable outcome and must not regress.
+func TestLonghandPrefixFormKeepsComments(t *testing.T) {
+	tests := []formatTest{
+		{
+			name:     "funref/leading",
+			input:    "(lisp:function\n; c\nf)\n",
+			expected: "(lisp:function\n  ; c\n  f)\n",
+		},
+		{
+			name:     "funref/trailing-on-head",
+			input:    "(lisp:function ; c\nf)\n",
+			expected: "(lisp:function ; c\n  f)\n",
+		},
+		{
+			name:     "funref/inner-trailing",
+			input:    "(lisp:function f\n; c\n)\n",
+			expected: "(lisp:function f\n               ; c\n               )\n",
+		},
+		{
+			name:     "unbound/trailing-on-head",
+			input:    "(lisp:expr ; c\n%)\n",
+			expected: "(lisp:expr ; c\n  %)\n",
+		},
+		// Controls: no comment, so the shorthand is still written.
+		{name: "funref/plain", input: "(lisp:function f)\n", expected: "#'f\n"},
+		{name: "unbound/plain", input: "(lisp:expr %)\n", expected: "#^%\n"},
+		{name: "funref/already-short", input: "#'f\n", expected: "#'f\n"},
+		// A comment INSIDE the operand belongs to the operand's own printer
+		// and survives the shorthand, so it must not block re-sugaring.
+		{name: "unbound/operand-inner", input: "#^(\n; c\n)\n", expected: "#^(\n   ; c\n   )\n"},
+	}
+	runFormatTests(t, tests)
+	for _, tt := range tests {
+		requireCommentsPreserved(t, tt.input)
+	}
+}
+
 // TestCommentInEmptySExpr covers a comment between the brackets of a form with
 // no children.
 //
--- a/formatter/comment_preservation_test.go
+++ b/formatter/comment_preservation_test.go
@@ -38,6 +38,8 @@ func commentTokens(t *testing.T, src string) []string {
 				out = append(out, tok.Text)
 			case token.EOF, token.ERROR, token.INVALID:
 				return out
+			default:
+				// Every other token type is irrelevant here.
 			}
 		}
 	}
--- a/formatter/comment_preservation_test.go
+++ b/formatter/comment_preservation_test.go
@@ -1,0 +1,275 @@
+// Copyright © 2026 The ELPS authors
+
+package formatter
+
+import (
+	"bytes"
+	"strings"
+	"testing"
+
+	"github.com/luthersystems/elps/parser/lexer"
+	"github.com/luthersystems/elps/parser/token"
+	"github.com/stretchr/testify/require"
+)
+
+// `elps fmt` deleting a comment is not visible to any AST-shaped check.
+// Comments are not part of the tree, so a dropped one leaves the parsed
+// program identical AND leaves Format idempotent -- the output is a perfectly
+// stable fixed point that is simply missing a line of the user's file.  Five
+// separate defects presented exactly that way, and every one of them was found
+// by comparing the COMMENT TOKENS on either side of a format rather than the
+// trees.
+//
+// The tests below pin each shape with its exact expected output (via
+// runFormatTests, which also asserts idempotency), and commentsPreserved
+// re-checks the whole set token-by-token so a future defect in one of these
+// paths cannot hide behind a plausible-looking expectation.
+
+// commentTokens returns the text of every comment in src, in order.  A
+// hash-bang lexes as HASH_BANG plus a COMMENT holding the rest of the line.
+func commentTokens(t *testing.T, src string) []string {
+	t.Helper()
+	lex := lexer.New(token.NewScanner("test", bytes.NewReader([]byte(src))))
+	out := []string{}
+	for range 1 << 12 {
+		for _, tok := range lex.ReadToken() {
+			switch tok.Type {
+			case token.COMMENT, token.HASH_BANG:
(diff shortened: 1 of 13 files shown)


sam-at-luther:

```yaml
on:
  pull_request:
    branches:
      - main
```

This PR targets `claude/elps-fieldalignment` (it is stacked on #311 for the fuzzing infrastructure), so `Lint & Test`, `Fuzz targets`, `CI Gate Self-Test`, `govulncheck`, `Build (Windows)` and `Benchmark Comparison` **never fired**. The only checks present are the two Socket ones, which are a GitHub App and not branch-filtered.

---

PR #327, opened 2026-08-08
fix(formatter): keep a comment written in column 0 in column 0

sam-at-luther:
`elps fmt` re-indented every own-line comment to its enclosing form's indent. When a comment's content is column-aligned against the line **above** it, the column *is* meaning, and re-indenting destroys it silently.

## The victims

`editors/vscode/test/grammar/*.lisp` are fixtures for `vscode-tmgrammar-test`, where `; <-` and `; ^^^` comments are **positional assertions about the preceding line**:

```lisp
(defun f (x &optional y)
;           ^^^^^^^^^ variable.parameter.elps
```

Formatting shifted those carets two columns right. The fixture kept asserting — about the wrong characters. `elps fmt` corrupted the test while reporting success.

## The rule

gofmt's, adopted for gofmt's reason: a `//` comment starting in column 1 stays in column 1, because a flush-left comment inside an indented body is a deliberate signal rather than an accident of layout.

`commentIndent` keys off the comment's **original** column, so an indented comment is still re-indented to its form. Only column 0 is sticky.

**Scoped to leading comments.** Inner-trailing comments sit between the last child and the closing bracket, where there is no line above and nothing to align against. Applying the rule there produced strictly worse output for no benefit:

```
(lisp:function f
; c                     <- worse
               )
```

## Blast radius, measured

Of **33** `.lisp` files in the repo, exactly **3** format differently — and all 3 are the grammar fixtures. No real ELPS code in the tree writes a column-0 comment inside a form. All five fixtures are now byte-identical under `elps fmt`, and every file in the tree is idempotent under it.

## The tradeoff, stated plainly

An unindented comment inside a nested form now stays flush left:

```lisp
(defun foo ()
; setup

  (init)
  (run))
```

That is intended, and gofmt does the same to the equivalent Go. It is the cost of the rule, and it is why the measurement above matters: no real code in this repo hits it.

## Tests changed

Nine assertions from #326 moved. Every one is a case whose *input* put `; c` in column 0 because it was written as a terse Go string literal, not because anything was expressing intent about the column.

Each changed row gained a **companion row with the comment indented**, so the table now pins both directions. Without those, a bug that flattened every comment to the left margin would pass every one of these tests.

## Guards, both verified to fail

| Guard | What it pins |
|---|---|
| `TestCommentInColumnZeroIsPreserved` | Paired column-0 / indented cases, including the alignment shape itself. Each case also asserts the **second** format pass is a no-op — a rule that reads the original column has to survive being fed its own output. |
| `TestGrammarFixturesAreFormatFixedPoints` | Reads the five real fixture files and requires byte-identical output. |

The second exists as a test rather than a note for a specific reason: **CI runs `elps fmt -l --exclude 'grammar'`**, so the exclusion means nothing else in the build would ever notice this regression. A guard that depends on the tool whose exclusion hid the bug is not a guard.

Reverting `commentIndent` fails both: 3 of 6 subtests in the first, 3 of 5 fixtures in the second.

## Verification

- `go test ./...` — pass
- `golangci-lint run ./formatter/...` — 0 issues
- `gofmt -l formatter/` — clean
- `./elps doc -m` — all documented
- `./elps fmt -l --exclude 'grammar' ./...` — clean
- Every `.lisp` in the tree is idempotent under `elps fmt`; 0 non-idempotent files

Refs #320

---

### DIFF
--- a/.github/workflows/elps.yml
+++ b/.github/workflows/elps.yml
@@ -51,7 +51,15 @@ jobs:
       - name: Check formatting
         run: |
           go build -o elps .
-          ./elps fmt -l --exclude 'grammar' ./...
+          # No --exclude here, deliberately. The 'grammar' carve-out existed
+          # only because `elps fmt` rewrote editors/vscode/test/grammar/*.lisp,
+          # whose `; <-` / `; ^^^` comments are POSITIONAL assertions about the
+          # line above -- so the reflow changed what the fixture asserted. The
+          # carve-out made CI green by not looking, which also meant CI could
+          # never catch a future formatter change that broke those files again.
+          # Now that a column-0 comment stays in column 0 they are fixed points,
+          # so the fixtures are back under the gate.
+          ./elps fmt -l ./...
 
       - name: Lint ELPS source
         run: ./elps lint --workspace=. --exclude 'grammar' --include '_examples' ./...
--- a/formatter/comment_column_test.go
+++ b/formatter/comment_column_test.go
@@ -1,0 +1,128 @@
+// Copyright © 2026 The ELPS authors
+
+package formatter_test
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/luthersystems/elps/formatter"
+	"github.com/stretchr/testify/require"
+)
+
+// TestCommentInColumnZeroIsPreserved pins the rule in printer.go's
+// commentIndent: a comment written in column 0 stays in column 0, even inside
+// an indented form.
+//
+// The rule exists because a flush-left comment can be column-aligned against
+// the line ABOVE it, in which case its column carries meaning that
+// re-indenting destroys. The case that forced it is
+// editors/vscode/test/grammar/*.lisp, whose TextMate syntax-test assertions
+// look like
+//
+//	(defun f (x &optional y)
+//	;           ^^^^^^^^^ variable.parameter.elps
+//
+// where the caret column IS the assertion. `elps fmt` indented those comments
+// by the form's indent, silently changing which characters they asserted
+// about while appearing to succeed. TestGrammarFixturesAreFormatFixedPoints
+// below is the end-to-end guard on the real files.
+//
+// Every case comes in a pair: the column-0 input and the same shape written
+// with the comment indented. The indented half is the one that proves the
+// rule keys off the comment's ORIGINAL column instead of just always writing
+// column 0 -- without it, a bug that flattened every comment to the left
+// margin would pass.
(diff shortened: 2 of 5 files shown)


sam-at-luther:
Two follow-ups from review, plus a correction to a number I posted.

## Correction: substrate is 4 of 93, not 0

I measured substrate before this comment and reported zero impact. **That was a false negative.** The sandbox reset the elps working tree mid-session and wiped `/tmp`, so both binaries in that comparison were rebuilt from unfixed `main` — I was diffing the formatter against itself. Re-measured from the pushed branch with a verified before/after pair:

| | result |
|---|---|
| substrate `.lisp` corpus | 93 files, 15,417 lines, 1,153 column-0 comment lines |
| files that format differently under the new rule | **4** — all `editors/vscode/test/grammar/*.lisp` |
| shirocore / phylum / test `.lisp` affected | **0** |

So the conclusion survives, and is now actually measured: no production ELPS is touched. The 89 non-fixture files are unchanged.

**Testing there was worth it** — substrate has a sixth fixture elps doesn't, `substrate.lisp`, and it has the same bug:

| fixture | today | with this change |
|---|---|---|
| basics.lisp | ✗ | ✓ |
| builtins.lisp | ✓ | ✓ |
| definitions.lisp | ✗ | ✓ |
| special-ops.lisp | ✗ | ✓ |
| **substrate.lisp** | ✗ | ✓ |
| symbols.lisp | ✓ | ✓ |

The one file `elps fmt -l` still flags in substrate is `internal/substrate/shirocore/shirocore.lisp`, which is generated minified output and isn't expected to be fmt-clean.

## The grammar carve-out is gone (`b9727fd`)

Fair challenge — it shouldn't need one, and it no longer does. `--exclude 'grammar'` on `elps fmt -l` existed *only* because the formatter rewrote those files. It made CI green by not looking, which is also why this defect survived unnoticed: nothing in the build was ever going to catch it.

With column-0 comments preserved, `./elps fmt -l ./...` passes with **no exclusion at all** (verified with the exact CI command), so the fixtures are back under the gate.

**The `elps lint --exclude 'grammar'` beside it is deliberately kept**, because it is a different kind of carve-out. Those fixtures are intentionally odd ELPS written to exercise a TextMate grammar — duplicate definitions, unusual forms — so the linter produces **64 genuine findings** in them and exits 1 without the exclusion. That one describes a real property of the files rather than hiding a defect.

The fixed-point Go test stays alongside the now-unexcluded CI check rather than being replaced by it: it exercises the library rather than the CLI, so it still fires if the two diverge, and it fails with a message explaining *why* these files must not move — which a bare "file is not formatted" does not.

---

PR #340, opened 2026-08-09
Give tree-sitter-elps dependabot coverage, and enforce that every manifest has it

sam-at-luther:
## The gap

`tree-sitter-elps/` is a separate Go module **and** an npm package with a committed `package-lock.json`. It is built and tested on every pull request by `.github/workflows/tree-sitter.yml`. It had **no dependabot entry at all**.

`.github/dependabot.yml` listed `gomod` at `/` and `npm` at `/editors/vscode`. Neither reaches a nested module — `go list ./...` from the repo root does not descend into `tree-sitter-elps`, and neither does Dependabot. So these were frozen at whatever they were when they were written, receiving neither version bumps nor security updates:

| Manifest | Dependencies left unwatched |
|---|---|
| `tree-sitter-elps/go.mod` | `go-tree-sitter v0.25.0`, `testify v1.11.1` (+ 4 indirect) |
| `tree-sitter-elps/package.json` | `node-addon-api ^8.3.0`, `node-gyp-build ^4.8.4`, `prebuildify ^6.0.1`, `tree-sitter-cli ^0.24.0`, `tree-sitter ^0.22.4` |

Two entries added, each in its own update group so a tree-sitter bump does not arrive mixed into the root `go-deps` PR.

## The guard, which is the point

This failure mode has **no symptom**. The dependabot dashboard is green because it is not looking, which is indistinguishable from green because there is nothing to find. Nobody notices until a CVE lands in a dependency nobody is watching. The fix for the instance is two YAML entries; the fix for the *class* is that adding a module and forgetting the entry has to fail CI.

`scripts/ci-gates-test.sh` now walks the tree for `go.mod` and `package.json` (skipping `node_modules/`, `vendor/`, `testdata/`, `build/`) and fails if any manifest lacks a matching entry. **Discovery, not a second hardcoded list** that could fall out of date the same way the first one did. It accepts both the `directory:` and the newer plural `directories:` spelling, so switching to the latter will not read as a regression.

It checks the converse too: an entry whose directory holds no manifest is the same failure seen from the other side, since it looks like coverage and provides none.

## Verification

The guard was mutation-tested against synthetic trees, not just observed to pass on a tree that already satisfies it:

| Case | Result |
|---|---|
| the **pre-change** config against this repo's real manifest layout | **2 FAIL**, naming `gomod /tree-sitter-elps` and `npm /tree-sitter-elps` exactly |
| an entry pointing at a directory with no manifest | **1 FAIL**, naming the stale entry |
| `package.json` planted under `node_modules/` at two different depths | still clean — the walk cannot be tripped into demanding entries for installed packages |

On the real tree it reports 4 manifests, all covered. `bash -n` and `shellcheck -S warning` clean. PyYAML is already a hard dependency of the existing workflow-shape guards in the same script, so the new check adds no new requirement — and it degrades to `SKIP` rather than a false pass if it is ever absent.

## Note on overlap

#339 also touches `scripts/ci-gates-test.sh`, in the shell-lint section near the end of the file; this change inserts a new section in the middle. Whichever lands second I will rebase.

---

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,4 +1,21 @@
 version: 2
+
+# EVERY dependency manifest in the repository must have an entry here.
+#
+# Dependabot only looks where it is told. A manifest with no matching
+# `directory:` gets no security updates and no version bumps, and nothing
+# anywhere reports that — the dashboard shows green because it is not looking,
+# which is indistinguishable from green because there is nothing to find.
+#
+# That is not hypothetical here: tree-sitter-elps/ is a separate Go module AND
+# an npm package with a lockfile, it is built and tested on every PR by
+# .github/workflows/tree-sitter.yml, and it had no entry at all. Its
+# go-tree-sitter, node-addon-api and tree-sitter-cli dependencies were frozen
+# at whatever they were when they were added.
+#
+# scripts/ci-gates-test.sh enforces the rule: it walks the tree for go.mod and
+# package.json (ignoring vendor/ and node_modules/) and fails if any of them
+# lacks a matching entry below. Adding a module means adding an entry here.
 updates:
   - package-ecosystem: "github-actions"
     directory: "/"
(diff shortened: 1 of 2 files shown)


sam-at-luther:
Correcting the "Note on overlap" at the bottom of the description: **no rebase will be needed.** I said I'd rebase whichever of #339 / #340 landed second, but I hadn't actually tried it. A trial merge of the two branches auto-merges `scripts/ci-gates-test.sh` cleanly with zero conflicted files — #339 rewrites the shell-lint block at the end, this one inserts a new section in the middle, and git's 3-way merge has ample shared context between them. They can land in either order.

`CI Gate Self-Test` on this branch confirms the new guard runs and passes in CI, not just locally:

```
== dependabot covers every dependency manifest ===========================
PASS  every dependency manifest has a dependabot entry
      (4 manifests: gomod/, gomod/tree-sitter-elps, npm/editors/vscode, npm/tree-sitter-elps)
PASS  no dependabot entry points at a manifest that does not exist
```

Worth noting for review: that job runs with `CI_GATES_SKIP_GO=1`, so it skips the Go-backed fuzz-gate checks. The new guard needs only Python and PyYAML — both already hard dependencies of the existing workflow-shape guards in the same script — so it runs in full there, and it degrades to `SKIP` rather than a silent pass if PyYAML is ever absent.


---

PR #341, opened 2026-08-09
Require both halves of a package-qualified symbol to be names (#319)

### LINKED ISSUE #319: rdparser accepts package-qualified symbols whose name is a bare digit (`a:1`, `:1`)
Surfaced by the adversarial review of the parser work in PR #311, while spot-checking the rdparser/regexparser divergence table. Flagged as latent, not fixed there.

## The issue

`rdparser` accepts `a:1` and `:1` as symbols — a package-qualified name whose local part is a bare digit.

This is the **same class** as the `#'0` defect fixed on that PR. The root cause there was recorded in `readFunRef`:

> `isWord` is the CONTINUATION class and admits digits; `readSymbol` gets away with using it because it is only ever entered after `isWordStart` or `':'` has already matched.

The `#'` path was fixed by requiring the operand to read back as a symbol. The `pkg:name` path has the same shape — `':'` matching is what admits the digit — and was left alone.

## Why it is lower severity than `#'0`

`a:1` prints and re-reads **stably**, so unlike `(lisp:function 0)` it does not corrupt a round trip through `elps fmt`. That is the whole reason `#'0` was urgent and this is not.

It is still a name nothing else in the language can write, and it is reachable from source, so it is worth deciding about deliberately rather than leaving as an accident of the rune-class reuse.

## Suggested direction

Decide whether a package-qualified local part should be required to be a symbol name (rejecting `a:1`), and if so apply the same treatment `readFunRef` got. Check `:1` separately — a leading `':'` yields a keyword, and keywords have their own rules.

Worth grepping phylum sources in `luthersystems/substrate` for `:<digit>` symbols before tightening, in case anything relies on it.

## Provenance

Found by inspection during review, not by fuzzing. The divergence table pins `a:1` as "rdparser accepts, regexparser rejects" and treats rdparser as authoritative, which silently normalises this — noting it here so the row does not read as a blessing.

Note: that divergence table is being removed along with `parser/regexparser`, so this issue is the only remaining record.

sam-at-luther:
Fixes #319.

`rdparser` accepted `a:1` — a package-qualified symbol whose local part is a bare digit. Same root cause as the `#'0` defect recorded in `lexer.readFunRef`: `isWord` is the CONTINUATION rune class and admits digits, and `readSymbol` gets away with it only because it is entered after `isWordStart` or `':'` has already matched. Matching `':'` is what admits the digit. The `#'` path was fixed by requiring the operand to read back as a symbol; the `pkg:name` path had the same shape and was left alone.

## Current behaviour, measured before deciding anything

Every row observed against `ParseProgram` on `main`:

| Input | Before | After | Note |
|---|---|---|---|
| `a:1` | symbol `a:1` | **rejected** | the reported case |
| `a:1b` | symbol `a:1b` | **rejected** | |
| `a:1abc` | symbol `a:1abc` | **rejected** | not a number by `strconv`, still reads back as INT + symbol |
| `a:0x10` | symbol `a:0x10` | **rejected** | reads back as INT `0` + symbol `x10` |
| `a:-1` | symbol `a:-1` | **rejected** | reads back as INT |
| `a:1.5`, `a:1e5` | symbol | **rejected** | read back as FLOAT |
| `a:1_`, `a:9` | symbol | **rejected** | |
| `a:b`, `lisp:set`, `xyz:abc?` | symbol | symbol | unchanged |
| `a:+1`, `a:.1`, `a:*1`, `a:_1` | symbol | symbol | `+ . * _` are ordinary symbol characters, so these *are* writable bare |
| `a:-`, `a:--`, `a:-a`, `-a:b` | symbol | symbol | lex as NEGATIVE, only become symbols once `ParseNegative` merges — a token-level guard would wrongly reject them |
| `a:true`, `a:<=`, `a:set!`, `a:->` | symbol | symbol | unchanged |
| `:1`, `:1.5`, `:-1`, `:+1`, `:1a` | keyword | **keyword (unchanged)** | deliberate — see below |
| `:b` | keyword | keyword | unchanged |
| `1:1` | INT `1` + keyword `:1` | same | never one symbol: the lexer stops the number at `':'` |
| `-1:b` | INT `-1` + keyword `:b` | same | |
| `1a:b` | INT `1` + symbol `a:b` | same | |
| `a:`, `:`, `::`, `a::b`, `a:b:c` | rejected | rejected | pre-existing |
| `#'a:1` | `(lisp:function a:1)` | **rejected** | `readsBackAsSymbol("a:1")` was true, so the funref guard did not catch it |
| `#'a:b` | `(lisp:function a:b)` | unchanged | |

The guard is exactly "the local part must not start with a digit": a digit start always routes to `readNumber`, which can only emit INT/FLOAT or an INT followed by a separate token, never a single symbol.

## Decision

**Tighten `pkg:name`. Leave `:name` (keywords) alone. Document both.**

`pkg:name` is the qualified *spelling of a binding's name*. `docs/lang.md` has always said identifiers cannot start with a number, and ELPS has no `|...|`-style escape, so `a:1` is the only way to write the symbol named `1` — a binding reachable only through the qualified spelling, never importable usefully by `use-package`. That is a trap, not a capability, and qualified access should be another way to spell a name rather than a way to introduce one that cannot be written unqualified.

`:1` is a different thing and is **not** tightened. A leading `':'` with no package yields a keyword: a self-evaluating literal that names no binding and that `PutGlobal` explicitly refuses to bind (`(set ':k 1)` errors). The identifier rule governs names of bindings, so it has nothing to say about a keyword. Keywords are the language's only interned-literal facility and `:1` is usable data — verified: it evaluates to itself and works as a sorted-map key, `(sorted-map :1 "x")` → `(sorted-map ':1 "x")`. The one consequence worth knowing is that a keyword whose name is not a legal identifier can never match a `&key` formal (no formal can be spelled that way), so `(f :1 2)` is always "unrecognized keyword argument: 1". That is now documented and pinned rather than left implicit.

Round-trip stability was the reason this was lower severity than `#'0`, and it remains true — `a:1` printed and re-read stably, so nothing here fixes a formatter corruption. It is fixed because it is a name the language otherwise forbids, reachable from source.

## Implementation

Same treatment `readFunRef` got — re-read and require a symbol back — rather than a second, differently-worded rune test. `ParseSymbol` checks **both** halves, not just the local one: the package half is provably safe today only because of the lexer's entry conditions, and relying on that reasoning is exactly what produced the bug.

Two supporting changes fell out of it:

- **`token.NewScannerString`** sizes the scanner window to the source instead of allocating the fixed 128KiB window per scanner. `readsBackAsSymbol` builds one scanner per call, so through `NewScanner` the check cost **2.6GB and 1.7s** to parse 10k qualified symbols. That amplification already existed for every `#'` operand; this removes it there too.
- **`Parser.readsBackAsSymbol` memoizes** per name. Real source repeats names, so the sub-parse runs once per distinct name per file.

Cost, `benchstat` over substrate's `shirocore` + `e2e` corpus (667KB, n=8):

| | sec/op | B/op | allocs/op |
|---|---|---|---|
| with cache | ~ (p=0.645) | +2.02% | +1.78% |
| without cache | +17.76% (p=0.001) | +6.61% | +5.97% |

Adversarial worst case (10k *distinct* qualified names) is 3.7x time / 3.2x memory — a bounded constant, linear in input, versus the 2.6GB the un-right-sized scanner would have cost.

## Substrate check — before tightening

Token-level scan of every `.lisp` file for SYMBOL tokens containing `':'`:

- **`luthersystems/substrate`: 93 files, 207 distinct qualified symbols, ZERO with a digit-leading local part.**
- 78 more files across `ui-core`, `elps` and `reliable`: 128 distinct, zero.
- All 171 files still parse clean with the guard in place.
- The minifier cannot produce one either: `newName` is always `fmt.Sprintf("x%d", …)`.

No existing phylum source is affected.

## Mutation verification

- Guard disabled (`if false && …`): all **14** `rejected/` subtests fail; every `accepted/` subtest, the keyword table, the token-split test and the pre-existing `#'` test still pass — the tests are sensitive to the guard and nothing else. Restored, all green.
- Cache key mutated to `strconv.Itoa(len(s))` (collides `"a"` with `"1"`): `TestQualifiedSymbolCheckIsMemoizedPerName` fails with `a cached verdict for the package half must not excuse the local half`, i.e. the parser accepts `a:1` again. Restored, all green.

## Test plan

- `go build ./...`, `go test ./...` — green
- `golangci-lint run ./...` and the `elpscheck`-tagged pass — 0 issues
- `./elps doc -m` — all documented
- `./elps fmt -l --exclude 'grammar' ./...` — clean
- `./elps lint --workspace=. --exclude 'grammar' --include '_examples' ./...` — 9 warnings, identical to `main`
- `bash scripts/ci-gates-test.sh` — 116 passed, 0 failed

`a:1`, `:1`, `a:+1` and `1:1` are added to the adversarial fuzz seed corpus so both sides of the split stay exercised.

---

### DIFF
--- a/docs/lang.md
+++ b/docs/lang.md
@@ -26,6 +26,29 @@ number.
 
 Symbols are currently case sensitive although this may change.
 
+A symbol may also be written in its package-qualified form, `pkg:name` (see
+[Packages](#packages)).  Both halves are identifiers and the rule above applies
+to each of them independently, so `a:1` is not a symbol: the reader rejects it
+with an `invalid-symbol` condition rather than producing a symbol named `1` in
+package `a`, a name no other part of the language can write.  Note that `+1`
+and `.1` *are* identifiers — `+` and `.` are ordinary symbol characters — so
+`a:+1` and `a:.1` are both fine.
+
+### Keywords
+
+A symbol written with a leading colon and no package, `:name`, is a *keyword*.
+
+Keywords are values, not identifiers.  A keyword evaluates to itself, is never
+looked up in a package, and cannot be assigned to — `(set ':k 1)` is an error.
+Because a keyword names no binding, the "identifiers cannot start with a
+number" rule does not apply to it: `:1` is an ordinary keyword whose name is
+`1`, in the same way that `:foo` is one whose name is `foo`.
+
+Keywords are used to label [keyword arguments](#keyword-arguments), and a
+keyword only matches a formal argument spelled the same way.  A keyword whose
+name is not also a legal identifier therefore cannot match any formal argument,
+so `:1` is only useful as data.
+
 ### Numbers
 
 Numbers can be either int or floating point and will be converted between the
@@ -640,6 +663,10 @@ of whether it was exported.
 NOTE:  Qualified access (`pkg:sym`) works for all symbols in a package, not
 just exported ones.  Exports only control what `use-package` imports into the
 caller's namespace — they do not restrict visibility.
+
+NOTE:  Both halves of a qualified symbol must be [identifiers](#symbols).
+Qualified access is another way to spell a name, not a way to introduce one
+that could not be written unqualified, so `a:1` is rejected by the reader.
 
 Scheme-like symbol bindings and assignment are also possible using the `define`
 and `set!` operators.
--- a/internal/fuzzseed/fuzzseed.go
+++ b/internal/fuzzseed/fuzzseed.go
@@ -190,6 +190,13 @@ func Adversarial() [][]byte {
 		"a:",
 		":a",
 		"::",
+		// The local half of a qualified symbol has to be a name; the local
+		// half of a KEYWORD does not (issue #319).  Both spellings reach the
+		// same readSymbol loop, so seed both sides of that split.
+		"a:1",
+		":1",
+		"a:+1",
+		"1:1",
 		strings.Repeat("a", 256),
 
 		// --- comments and hash-bang ---
--- a/parser/rdparser/parser.go
+++ b/parser/rdparser/parser.go
@@ -52,6 +52,7 @@ const DefaultMaxParseDepth = 10000
 // Parser is a lisp parser.
 type Parser struct {
 	src             *TokenSource
+	nameReadback    map[string]bool
 	pendingComments []*token.Token
 	depth           int
 	maxDepth        int
@@ -394,7 +395,7 @@ func (p *Parser) ParseFunRef() *lisp.LVal {
 	// wrong layer to ask: "--" and "-a" lex as NEGATIVE, and only become
 	// symbols when ParseNegative merges them. A token-level guard rejects
 	// those valid names; re-reading through the parser gets them right.
-	if !readsBackAsSymbol(name.Str) {
+	if !p.readsBackAsSymbol(name.Str) {
 		return p.errorf("parse-error",
 			"invalid symbol following #': %q does not read back as a symbol", name.Str)
 	}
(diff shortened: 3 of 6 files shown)


sam-at-luther:
Independently verified before handing this over, rather than taking the description at its word. Two things changed on the branch; neither touches the diff.

**Rebased onto current `main` (`da347a6`).** The branch was cut from `2f0c22a` and so predated #336 and #330 — `mergeable_state` read `behind`. Re-tested after the rebase: `go build ./...`, `go test ./...`, `golangci-lint run ./...` (0 issues), `./elps doc -m` all clean.

**Dropped the `Co-Authored-By` / `Claude-Session` trailers** from both commit messages. No commit in this repo's history carries them. Verified the rewrite changed nothing but the messages: the resulting tree hash is identical to the pre-rewrite one.

## Re-ran both mutations myself

| Mutation | Result |
|---|---|
| `if false && len(pieces) == 2 …` — guard disabled | **14** `rejected/` subtests fail (`a:1`, `a:9`, `a:1b`, `a:1abc`, `a:0x10`, `a:-1`, `a:1.5`, `a:1e5`, `a:1_`, and the `(f a:1 b)` / `[a:1]` / `(a:1)` / `'a:1` / `#'a:1` contexts). Every `accepted/` subtest still passes. |
| cache key → `strconv.Itoa(len(s))` | `TestQualifiedSymbolCheckIsMemoizedPerName` fails, plus 8 `accepted/` subtests and `TestEndPos_FunRef` — the memo test is genuinely measuring the cache, not the guard. |

Both restored; `./parser/...` green.

## Re-ran the corpus scan, and found my first attempt was wrong

Worth recording because it nearly produced a false pass. My initial scan used `elps fmt -l <file>` and reported **1 failure in substrate** (`shirocore.lisp`). That is not a parse failure — `fmt -l` is `gofmt -l`-shaped, so it lists *unformatted* files and exits non-zero for them, and `shirocore.lisp` is the generated minified file. Conflating "unformatted" with "unparseable" would have manufactured a defect that is not there.

Re-ran with `elps fmt <file> >/dev/null`, which exits 1 only on a genuine parse error, after pinning both controls:

- `(a:1)` → exit 1, `invalid-symbol: invalid symbol "a:1": "1" does not read back as a symbol`
- `(a:b   1)` (valid, unformatted) → exit 0

| Repo | `.lisp` files | Parse failures |
|---|---|---|
| `luthersystems/substrate` | 93 | 0 |
| `luthersystems/elps` | 33 | 0 |
| `luthersystems/ui-core` | 12 | 0 |

138 files, zero failures. (`reliable` isn't checked out here, so I confirmed three of the four repos the description cites rather than all four.)

## One extra probe on `NewScannerString`

`make([]byte, len(src))` is a zero-length window for `""`, so I checked the empty, single-byte and multi-byte cases directly before noticing `TestNewScannerString` already covers them — and covers them better than my probe did, since it runs a rune-for-rune differential against `NewScanner` with a non-termination guard. Nothing to add.

Leaving this unmerged for review — it's a language-visible behaviour change.


---

PR #342, opened 2026-08-09
Bound evaluator recursion depth, not just call-stack frames

### LINKED ISSUE #316: MaxHeightPhysical does not bound Go recursion during argument evaluation
Structural observation from the evaluator fuzzing work (PR #311). **No exploit — reported as a gap in the containment model, not as a demonstrated crash.**

## The gap

`CallStack.MaxHeightPhysical` exists to stop unbounded non-tail recursion from exhausting the Go goroutine stack, which aborts the process with an unrecoverable `fatal error: stack overflow` that no `handler-bind` can catch. It is the memory guard, and its default (25,000) is deliberately set ~8-16x below the measured crash point.

But `evalSExprCells` recurses into `eval` to evaluate arguments **before any frame is pushed**. So nesting like:

```
((lambda (x) x) ((lambda (x) x) ... 1 ...))
```

consumes Go stack while the ELPS call stack height stays at 0. `MaxHeightPhysical` never sees it.

## What bounds it today

`rdparser.DefaultMaxParseDepth` (10,000) — incidentally, because you have to *parse* the nesting before you can evaluate it. 9,000 levels was measured as survivable.

## Why this is only an observation

I could not construct a deeper case from ELPS at runtime: list builders return quoted values that `eval` returns unchanged, so the nesting has to come from source, and source is capped by the parse-depth limit. So today the two limits happen to compose.

The concern is that they compose **by accident**. Anything that lets a deeply-nested s-expression reach `eval` without going through the parser — a macro that constructs nesting programmatically, an embedder building an AST directly via the Go API, a future reader — removes the only thing currently bounding it. An embedder that raises `MaxParseDepth` also silently removes it.

## Possible directions

- Count Go-recursion depth in `evalSExprCells` against `MaxHeightPhysical` (or a sibling limit), so the guard covers the path it was written for.
- Or document explicitly that `MaxHeightPhysical` bounds *frames*, not evaluation depth, and that `MaxParseDepth` is load-bearing for stack safety — so nobody raises it casually.

The second is cheap and would at least stop the accident being invisible.

## Related

- `Runtime.MaxAlloc` is per-operation, not cumulative: nothing bounds total heap across many small allocations. The default `MaxTailIterations` (1e6) is the incidental bound there. Already documented in `runtime.go`; noting it here as the same shape of gap.
- Documented on the `deep-nested-lambda-call` seed in the eval fuzz corpus.

sam-at-luther:
Closes #316.

## Direction: (1), the counter — and the gap is live, not latent

The issue offered two directions: count Go-recursion depth against a limit, or
document that `MaxHeightPhysical` bounds frames rather than evaluation depth.
This PR does (1), because establishing the ground truth first turned the
finding from "latent gap" into "live defect".

### Ground truth (measured, linux/amd64, Go 1.25, 1GB default goroutine stack)

**1. `MaxHeightPhysical` really does not see argument-evaluation depth.**
With the stock default of 25,000, argument nesting of 400,000 levels evaluates
to completion and the ELPS call stack is *empty* throughout. Under a
deliberately tiny 5-frame ceiling — low enough that 10 levels of ordinary
non-tail recursion is refused — 5,000 levels of argument nesting still
evaluate fine. Pinned by `TestArgumentNestingIsNotBoundedByStackHeight`.

**2. The real Go crash point, and where `DefaultMaxParseDepth` sits.**
Bisected in subprocesses: 700,000 levels of argument nesting completes,
800,000 aborts with `fatal error: stack overflow` (`goroutine stack exceeds
1000000000-byte limit`), i.e. ~1.4KB of Go stack per level. `DefaultMaxParseDepth`
(10,000) is ~75x below that, so the issue's account of the two limits
composing by accident is correct as far as it goes.

**3. Deep nesting IS reachable without the parser — two ways. This is the
headline, and it changes the priority.**

The issue reports the author could not construct a deeper case from ELPS at
runtime, because list builders return quoted values that `eval` returns
unchanged. That observation is correct — I reproduced it, `(eval (build N 1))`
via `list` fails with `first element of expression is not a function:
'identity` at every depth, because `builtinEval` shallow-unquotes only the
outermost level. But it does not close the door:

* **A recursive macro does it.** Macro arguments are passed *unevaluated and
  unquoted*, and a macro expansion is re-evaluated by `eval`. So a macro can
  emit one level of nesting per expansion and recurse on an integer:

  ```lisp
  (defmacro nest (n)
    (if (<= n 0) 1 (quasiquote (identity (nest (unquote (- n 1)))))))
  (nest 800000)
  ```

  ~120 bytes of source, constant size at any depth, no parse nesting at all.
  Against `main`, at the defaults an embedder gets from `lisp.NewEnv(nil)`,
  **that program kills the host process with `fatal error: stack overflow`.**
  It also slips past `MaxMacroExpansionDepth` (1,000), because each expansion
  happens at a *different* `eval` frame, so the per-frame `macroDepth` counter
  never exceeds 1.

* **The Go API does it**, as the issue anticipated: `lisp.SExpr` nested
  800,000 deep, evaluated with `env.Eval`, same fatal abort.

So this is not two limits composing by accident. It is one limit that a
constant-size ELPS program walks straight past. That is why this PR ships the
guard rather than the doc note.

## The fix

Count `eval`'s own recursion depth. Every nested evaluation passes through
`LEnv.eval` — a call's arguments, a special operator's subforms, a builtin
re-entering via `env.Eval` — so a single counter incremented on entry and
decremented in the `defer` that is already there bounds all of them.

Deliberately a counter and **not** a stack frame, which is what
`evalSExprCells` is avoiding when it says *"We don't want to push anything on
the stack here because that would cause improper error messages/stack-dumps"*.
I tested that claim rather than trusting it: pushing a frame per argument
evaluation instead of counting fails **13 tests** —

* `TestStack`: the reported function name changes, `recursive: physical stack
  height exceeded` becomes `argeval: physical stack height exceeded`;
* `TestSpecialOp`: `debug-stack` gains a phantom `argeval` frame
  (`Stack Trace [4 frames]` → `[5 frames]`);
* 7 debugger tests across `x/debugger`, `dapserver` and `debugrepl`:
  step-into resolves to the wrong line, because stepping is keyed off stack
  depth.

The comment is accurate; a counter threads the needle.

New API: `Runtime.MaxEvalNesting` / `lisp.WithMaxEvalNesting(n)` /
`lisp.DefaultMaxEvalNesting` (100,000) / condition `eval-nesting-exceeded` /
`Runtime.EvalNesting()`. Zero means "use the default" (so a hand-built
`Runtime` is guarded, not unguarded); negative disables.

### Why 100,000

* 7-8x below the measured 700k–800k crash band — the same order of margin
  `DefaultMaxPhysicalStackHeight` keeps (8-16x), and the margin matters because
  the crash point moves with Go's stack settings, per-frame size and
  architecture.
* Above ordinary recursion, so `MaxHeightPhysical` stays the binding
  constraint for ordinary recursive code. Measured with
  `(defun sum (n) (if (<= n 0) 0 (+ n (sum (- n 1)))))`: the physical limit
  wins for any `MaxEvalNesting` >= 38,000, i.e. that recursion costs ~1.5 eval
  levels per physical frame and reaches height 25,001 at nesting ~38,000. Code
  must nest expressions more than ~4 deep per recursion level to reach 100,000
  nesting before 25,000 frames — 4 eval levels per frame is ~2.6x what ordinary
  recursion costs, so such code really is using that much more Go stack.

## Benchmarks

`eval` is the hottest path in the interpreter, so the cost is one increment,
one compare and one decrement per `eval` call, with no allocation and no new
`defer` (the decrement goes inside the recover handler that already exists).

Measured exactly as CI does — two trees, interleaved round-robin so
time-varying machine load hits both arms equally, `benchstat` over the pair:

**Full suite, `-benchmem -benchtime=100ms -count=1` x 10 rounds, all packages
(142 rows):**

| metric | geomean delta | significant bad-direction moves |
|---|---|---|
| sec/op (`lisp`) | +1.03% | 0 |
| B/op (`lisp`) | -0.00% | 0 |
| allocs/op (`lisp`) | +0.00% | 0 |

`scripts/benchstat-gate.sh` verdict: *interpreted 11 delta rows + 131
no-change rows; 0 significant moves in the bad direction; 0 at or above the
gate.* Every per-benchmark row is `~` (p > 0.05). Allocation counts are
byte-identical — the guard allocates nothing.

**Focused re-run on the eval-heavy benchmarks (`EnvGet`, the seven
`EnvFunCall*`, `MacroDefun`), `-benchtime=200ms -count=1` x 20 interleaved
rounds:**

| benchmark | base sec/op | PR sec/op | verdict |
|---|---|---|---|
| `EnvGet` | 16.41m ± 18% | 16.82m ± 11% | ~ (p=0.820, n=20) |
| `EnvFunCallBuiltin` | 932.6µ ± 3% | 959.4µ ± 7% | ~ (p=0.478, n=20) |
| `EnvFunCallRecursion` | 11.33m ± 5% | 11.15m ± 5% | ~ (p=0.841, n=20) |
| `EnvFunCallTailRec` | 5.517m ± 10% | 5.452m ± 5% | ~ (p=0.820, n=20) |
| `EnvFunCallPos` | 2.519m ± 8% | 2.580m ± 7% | ~ (p=0.620, n=20) |
| `EnvFunCallKey` | 2.975m ± 8% | 2.905m ± 8% | ~ (p=0.355, n=20) |
| `EnvFunCallOpt` | 2.514m ± 9% | 2.590m ± 7% | ~ (p=0.758, n=20) |
| `EnvFunCallVar` | 3.303m ± 10% | 3.190m ± 5% | ~ (p=0.253, n=20) |
| `MacroDefun` | 44.00µ ± 5% | 43.97µ ± 4% | ~ (p=0.659, n=20) |
| **geomean** | 2.398m | 2.403m | **+0.21%** |

`B/op` and `allocs/op` geomeans: -0.00% / +0.00%, every row `~`, seven of nine
`allocs/op` rows flagged "all samples are equal".

This sandbox is noisy (per-row variance ±10-40%, versus ~1.5% on the CI
fleet), so treat these as an upper bound rather than a precise figure: the
measurement can exclude a large regression, and it does, but it cannot resolve
a sub-1% one. What it can settle exactly is the allocation metrics, which are
deterministic and unchanged.

## Mutation verification

Each new guard was reverted in place and the suite re-run, to confirm the
tests actually fail without it:

| mutation | result |
|---|---|
| delete the `evalNestingExceeded` check in `eval` (keep the counter) | 6 tests fail; `TestDefaultLimitsContainFatalNesting` does not merely fail — **the test binary dies with `runtime: goroutine stack exceeds 1000000000-byte limit` / `fatal error: stack overflow`**, which is the defect itself, reproduced on demand |
| `>` → `>=` in the limit comparison (off-by-one) | `TestEvalNestingLimitBoundary/depth_99` and `/depth_100` fail |
| delete the `evalNesting--` decrement (counter leaks) | 4 tests fail, including every `TestEvalNestingCounterUnwinds` subtest — a leaked counter degrades a long-lived `Runtime` into one that refuses to evaluate anything |

The naive alternative — a stack frame instead of a counter — was mutation-tested
the same way and fails 13 tests; see above.

## Also addressed: `Runtime.MaxAlloc` is per-operation

Documented rather than changed, and I think that is the right call. Making it
cumulative would mean tracking every allocation and every release across the
interpreter — a real allocator — and it still would not bound memory the host
does not route through `CheckAlloc` (Go's own map growth, the `LVal` tree
itself). The honest statement is that ELPS has no total-memory bound and the
host must supply one: a cgroup, a container limit, or `GOMEMLIMIT` plus a
watchdog. That now appears in the `DefaultMax*` block in `lisp/runtime.go`
under "Bounding total memory" and in the execution-limits section of
`docs/lang.md`, alongside the existing "none of these bound elapsed time"
note, so the two holes in the containment model are stated in the same place.

## Test Plan

- [x] `make test` passes
- [x] `make static-checks` (`golangci-lint run ./...`) passes
- [x] `./elps fmt -l ./...` reports no changes
- [x] `./elps lint ./...` unchanged from `main` (byte-identical output)
- [x] `./elps doc -m` reports no missing docs
- [x] `make race`, `make test-elpscheck`, `bash scripts/ci-gates-test.sh` pass
- [x] new guard mutation-verified three ways

Base for the diff is `da347a6` (rebased onto current `main`, which moved during
the work). All run in the sandbox, all clean:

| check | result |
|---|---|
| `go build ./...` | pass |
| `go test ./...` | pass |
| `make test` (Go tests + `_examples` lisp) | pass |
| `make race` | pass, no data races |
| `./elps fmt -l ./...` | no files listed |
| `bash scripts/ci-gates-test.sh` | 116 passed, 0 failed |
| `make bench-gate` on the benchstat output | 0 significant moves in the bad direction |

`./elps lint ./...` exits 1 on both `main` and this branch with byte-identical
output (44 lines, pre-existing diagnostics in example/test lisp sources), so it
is unchanged by this PR.

## Files

* `lisp/env.go` — the counter and the check in `eval`.
* `lisp/runtime.go` — `MaxEvalNesting`, `DefaultMaxEvalNesting`, accessors,
  and the measurement record in the `DefaultMax*` doc block.
* `lisp/config.go` — `WithMaxEvalNesting`.
* `lisp/conditions.go` — `CondEvalNestingExceeded`.
* `lisp/evalnesting_test.go` — new.
* `lisp/eval_fuzz_test.go`, `internal/fuzzseed/evalseed.go` — the
  macro-generated shape as a runaway seed, plus an explicit nesting budget for
  the harness. Against `main` that seed was stopped only by the harness's 2s
  wall-clock deadline, after hundreds of thousands of Go frames had already
  been pushed; it now trips the nesting limit deterministically in ~250ms.
* `docs/lang.md`, `parser/rdparser/parser.go` — documentation.

---

### DIFF
--- a/docs/lang.md
+++ b/docs/lang.md
@@ -942,11 +942,17 @@ an embedder can identify the offending Go function.
 
 ## Execution Limits
 
-ELPS bounds evaluation with four independent mechanisms: **context
-cancellation**, **step limits**, **stack height limits** and a
-**tail-iteration limit**.  Context cancellation and step limits are optional
-and impose negligible overhead when not configured; the physical stack limit
-and the tail-iteration limit are on by default.
+ELPS bounds evaluation with five independent mechanisms: **context
+cancellation**, **step limits**, **stack height limits**, an **evaluation
+nesting limit** and a **tail-iteration limit**.  Context cancellation and step
+limits are optional and impose negligible overhead when not configured; the
+physical stack limit, the evaluation nesting limit and the tail-iteration
+limit are on by default.
+
+None of them bound *total* memory: `Runtime.MaxAlloc` caps the output size of
+a single builtin call, not the sum across calls, so a loop that allocates many
+smaller values is bounded only by whatever stops the loop.  A host that must
+bound total memory has to do it outside the interpreter.
 
 ### Context Cancellation
 
@@ -1010,9 +1016,9 @@ inside a builtin.  **Context cancellation with a deadline is the only limit
 here that measures elapsed time**, and it is what you want if the real
 requirement is "give up after N seconds".
 
-### Stack Height and Tail-Call Limits
+### Stack Height, Nesting and Tail-Call Limits
 
-ELPS distinguishes three things that a naive "stack limit" conflates.
+ELPS distinguishes four things that a naive "stack limit" conflates.
 
 **Physical stack height** is the number of frames actually on the call
 stack.  This is the memory guard: unbounded *non-tail* recursion exhausts the
@@ -1023,6 +1029,42 @@ below the measured crash threshold, and exceeding it produces an ordinary,
 catchable ELPS error.  Override with
 `lisp.WithMaximumPhysicalStackHeight(n)`; 0 disables the check, which is not
 recommended.
+
+It bounds *frames*, not evaluation depth — see below.
+
+**Evaluation nesting** is how deeply the evaluator recurses into itself, which
+is the true measure of Go stack consumed.  It is not the same as stack height
+and is not implied by it: a call's arguments are evaluated *before* the call's
+frame is pushed, so
+
+```lisp
+(identity (identity (identity ... 1)))
+```
+
+recurses through the whole evaluator while the physical stack height stays at
+**zero**.  That is the exact shape the physical limit exists to stop and the
+one shape it cannot see.  Nesting does not have to be written out by hand
+either — a recursive macro generates it at expansion time from an integer, so
+neither the parser's depth limit nor the source size bounds it:
+
+```lisp
+(defmacro nest (n)
+  (if (<= n 0) 1 (quasiquote (identity (nest (unquote (- n 1)))))))
+(nest 800000)   ; without the nesting limit: fatal error: stack overflow
+```
+
+Nesting is bounded by default (`DefaultMaxEvalNesting`, 100000), several times
+below the measured crash threshold and well above the nesting an ordinary
+recursion reaches before it hits the 25000-frame physical limit.  Exceeding it
+raises a catchable `eval-nesting-exceeded` condition:
+
+```lisp
+(handler-bind ((eval-nesting-exceeded (lambda (c &rest args) 'too-deep)))
+    (nest 800000))
+```
+
+Override with `lisp.WithMaxEvalNesting(n)`; a negative value disables the
+check, which re-exposes the host process to an unrecoverable stack overflow.
 
 **Tail-call iterations** count the turns of a tail-recursive loop.  Tail
 calls run in constant stack space, so no stack-height limit can bound a
--- a/internal/fuzzseed/evalseed.go
+++ b/internal/fuzzseed/evalseed.go
@@ -48,6 +48,20 @@ func EvalRunaway() map[string]string {
 		"non-tail-recursion-forever": `(defun rec (n) (+ 1 (rec n))) (rec 0)`,
 		"non-tail-mutual":            `(defun p (n) (+ 1 (q n))) (defun q (n) (+ 1 (p n))) (p 0)`,
 
+		// --- unbounded ARGUMENT nesting (MaxEvalNesting).  Same failure
+		// mode as the two above -- Go stack exhaustion -- reached by a
+		// shape no stack-height limit can see: evalSExprCells evaluates a
+		// call's arguments BEFORE pushing the call's frame, so this
+		// recurses through the whole evaluator at physical height zero.
+		//
+		// The nesting is generated during macro expansion from an integer,
+		// so the source is constant-size and rdparser's parse-depth limit
+		// -- which was the only thing incidentally bounding this -- never
+		// sees it.  Measured against the pre-fix tree at stock defaults,
+		// the 800000 form aborted the process with "fatal error: stack
+		// overflow" (issue #316).  MaxEvalNesting is what stops it now. ---
+		"macro-generated-arg-nesting": `(defmacro nest (n) (if (<= n 0) 1 (quasiquote (identity (nest (unquote (- n 1))))))) (nest 800000)`,
+
 		// --- non-recursive infinite loop.  Neither stack limit can see this:
 		// it grows no frames and performs no tail calls.  Only a step budget
 		// or a context deadline stops it. ---
--- a/lisp/conditions.go
+++ b/lisp/conditions.go
@@ -19,6 +19,12 @@ const (
 const (
 	CondContextCancelled  = "context-cancelled"
 	CondStepLimitExceeded = "step-limit-exceeded"
+
+	// CondEvalNestingExceeded reports that the evaluator recursed into
+	// itself more deeply than Runtime.MaxEvalNesting allows.  It is the
+	// recoverable substitute for a Go stack overflow, which is a
+	// runtime.throw that neither recover() nor handler-bind can intercept.
+	CondEvalNestingExceeded = "eval-nesting-exceeded"
 )
 
 // CondInternalPanic is the condition type of an error produced by recovering
--- a/lisp/config.go
+++ b/lisp/config.go
@@ -41,6 +41,27 @@ func WithMaximumPhysicalStackHeight(n int) Config {
 	}
 }
 
+// WithMaxEvalNesting returns a Config that bounds how deeply the evaluator may
+// recurse into itself while evaluating a single expression.
+//
+// This is a distinct quantity from stack height and is not implied by it.  A
+// call's arguments are evaluated before the call's frame is pushed, so
+// ((lambda (x) x) ((lambda (x) x) ... )) recurses through the Go evaluator
+// while the physical stack height stays at zero — the exact shape
+// WithMaximumPhysicalStackHeight exists to stop and the one shape it cannot
+// see (issue #316).  Nesting is bounded here instead.
+//
+// A value of 0 selects DefaultMaxEvalNesting.  A negative value disables the
+// check, which re-exposes the host process to an unrecoverable
+// "fatal error: stack overflow"; do that only when some other bound on
+// expression depth is guaranteed.
+func WithMaxEvalNesting(n int) Config {
+	return func(env *LEnv) *LVal {
+		env.Runtime.MaxEvalNesting = n
+		return Nil()
+	}
+}
+
 // WithMaxTailIterations returns a Config that bounds the number of tail-call
 // iterations a single stack frame may perform.  Tail calls run in constant
 // stack space, so neither stack-height limit can bound a runaway tail loop;
--- a/lisp/env.go
+++ b/lisp/env.go
@@ -937,11 +937,23 @@ func (env *LEnv) Eval(v *LVal) *LVal {
 // evaluation into an LError, preventing panics from crashing the host process
 // when ELPS is embedded.
 //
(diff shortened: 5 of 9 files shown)


sam-at-luther:
Verified independently rather than taking the description's word for it. The headline claim holds, and it is the important one — this reclassifies #316 from latent to live.

## The macro path reproduces on current `main`

Built `main` at `da347a6` and ran the ~120-byte program at stock `lisp.NewEnv(nil)` defaults:

```lisp
(defmacro nest (n)
  (if (<= n 0) 1 (quasiquote (identity (nest (unquote (- n 1)))))))
(nest 800000)
```

```
runtime: sp=0xc030b00388 stack=[0xc030b00000, 0xc050b00000]
fatal error: stack overflow
```

Exit 2. That is a `runtime.throw` — no `recover()`, no `handler-bind`, the host process is gone. Reachable from ordinary phylum source with every documented limit at its default.

Same input on this branch:

```
error: eval-nesting-exceeded: lisp:quasiquote: evaluation nesting depth exceeded maximum: 100001 …
  --> zz_nest.lisp:2:55
 2 |    (if (<= n 0) 1 (quasiquote (identity (nest (unquote (- n 1)))))))
   |                                                        ^
```

A recoverable condition with a source location. Both directions confirmed on this machine.

## Mutation re-run

Deleting the check (`if false && env.Runtime.evalNestingExceeded()`) does not merely fail assertions — the **test binary itself dies**:

```
--- FAIL: TestEvalNestingLimitBoundsArgumentNesting
--- FAIL: TestMacroGeneratedNestingIsBounded
fatal error: stack overflow
FAIL	github.com/luthersystems/elps/lisp	8.061s
```

The suite reproduces the defect on demand, which is a stronger property than a failing assertion. Restored, green.

## One review note: the default's margin is pointer-size dependent

Not a blocker, and I have not changed anything — but the rationale in `runtime.go` says "Measured on linux/amd64 against the 1GB default goroutine stack" and stops there. Checked against the Go source on this machine (`runtime/proc.go:155-162`):

```go
// Max stack size is 1 GB on 64-bit, 250 MB on 32-bit.
if goarch.PtrSize == 8 {
    maxstacksize = 1000000000
} else {
    maxstacksize = 250000000
}
```

At the measured ~1.4KB of Go stack per nesting level, `DefaultMaxEvalNesting = 100000` costs ~140MB. That is the stated 7-8x margin on a 64-bit build, but roughly **1.8x** on a 32-bit one — and an embedder whose own call stack is already deep when it enters `Eval` eats into what is left. CI builds only `ubuntu-latest` and `windows-latest`, both amd64, so nothing here is exercised on 32-bit and nothing in Luther's fleet ships 32-bit.

Since this is a public API default rather than an internal constant, I'd suggest one sentence in the `DefaultMaxEvalNesting` doc saying the margin is measured on 64-bit and that a 32-bit embedder should lower it via `WithMaxEvalNesting`. Your call whether that is worth a commit.

## Everything else checked

- No `Co-Authored-By` / `Claude-Session` trailers on any of the four commits (no commit in this repo's history carries them).
- Branch is based on current `main` (`da347a6`).
- CI so far: `Required: build & test`, `Lint & Test`, `Build (Windows)`, `Required: govulncheck`, `Required: tree-sitter`, `CI Gate Self-Test` all green. Benchmark Comparison and the four fuzz shards still running — the benchmark one is the check worth watching, since the counter sits on the hottest path in the interpreter.

Leaving this unmerged for your review.


---

Issue #354, opened 2026-08-10
astutil.SourceOf nil-derefs on the nil parent Walk documents it will pass


sam-at-luther:
`astutil.Walk` documents that it passes a nil parent, and `astutil.SourceOf` dereferences its argument unconditionally. The two are exported side by side from `lint/walk.go`, and pairing them is the obvious thing to do.

`astutil/walk.go:12`:

```go
// parent is nil for top-level expressions.
```

`astutil/walk.go:133`:

```go
func SourceOf(v *lisp.LVal) *lisp.LVal {
	if v.Source != nil && v.Source.Line > 0 {
```

## Reproducer

```go
astutil.Walk(exprs, func(node, parent *lisp.LVal, _ int) {
	_ = astutil.SourceOf(parent)
})
```

Segfaults on the first top-level form.

## Why it is worth fixing despite being small

It needs an embedder to pass nil, so nothing in-tree hits it. But the nil is not an accident on the caller's part — it is the documented contract of the function directly above it in the same file, and "find the source location of this node's parent" is a natural thing for a lint rule to want.

`SourceOf`'s own contract also has no sensible answer for nil other than nil, so returning nil is both safe and meaningful; there is no behaviour to preserve.

## Suggested fix

Return nil for a nil argument, and say so in the doc comment. Worth a quick scan for sibling helpers in `astutil` with the same shape.

Found by the `analysis`/`lint` fuzzing work for #318, on its harness's first run. Verified independently against `main` at 5bb31c9.

Issue #353, opened 2026-08-10
analysis.AnalyzeFile silently drops Config.MacroExpander


sam-at-luther:
`AnalyzeFile` rebuilds its `Config` field by field and omits `MacroExpander`. `analysis/workspace.go:593`:

```go
fileCfg = &Config{
    ExtraGlobals:   fileCfg.ExtraGlobals,
    PackageExports: fileCfg.PackageExports,
    PackageSymbols: fileCfg.PackageSymbols,
    DefForms:       fileCfg.DefForms,
    PackageImports: fileCfg.PackageImports,
    DefaultPackage: fileCfg.DefaultPackage,
    WorkspaceRefs:  fileCfg.WorkspaceRefs,
    Filename:       filename,
}
return Analyze(exprs, fileCfg)
```

`MacroExpander` is a documented field of `Config` ("optionally expands user-macro calls at analysis time") and `AnalyzeFile` takes a `*Config`, advertising no exception. A caller that sets it gets it silently ignored.

## Reproducer

```go
src := []byte("(in-package 'user)\n(no-such-head 1 2)\n")

exprs := rdparser.New(token.NewScanner("x.lisp", bytes.NewReader(src))).
    ParseProgramFaultTolerant().Exprs
Analyze(exprs, &Config{MacroExpander: e1, Filename: "x.lisp"})  // e1 called 1 time
AnalyzeFile(src, "x.lisp", &Config{MacroExpander: e2})          // e2 called 0 times
```

## Consequence

The two `AnalyzeFile` callers both compute cross-file information:

- `ScanWorkspaceRefs`, `analysis/workspace.go:829`
- the LSP's `updateFileRefs`, `lsp/server.go:537`

Both therefore analyse without macro expansion, while per-document analysis (`lsp/document.go:80`, `lint/lint.go:648`) analyses *with* it. So a symbol reachable only through a macro expansion is visible to one and invisible to the other, which is a plausible source of inconsistent references, rename and call-hierarchy results across files.

## This may be deliberate

Expanding macros for every file during a workspace-wide reference scan is expensive, and skipping it may be an intentional trade. If so the fix is a comment and an explicit field rather than a silent omission — for example taking the expander as a separate parameter, or a `Config` flag named for the decision, so the difference is visible at the call site.

If it is not deliberate, the fix is to carry the field through.

Either way the current state is the worst of the three: the field is accepted, documented, and dropped.

[...]

PR #358, opened 2026-08-10
Make swallowed macro-expansion panics observable, and fuzz analysis and lint directly

### LINKED ISSUE #318: Fuzzing follow-up tracker: what round 1 and 2 did not cover
Umbrella for the coverage gaps left by the fuzzing work in PR #311. Fuzzing found **22 defects** across two rounds; this records what it deliberately did not reach, so the gaps are visible rather than implied by silence.

Individually-ticketed findings: #314 (`time:sleep` unbounded), #315 (goparsec `scanString` panic), #316 (`MaxHeightPhysical` vs Go recursion), #317 (`invalid-float` condition name).

> ### State of this tracker, against `main` = 663c0f3
>
> Bullets that merged work has closed are struck through below; the reasoning is kept rather than deleted, so this still reads as a record of what was and was not reached. `main` now carries **20 fuzz targets over 5 shards**, which is exactly the five-shard ceiling (`ceil(20/5) = 4`).
>
> Closed since filing: the `lisp/x/debugger` half of the first bullet (`FuzzDebugEval`); `lsp/` (#348, `FuzzLSPSession`); `parser/regexparser` deleted outright (#321, 84d09b8); the carried-forward `FuzzFormat` mystery (#335); `elpsutil/` (#357, `FuzzElpsutilEmbed`); the thin-stdlib budget and the two fingerprint-exclusion bullets (#356, `FuzzApplyThinStdlib` + `internal/fuzzfp`); and, arriving from outside this tracker, the cyclic-value process kill (#390 → #391, `FuzzCyclicValueWalks`).
>
> Still in flight: **#358** — dedicated `analysis/` and `lint/` targets, and the **sixth shard** the 21st target requires. Not closed here until it merges.

## Unfuzzed packages

- ~~**`lisp/x/debugger`**~~ — **closed.** `FuzzDebugEval` is on `main`. The premise recorded here checked out and is sharper than written: TRO-disable and macro-expansion stamping both key off `Runtime.Debugger != nil`, not `IsEnabled()`, so a *dormant* debugger already gets a materially different eval path. `TestDebuggerChangesTheEvalPath` pins it.
  **`lisp/x/profiler` is still unfuzzed** and is the surviving half of this bullet. Deliberately deprioritised rather than skipped: its only hook is `env.trace`, it changes no budget's identity and mutates no `LVal`, and `callgrindProfiler` calls `runtime.ReadMemStats` twice per traced call, which stops the world.
- ~~**`lsp/`**~~ — **closed** by #348 (`FuzzLSPSession`). One correction to the reason recorded here, kept because it would otherwise be rediscovered: this repository pins `TextDocumentSyncKindFull` and `textDocumentDidChange` takes the last content change as the whole document, so there is **no incremental edit splicer** and none of the range-splicing arithmetic that would make one worth fuzzing. The real case was stronger — client-supplied positions converted by hand between LSP 0-based, ELPS 1-based `token.Location` and byte offsets (no other target in this repository supplies a position at all), plus `lsp/` being the only route into `analysis/` and `lint/`.
- ~~**`elpsutil/`**~~ — **closed** by #357 (`FuzzElpsutilEmbed`), which took the embedding surface off its 6.2% coverage baseline.
- **`analysis/`** and **`lint/`** — **not closed.** `FuzzLSPSession` drives both transitively (34.1% and 38.7% mean per-function from seeds alone), but the dedicated targets — `FuzzAnalyzeSource`, `FuzzWorkspaceScan`, `FuzzLintSource` — are on the unmerged #358. A downgrade, not a close, until that lands.
- **`mcpserver/`**, **`repl/`** — still unfuzzed.
- **`tree-sitter-elps`** — separate Go module; `./...` from the repo root does not descend into it, so no target and no lint. Unchanged.

## Known-thin coverage in packages that ARE fuzzed

- ~~**`libtesting`, `libhelp`, `libgolang` share one target's budget**~~ — **closed** by #356. `FuzzApplyThinStdlib` gives those 19 callables a dedicated budget. They are 8% of the index space, so the shared budget gave them roughly 8% of a target's executions; this is a second full budget on top, not a partition, with a per-package coverage floor (a total floor loose enough not to be brittle would not notice `help` or `golang` vanishing entirely).
- ~~**`regexparser` string literals are entirely unfuzzed**~~ — **the bullet no longer names anything.** #321 (84d09b8) deleted `parser/regexparser` and its fuzz target the same day this tracker was filed.
- ~~**`LNative` contents are excluded from the mutation fingerprint**~~ and ~~**`LVal.Source` is excluded from the fingerprint**~~ — **both bullets were recorded WRONGLY, and are now closed.** There was no mutation fingerprint, in the tree or in its history: `FuzzApplyStdlib` verified the three shared singletons via `lisp.SingletonSnapshot` and nothing else, so no argument was watched at all — `LBytes` and `LSortMap` were not "handled specially and caught" either. #356 built `internal/fuzzfp` (now on `main`) as that check, with both named exclusions *narrowed at construction rather than adopted*:
  - **LNative** contents are walked reflectively, rendering map entries independently and sorting the renderings, reading unexported fields through kind-specific accessors, terminating cycles on pointer identity, and rendering pointers as dense per-walk indices. Determinism is proven by 2000 in-process repetitions per value rather than asserted; deleting the sort fails it.
  - **Source** is watched under the rule the legitimate writers actually obey — `macroCall` and `LEnv.errorSource` both guard on `Source == nil || Source.Pos < 0` — so a *real* location may never be replaced, and an unchanged `Source` pointer must still point at the same `Location` value. That second clause is the one with the widest blast radius, since nearly every value in the process shared one `token.Location` and `SingletonSnapshot` compared `Source` by pointer.
  Deliberately still NOT asserted: that arguments are unchanged (`append!`, `assoc!` and `MacroCall` legitimately write), and the nil-or-synthetic `Source` case. `TestArgumentGuardIsWiredIn` is the negative control.
- **The value generator produces only well-formed values.** By design (a malformed `LArray` is unreachable from lisp), but it means a builtin that corrupts an `LVal` internally and hands it to another builtin is not modelled. Unchanged.
- **Native values are limited to what the stdlib actually produces**, plus one embedder struct. Narrowed by #356 — `time.Time`, `time.Duration`, `*regexp.Regexp` and `json.RawMessage` are now in the corpus, where previously every native in it was a shape no stdlib function has a branch for — but an embedder injecting a type whose `String()` panics is still unmodelled.
- **No `FuzzEval` in the builtins work** — the evaluator's own loops are reached only incidentally there. Covered separately by the eval targets, but the two do not compose into one campaign. Unchanged.

## Structural limits

- **All targets take raw `[]byte` or a fuzzer-driven script.** The grammar-aware generator (`internal/fuzzgen`) covers the parser pipeline; nothing equivalent exists for the evaluator or the builtins. Unchanged.
- ~~**The `Core` dialect used by the differential target is hand-derived.**~~ — **struck: this bullet no longer names anything.** Neither `TestCoreDialectAlwaysParses` nor `TestGeneratorReachesItsTargetConstructs` exists on `main`, and `CoreDialect` appears in no commit reachable from `origin/main`. #321 removed the premise by deleting `parser/regexparser`, so the second parser a differential target would compare against is gone; `internal/fuzzgen`'s package documentation records the consequence and deliberately carries only one grammar.
- **The parser-equivalence proof is evidence, not proof** — exhaustive to 3 bytes over a 12-symbol alphabet plus a sampled tail. It cannot cover deep nesting (the pre-change grammar is exponential there) or strings. Unchanged, and not re-verified against the current tree.
- **Concurrency is unfuzzed**, correctly — `Runtime` is documented single-goroutine. Unchanged.

## Campaign depth

Everything to date ran on one contended 4-core sandbox, longest ~10 minutes per target. The nightly schedule (10m/target) is where real depth will come from. **Treat "no further crashes" as weak evidence, not a clean bill of health** — the special-operator corpus was still growing when its agent stopped.

**This caveat still stands, and every campaign added since has confirmed it rather than retired it.** #348 finished at +534 new interesting inputs in its second sweep; #358 had 431 and 419 still arriving at its 10-minute cutoffs; #356's two targets were still growing, decelerating, at every cutoff across 2.26M executions. Nothing here is saturated.

## ~~One unexplained result, carried forward~~ — explained, and the gate now says so

The single `FuzzFormat` failure that wrote no crasher was **not** a fuzzing worker-process death. It is a race in the fuzzing *coordinator*, diagnosed in #335: `internal/fuzz.CoordinateFuzzing` suppresses the `-fuzztime` deadline by comparing the parent context's error against the child's, and `cancelCtx.cancel` stores the parent's error and closes its `done` channel *before* walking its children, so a preemption in that ~50ns window makes the suppression miss. Target-independent by construction; demonstrated by widening the window under `go test -overlay` and reproducing it on five targets, with unpatched controls passing.

The gate change landed with it: `scripts/fuzz.sh` on `main` diffs `testdata/fuzz/<Target>/` around each target and classifies the outcome as `CRASHER` / `SEED` / `HARNESS` / `UNATTRIBUTED`, so "no crasher was produced" is stated in the job log rather than left to be inferred, and `make fuzz-classify-test` guards the classifier. The artifact upload no longer sweeps up the whole committed corpus.

## Found and fixed after this tracker was filed, recorded here so the sweep's history is in one place

- **Cyclic values killed the process** (#390). `(set 'm (sorted-map)) (assoc! m "k" m) m` — 32 bytes of legal lisp — overflowed the Go stack in `LVal.String()`, which is a fatal runtime error `recover()` cannot intercept, so the evaluator could not turn it into a condition. Same shape in `equal?` and `json:dump-string`. Found in about 7 seconds by the container-family targets on #389, and **`FuzzEval` on `main` was exposed to it**, so a nightly sweep could have gone red on it at any time. Fixed by #391 (`050ade5`): the recursive walks are bounded, and `lisp.FuzzCyclicValueWalks` is now a target on `main`.
- **#350** belongs on the open list, as that issue itself asks: `json:load` decodes every JSON number through `float64`, so integers above 2^53 are silently rounded and no combination of existing options round-trips a large integer as a bare JSON number. Characterised by the fuzzing work and then deliberately scoped out of the harness — asserting numeric-text preservation would have been permanent fuzz noise — but until now it was recorded only in a comment in `lisp/lisplib/libjson/fuzz_test.go` and two seed files, which is the rediscovery-from-scratch failure mode this tracker exists to prevent.

## Shard budget

`main` sits at **20 targets over 5 shards** — `ceil(20/5) = 4`, so `needed` is 50 min against a 60-minute `timeout-minutes` and `make fuzz-budget-check` passes with the full 10 minutes of headroom. That is *exactly* the ceiling: the 21st target takes `ceil(21/5)` to 5 and `needed` to exactly 60. `scripts/fuzz-budget-check.sh` tests `needed > timeout`, so it would go **green at zero headroom**, describing the silently-truncated sweep its own error text warns about. Observed shard durations already vary from 4m25s to 5m42s.

**The sixth shard is carried by #358** (`5ccc6d4`), which adds three targets — the ones that consume the fifth shard — and the shard together: 23 targets / 6 shards / largest shard 4 / required 50 / timeout 60 / headroom 10, partition measured 4+4+4+4+4+3. The earlier note on this tracker that "whichever of #356/#357/#358 merges last must add a sixth shard" is superseded: #356 and #357 merged without needing it (they landed at 19 and 20), and #358 is the one that crosses the line. The gate cannot ask for it, because it is green either way.

sam-at-luther:
Closes the `analysis/` and `lint/` bullets on #318, and the `EnvMacroExpander.ExpandMacro` gap #348 identified. Targeted at v1.50.0, not the v1.49.1 patch.

## The blind spot #348 named

`EnvMacroExpander.ExpandMacro` had a blanket `recover()` whose recovery sets `result = nil` — which is *also* the ordinary "not a macro" answer. A panic in analysis-time macro expansion was invisible to any fuzzer, with no marker to key off. #348 said closing it needed a change in `analysis/`, not in the harness.

`ExpandMacro`'s body moved into an unexported `expand()`, and the wrapper holds one local:

```go
completed := false
defer func() {
    r := recover()
    if completed { return }
    e.recordAbort(r, form, pkg)   // atomic counter + Go stack + macro name
    result = nil
}()
result = e.expand(form, pkg)
completed = true
```

**Why it is non-forgeable.** `completed` is a Go local in `ExpandMacro`'s own frame, set by one statement reachable only via a normal return from `expand`. Nothing in lisp can touch a Go local, no expansion *result* can imitate it, and no embedder holds it because it does not exist outside the frame. The counter it feeds is unexported and only ever incremented, and `Reset()` deliberately does **not** clear it, so the code being watched cannot erase its own record.

This is the same family as `lisp.IsInternalPanic` and #348's `analysisCfg != nil` post-condition: a property only the non-panicking path can produce, not a flag. A `Panicked bool` field is the obvious alternative and is exactly what this avoids — any holder of the expander could set it either way.

Side catch: `runtime.Goexit` unwinding through `ExpandMacro` is counted too. `recover()` returns nil there, so a "did we recover?" flag would miss it.

Behaviour is unchanged — a recovered panic still returns nil and the analyser still falls back to opaque handling. The `NOT COVERED` row in `lsp/lsp_fuzz_test.go`'s recover-site table becomes `DETECTED`.

## Three new targets

| target | drives |
|---|---|
| `analysis.FuzzAnalyzeSource` | `Analyze`/`AnalyzeFile` over a fuzzer-built `Config`. The point is `DefForms`: `FormalsIndex`/`NameIndex` are integer indices into a fuzzer-chosen form, production derives them from workspace macro names, and no LSP-driven input can vary them. |
| `analysis.FuzzWorkspaceScan` | `PrescanWorkspace`, the six `ScanWorkspace*`, `ScanWorkspaceRefs`, `buildLoadTree`/`walkLoadFile`. Closes #318's "`ScanWorkspaceRefs` is never exercised". |
| `lint.FuzzLintSource` | `LintFile`/`LintFileWithAnalysis`/`LintFileWithContext`/`LintFiles` + `BuildAnalysisConfig`, with a fuzzer-drawn analyzer subset and an embedder-registered analyzer. |

**Containment for the workspace target is the load-bearing design decision.** `walkLoadFile` reads a file, finds `(load-file "…")`, joins to the containing dir and *recurses* — `../../../etc/passwd` escapes, and one escape in one file is enough. So every path, filename and dir is a harness constant; `load-file` is rewritten to `load-filx` (same length, so every `token.Location` is unchanged) in every fuzzer-derived file; and the load graph is harness-written from a fixed table. Asserted rather than commented: every collected path must resolve inside the temp root, and `TestWorkspaceFuzzContainsLoadFile` runs a hostile input end to end.

## Coverage against #318's baseline

Mean per-function, seed corpus alone, baseline measured on a stashed pristine tree:

| | transitive via `FuzzLSPSession` | dedicated |
|---|---|---|
| `analysis/` | **34.1%**, 67 of 125 funcs at 0% | **85.4%**, 3 of 130 at 0% |
| `lint/` | **38.7%**, 24 of 46 funcs at 0% | **90.1%**, 0 of 46 at 0% |

The three remaining zeroes are `recordAbort`, `safeHeadSymbol`, `LastExpansionPanic` — the panic-recording path, which by construction runs only when a defect is found. Unit tests cover them.

## Measuring first found two silent harness bugs

Same class #348 documented, and both invisible to a passing test:

- `ScanConfig.MaxFileBytes` was drawn as `intn(4)*16`, smaller than every file the harness writes. The scan collected **zero files**; everything past `collectLispFilesWithConfig` sat at 0% while every seed passed.
- `cfg.MacroExpander` was set on a coin flip and the guard's script cleared it, so `env.MacroCall` was never reached and the `ExpansionPanics` assertion was **vacuous**.

Both are now pinned by guard tests, along with requirements that the corpus reach a *successful* expansion, collect files/defs/preamble/DefFormSpecs, run the full analyzer set, and exercise all four `Pass.Report*` methods.

## Defects found, filed separately, not fixed here

- **#353** — `analysis.AnalyzeFile` silently drops `Config.MacroExpander`, so `ScanWorkspaceRefs` and the LSP's `updateFileRefs` compute cross-file references without macro expansion while per-document analysis uses it.
- **#354** — `astutil.SourceOf` nil-derefs on the nil parent `Walk` documents it will pass.

Both are judgement calls about intended behaviour, so they are in flight as their own change rather than bundled here.

## Campaign

~1.25M executions across two sweeps per target, `fuzz.sh` reporting 0 failures both times. **No crashers — verified on the filesystem**: `analysis/testdata` and `lint/testdata` do not exist after either sweep.

Weak evidence, and should be read as such: corpora were still growing steadily at the 10-minute mark (431 and 419 new interesting inputs still arriving at the end). Nowhere near saturated.

## Reviewer note: shard budget — resolved here, in `5ccc6d4`

*This section previously predicted 22 targets across the three in-flight #318 branches and asked whichever merged last to add a sixth shard. Reality has overtaken it; the numbers below are measured on this tree rather than projected.*

Two of the three siblings landed on `main` — #357 (`claude/fuzz-elpsutil`) and #356 (`claude/fuzz-core`, which also carried `lisp.FuzzCyclicValueWalks` forward from #391). That puts `main` at **20 targets — exactly the five-shard ceiling**, not the 22 the prediction assumed. The three targets added here take the sweep to **23**.

At 5 shards, `ceil(23/5) = 5` puts `needed` at 5 × 10m + 10m = **60 against a 60-minute timeout**. `fuzz-budget-check.sh` tests `needed > timeout`, so it goes **green at zero headroom** — the truncated sweep its own error text warns about, with observed shard durations already varying from 4m25s to 5m42s.

So this branch is the one that adds the **sixth shard**, in the same change as the targets that consume the fifth. Verified on this tree:

```
targets discovered      23
shards                  6
largest shard           4 target(s)
required                50 min
timeout-minutes         60 min
headroom                10 min
```

Checked rather than assumed:

- **Partition** — round-robin assignment measured per shard: `4 + 4 + 4 + 4 + 4 + 3 = 23`, every target assigned exactly once, nothing dropped.
- **Required checks** — the per-shard jobs are not required status checks by name; the aggregate `Required: fuzz` is. Renaming `Fuzz shard i/5` → `i/6` therefore strands nothing.
- `scripts/ci-gates-test.sh` — 127 passed, 0 failed with the new matrix. `scripts/confidentiality-guard.sh` — clean.

## Verification

Full `go test ./...` green; `golangci-lint run ./analysis/... ./lint/...` reports 0 issues. There is a pre-existing `gofmt` diff in `analysis/analyzer.go` on `main` which is left alone.

---

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -61,7 +61,7 @@ permissions:
 
 jobs:
   fuzz:
-    name: Fuzz shard ${{ matrix.shard }}/5
+    name: Fuzz shard ${{ matrix.shard }}/6
     runs-on: ubuntu-latest
     strategy:
       # Sharded so wall clock is set by the LARGEST shard rather than by the
@@ -81,19 +81,21 @@ jobs:
       # or a single failure costs the whole night's coverage.
       fail-fast: false
       matrix:
-        shard: [1, 2, 3, 4, 5]
+        shard: [1, 2, 3, 4, 5, 6]
     # DERIVED, not remembered. scripts/fuzz-budget-check.sh asserts
     #
     #     ceil(targets / shards) x FUZZTIME + overhead <= timeout-minutes
     #
     # on every PR, so this number cannot silently stop covering the sweep. At
-    # 20 targets over 5 shards that is 4 x 10m + 10m = 50; 60 leaves a shard's
-    # worth of headroom for a target being added before anyone reshards. That
-    # ceiling is now fully consumed: 20 targets is exactly ceil(20/5) = 4, so
-    # the NEXT target added is the one that runs out. ceil(21/5) = 5 puts
-    # `needed` at exactly 60, which the gate passes while describing a
-    # truncated sweep -- so the 21st target must arrive together with a sixth
-    # shard, not after it.
+    # 23 targets over 6 shards that is 4 x 10m + 10m = 50; 60 leaves a shard's
+    # worth of headroom for a target being added before anyone reshards.
+    #
+    # This branch is what consumed the previous ceiling. Five shards held to
+    # exactly 20 targets; the three targets added here take it to 23, and
+    # ceil(23/5) = 5 puts `needed` at exactly 60 -- which the gate PASSES while
+    # describing a truncated sweep. So the sixth shard arrives HERE, in the
+    # same change as the targets that need it, rather than being left for
+    # whoever next notices. ceil(23/6) = 4 restores the 10 minutes.
     #
     # THE HEADROOM IS THE POINT, NOT THE GREEN TICK. The gate's condition is
     # `needed > timeout`, so a sweep sized at exactly timeout-minutes PASSES --
@@ -170,7 +172,7 @@ jobs:
         run: ./scripts/fuzz-budget-check.sh
 
       - name: Fuzz
-        run: ./scripts/fuzz.sh --shard ${{ matrix.shard }}/5
+        run: ./scripts/fuzz.sh --shard ${{ matrix.shard }}/6
 
       # Only from main/schedule: a corpus grown from an unmerged branch would
       # otherwise be handed to every other PR.
--- a/.gitignore
+++ b/.gitignore
@@ -14,3 +14,8 @@ editors/vscode/bin/
 # rather than the whole committed corpus. Regenerated per run; never committed
 # (the crasher itself belongs in <package>/testdata/fuzz/<Target>/).
 fuzz-crashers/
+
+# scripts/ci-gates-test.sh synthesises a throwaway package here for its
+# empty-discovery control and deletes it again. Ignored so an interrupted run
+# cannot leave the tree looking dirty.
+internal/citest-emptyfuzz.*/
--- a/analysis/analysis_fuzz_guard_test.go
+++ b/analysis/analysis_fuzz_guard_test.go
@@ -1,0 +1,177 @@
+// Copyright © 2026 The ELPS authors
+
+package analysis
+
+import (
+	"os"
+	"path/filepath"
+	"regexp"
+	"strings"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Guards for the two analysis fuzz targets.
+//
+// These exist because of what PR #348 found by measuring rather than reading:
+// three separate harness bugs where every operation "ran", every test passed,
+// and whole functions sat at 0% because the harness was quietly starving them.
+// A fuzz target cannot report that it is covering less than it thinks. These
+// tests assert the specific reachability claims the harness comments make, so
+// that a future change which breaks one fails CI instead of silently reducing
+// what the nightly sweep explores.
+
+// TestAnalyzeFuzzSeedsExpandMacros is the sharpest of these guards.
+//
+// Setting cfg.MacroExpander looks sufficient for the analysis-time evaluation
+// path to be covered, and is not: the analyser consults the expander only for a
+// head symbol it does not already know, so a stdlib macro (known) and a macro
+// defined in the file under analysis (not in the env) both miss. Only a macro
+// that came from ANOTHER file is both present in the env and unknown to the
+// analyser. If macroSeeds ever stops producing that shape, env.MacroCall is
+// never reached, ExpandMacro returns nil every time for the boring reason, and
+// the target's ExpansionPanics assertion becomes vacuous while still passing.
+func TestAnalyzeFuzzSeedsExpandMacros(t *testing.T) {
+	// Exactly the cross product the seed corpus contains, so this guard is
+	// about the CORPUS rather than about a script invented here.
+	totalAsked, totalExpanded := 0, 0
+	for i, ms := range macroSeeds {
+		for j, sc := range analyzeScripts {
+			stats, err := runAnalyze([]byte(ms[0]), []byte(ms[1]), sc)
+			require.NoError(t, err, "macroSeeds[%d] with analyzeScripts[%d] failed", i, j)
+			totalAsked += stats.asked
+			totalExpanded += stats.expanded
+		}
+	}
+	assert.Positive(t, totalAsked,
+		"the analyser never consulted the macro expander at all")
+	assert.Positive(t, totalExpanded,
+		"no seed reached a SUCCESSFUL expansion, so env.MacroCall never ran and the"+
+			" analysis-time evaluation path this target claims to cover is covered by"+
+			" nothing. The shape that reaches it is a macro defined in the NEIGHBOURING"+
+			" file (so LoadWorkspaceMacros put it in the env) and called from the file"+
+			" under analysis (so the analyser does not already know it)")
+}
+
+// TestFuzzPreambleHeadsMatchScan pins the harness's copy of the preamble head
+// list to the switch it was copied from. If workspace.go starts collecting a
+// new form and this list does not, the harness replays a SUBSET of what
+// production replays and the difference is invisible.
+func TestFuzzPreambleHeadsMatchScan(t *testing.T) {
+	src, err := os.ReadFile("workspace.go")
+	require.NoError(t, err)
+
+	// The `case "in-package", "use-package", ...:` arm inside scanFileFull.
+	re := regexp.MustCompile(`(?s)switch astutil\.HeadSymbol\(expr\) \{\s*case ([^:]+):\s*preamble = append`)
+	m := re.FindSubmatch(src)
+	require.NotNil(t, m, "could not locate scanFileFull's preamble switch in workspace.go;"+
+		" this guard has gone stale and must be repointed, not deleted")
+
+	found := map[string]bool{}
+	for _, lit := range strings.Split(string(m[1]), ",") {
+		if s := strings.Trim(strings.TrimSpace(lit), `"`); s != "" {
+			found[s] = true
+		}
+	}
+	assert.Equal(t, found, preambleHeads,
+		"the fuzz harness's preambleHeads no longer matches scanFileFull's switch;"+
+			" the harness would replay a different preamble than production does")
+}
+
+// TestAnalyzeFuzzSeedsBuildScopes checks the seeds are not all degenerate: at
+// least one produces a Result with symbols and references. A corpus that parses
+// to nothing exercises the analyser's early returns and nothing else, which is
+// exactly how #348's didChange bug hid.
+func TestAnalyzeFuzzSeedsBuildScopes(t *testing.T) {
+	stats, err := runAnalyze([]byte(fuzzSeedUses), []byte(fuzzSeedDefs), []byte{4, 1, 2, 3, 1, 0, 1, 2})
+	require.NoError(t, err)
+	assert.Positive(t, stats.symbols, "the primary seed produced no symbols")
+	assert.Positive(t, stats.refs, "the primary seed produced no references")
+}
+
+// TestWorkspaceFuzzSeedsCollectFiles checks the workspace harness actually
+// materialises and collects a tree. A bug in writeWorkspace would leave the
+// scan looking at an empty directory: every call would return promptly, every
+// seed would pass, and the whole of workspace.go past collectLispFiles would be
+// unreached.
+func TestWorkspaceFuzzSeedsCollectFiles(t *testing.T) {
+	// Script with MaxFiles unconstrained and main.lisp written.
+	sc := []byte{1, 3, 1, 2, 0, 1, 1, 0, 1, 0, 1, 1, 1}
+	const defs = `(in-package 'wsdemo)
+(defun add (a b) (+ a b))
+(defmacro defthing (name args &rest body) (quasiquote (defun (unquote name) (unquote args) (unquote-splicing body))))
+(export 'add)`
+	stats, err := runWorkspaceScan(t.TempDir(), []byte(defs), []byte("(in-package 'user)\n(use-package 'wsdemo)\n(defun main () (add 1 2))"), sc)
+	require.NoError(t, err)
+	assert.Positive(t, stats.files, "the scan collected no files at all")
+	assert.Positive(t, stats.defs, "the scan found no definitions")
+	assert.Positive(t, stats.preamble,
+		"the scan collected no preamble forms, so LoadWorkspaceMacros had nothing to"+
+			" evaluate and the macro-loading path is unreached")
+	assert.Positive(t, stats.defForms,
+		"no DefFormSpec was derived from a def*-prefixed macro; extractDefFormSpecs"+
+			" and the whole def-like analysis path are unreached")
+}
+
+// TestWorkspaceFuzzContainsLoadFile is the containment claim as a test.
+//
+// walkLoadFile joins a load-file argument to the containing directory and
+// recurses, so `(load-file "../../../etc/passwd")` in a fuzzer-derived file
+// would make this target a file-read primitive pointed at the machine running
+// it. The harness strips the token; this proves the strip works, that it
+// preserves byte offsets, and that the only load edges that survive are the
+// harness's own.
+func TestWorkspaceFuzzContainsLoadFile(t *testing.T) {
+	hostile := []byte(`(in-package 'user)
+(load-file "../../../etc/passwd")
+(load-file "/etc/shadow")
+(defun f () 1)`)
+
+	safe := neutraliseLoadFile(hostile)
+	assert.Len(t, safe, len(hostile),
+		"the substitution changed the length, so every token.Location downstream"+
+			" of it now differs from what the original bytes would have produced")
+	assert.NotContains(t, string(safe), "load-file",
+		"a load edge survived neutralisation")
+
+	// End to end: run the harness on the hostile input and check every file the
+	// scan reports is inside the temp root.
+	root := t.TempDir()
+	stats, err := runWorkspaceScan(root, hostile, hostile, []byte{1, 3, 1, 2, 0, 1, 1, 0})
+	require.NoError(t, err)
+	assert.Positive(t, stats.files)
+
+	// And that the neutralised token is what is actually on disk.
+	for _, name := range []string{wsPlainFile, wsSubFile} {
+		b, readErr := os.ReadFile(filepath.Join(root, filepath.FromSlash(name))) //nolint:gosec // fixed name under t.TempDir()
+		require.NoError(t, readErr)
+		assert.NotContains(t, string(b), "load-file",
+			"%s went to disk with a live load edge in it", name)
+	}
+}
+
+// TestWorkspaceFuzzLoadTreeIsExercised checks the harness's own load edges
+// actually build a load tree, including the self-cycle that is the only thing
+// testing walkLoadFile's `visited` guard. Without this, main.lisp could quietly
+// stop being written and buildLoadTree would go to 0% with every seed passing.
+func TestWorkspaceFuzzLoadTreeIsExercised(t *testing.T) {
+	root := t.TempDir()
+	sc := &script{b: []byte{1, 3, 1, 2, 0, 1, 1, 0, 1, 0, 1, 1, 1}}
+	require.NoError(t, writeWorkspace(root, sc, []byte("(defun a () 1)"), []byte("(defun b () 2)")))
+
+	main, err := os.ReadFile(filepath.Join(root, wsMainFile)) //nolint:gosec // fixed name under t.TempDir()
+	require.NoError(t, err, "main.lisp was not written, so buildLoadTree never runs")
+	assert.Contains(t, string(main), "(load-file ",
+		"main.lisp carries no load edge, so walkLoadFile's recursion is unreached")
+
+	pkgMap, order := buildLoadTree(filepath.Join(root, wsMainFile))
+	assert.NotEmpty(t, order, "the load walk visited nothing")
+	for path := range pkgMap {
+		rel, relErr := filepath.Rel(root, path)
+		require.NoError(t, relErr)
+		assert.False(t, rel == ".." || strings.HasPrefix(rel, ".."+string(filepath.Separator)),
+			"the load tree escaped the workspace root: %q", path)
+	}
+}
--- a/analysis/analysis_fuzz_test.go
+++ b/analysis/analysis_fuzz_test.go
@@ -1,0 +1,722 @@
+// Copyright © 2026 The ELPS authors
+
+package analysis
+
+import (
+	"bytes"
+	"errors"
+	"fmt"
+	"testing"
+	"time"
+
+	"github.com/luthersystems/elps/astutil"
+	"github.com/luthersystems/elps/internal/fuzzseed"
+	"github.com/luthersystems/elps/internal/fuzzwatch"
+	"github.com/luthersystems/elps/lisp"
+	"github.com/luthersystems/elps/lisp/lisplib"
+	"github.com/luthersystems/elps/parser/rdparser"
+	"github.com/luthersystems/elps/parser/token"
+)
+
+// Fuzzing the ANALYSER on its own API, rather than through the language server.
+//
+// # Why a dedicated target, when FuzzLSPSession already reaches this package
+//
+// Issue #318 lists analysis/ as unfuzzed. FuzzLSPSession (PR #348) softened
+// that: driving didOpen runs parse -> analysis -> lint synchronously, and the
+// seed corpus alone measured 33.7% mean per-function coverage of this package.
+// That is a downgrade, not a close, for three separate reasons.
+//
+//  1. THE LSP IS NOT THE ONLY CALLER. luthersystems/substrate imports
+//     github.com/luthersystems/elps/analysis directly, as does lint/, the
+//     minifier and the MCP server. Whatever the language server happens not to
+//     construct is still constructible by those callers.
+//
+//  2. THE LSP CONSTRUCTS ONLY ONE SHAPE OF Config. Every field of
+//     analysis.Config is attacker-influenced in production -- ExtraGlobals and
+//     PackageSymbols come off disk, DefForms are DERIVED FROM MACRO NAMES in
+//     workspace source (extractDefFormSpecs), PackageImports and
+//     DefaultPackage from cross-file scans. The LSP builds them all from one
+//     workspace scan, so the fuzzer never varies them independently and never
+//     produces the combinations a caller with its own index would. DefForms is
+//     the sharpest of these: FormalsIndex and NameIndex are INTEGER INDICES
+//     into a fuzzer-chosen form's Cells, which is where an out-of-range slice
+//     lives, and no LSP-driven input can set them to an arbitrary value.
+//
+//  3. THE POSITION QUERIES ARE THIS PACKAGE'S, NOT THE SERVER'S.
+//     ScopeAtPosition and FindEnclosingFunction do their own line/column
+//     arithmetic on a scope tree, and are exported for embedders.
+//
+// # The assertion with teeth: ExpansionPanics
+//
+// The analyser evaluates code. When Config.MacroExpander is set -- which the
+// LSP, lint and the MCP server all do -- analysing a document CALLS
+// env.MacroCall on macros loaded from the workspace, so a macro body runs
+// while a file is merely being looked at. EnvMacroExpander.ExpandMacro wraps
+// that in a blanket recover() whose recovery sets result = nil.
+//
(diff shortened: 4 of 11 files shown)


sam-at-luther:
## The sixth shard is now in this branch — supersedes the "Reviewer note: shard budget" section above

That section was a prediction written when three #318 fuzz branches were in flight. The prediction was directionally right and numerically stale, so here is what actually happened, measured rather than projected.

Two of the three siblings landed — `claude/fuzz-elpsutil` (#357) and `claude/fuzz-core` (#356) — and #356 also carried `lisp.FuzzCyclicValueWalks` forward from #391. So `main` sits at **20 targets, exactly the five-shard ceiling**, not the 22 the note assumed. The three targets here take it to 23.

Measured on this tree at `64d117f` (main merged in, before the fix):

```
targets discovered      23      shards 5      largest shard 5
required                60 min  vs timeout 60 min      headroom 0 min
fuzz-budget-check: the sweep fits.
```

That last line is the failure mode, not the pass. The gate's condition is `needed > timeout`, so a sweep sized at *exactly* the timeout goes green while being the silently-truncated run the gate exists to prevent — and observed shard durations already vary from 4m25s to 5m42s.

`5ccc6d4` adds the sixth shard, verified on the same tree:

```
targets discovered      23      shards 6      largest shard 4
required                50 min  vs timeout 60 min      headroom 10 min
```

Checked rather than assumed:

- **Partition** — round-robin assignment measured per shard: `4 + 4 + 4 + 4 + 4 + 3 = 23`, every target assigned exactly once, nothing dropped.
- **Required checks** — the individual shard jobs are *not* required status checks by name; the aggregate `Required: fuzz` is. So renaming `Fuzz shard i/5` → `i/6` does not strand a required check. (#385 has already run a 6/6 matrix green, which is independent confirmation.)
- **`scripts/ci-gates-test.sh`** — 127 passed, 0 failed with the new matrix.
- **`scripts/confidentiality-guard.sh`** — clean.

The shard is added here rather than left for whoever merges next, because the gate cannot ask for it: it is green either way.

---

iamsamwood (commented):
Independent verification review at head `5ccc6d4` (isolated worktree; mutation-tested; all three new fuzz targets executed; CI-gate script rerun). Verdict from this pass: **approve** — with three corrections that were authored and verified locally; patches attached below since this session's credential cannot push to this branch.

## Verification summary

- **The behavioral fix is real and correctly scoped.** Restoring main's `analysis/expander.go` semantics (recover → nil, no `recordAbort`) while keeping the API makes `TestExpandMacroPanicIsCounted` fail ("the recovered panic was not counted; it is invisible again") and `TestLSPFuzzDetectsSwallowedExpansionPanic` fail ("the detector is not wired to the object the analyser uses") — exactly the right reasons. The false-positive half (`TestExpandMacroNoPanicNotCounted`, 9 lisp-reachable nil-returning cases) passes under both variants, as it should.
- **No live-server resilience regression**: `ExpansionPanics`/`LastExpansionPanic`/`PanicReporter` have zero non-test callers — a panic swallowed in a live LSP is still swallowed; only harnesses convert the count into failure.
- **All three fuzz targets are wired and execute** (`FuzzAnalyzeSource` 268-seed baseline, `FuzzWorkspaceScan`, `FuzzLintSource` all run real inputs). The `load-filx` containment proof, the lock ordering around `recordAbort`, and the sixth-shard arithmetic (23 targets / 6 shards / 10 min headroom) all check out. The CI self-test rework is **required**, not creep — main's empty-discovery control breaks the moment `FuzzLintSource` lands in `lint/`.
- Merge onto current main: already up to date (branch contains main via `64d117f`). CI 20/20 green.

## Corrections (authored, verified, attached as patches)

1. **CLAUDE.md is falsified by this PR and must be updated in it** — line 74's "`lsp.FuzzLSPSession` … is also the only target that reaches `analysis/` and `lint/`" and line 81's "`EnvMacroExpander.ExpandMacro`'s is not covered and the header says so" both stop being true here. Every prior fuzz PR that changed these facts (#311, #323, #343, #348) updated CLAUDE.md in the same change. Patch 1 rewrites both sentences to record the new state.
2. **`lintPreamble`'s head list had no drift guard** — it's a third copy of the preamble-head list; `analysis/` pins its copy to `scanFileFull`'s switch via `TestFuzzPreambleHeadsMatchScan`, but the lint copy could silently drift and replay a subset of what production replays. Patch 2 hoists the map and adds the same regexp-over-source guard (verified to fail when a head is removed).
3. **Three harness error paths dereferenced `LastExpansionPanic()` without a nil-guard** — `recordAbort` counts before it stores, so `ExpansionPanics() > 0` with a nil record is observable; the lsp-side `checkExpansionPanics` guards it, these didn't, and a nil-deref there replaces the swallowed-panic report with a harness panic — the exact failure mode this machinery exists to avoid. Patch 3 adds the guards.

All three verified on top of `5ccc6d4`: `go test -count=1 ./analysis/... ./lint/... ./lsp/...` green, `FuzzLintSource` still executes, `gofmt`/`go vet` clean, and the new drift guard red-tested by mutation.

Apply with `git am` on the branch head:

<details>
<summary>Patch 1/3 — docs: update CLAUDE.md fuzzing section for the new analysis/lint targets</summary>

```patch
diff --git a/CLAUDE.md b/CLAUDE.md
index d1045a3..71dbf08 100644
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -71,14 +71,14 @@ rc = env.InPackage(lisp.String(lisp.DefaultUserPackage))
 Tests exist in three forms:
 1. **Go unit tests** — Standard `_test.go` files using `testify/assert`.
 2. **Lisp test files** — `.lisp` files executed via `elpstest.Runner`, which loads them and runs as Go subtests. The `libtesting` stdlib package provides `test`, `test-let`, `assert=`, `assert-equal`, `assert-nil`, etc.
-3. **Fuzz targets** — native `go test -fuzz` targets in `fuzz_test.go` files, covering the parser (strict, fault-tolerant and format-preserving), the lexer and byte scanner, the formatter and minifier round-trips, the JSON decoder, the **evaluator** (`lisp.FuzzEval`), the **debugger's** eval path (`lisp/x/debugger.FuzzDebugEval` — attaching a debugger disables tail-call optimisation globally and stamps `MacroExpansionInfo` onto every macro-expanded node, so it is a materially different path from `FuzzEval`'s), and the **language server** as a whole session (`lsp.FuzzLSPSession` — a fuzzer-chosen sequence of LSP requests against fuzzer-chosen documents and *cursor positions*; it is also the only target that reaches `analysis/` and `lint/`). Seeds come from `internal/fuzzseed` (the repository's real `.lisp` sources plus hand-written adversarial input); each package's `testdata/fuzz/<Target>/` holds the crashers found so far, which plain `go test` replays as regression cases. Run with `make fuzz`; CI runs `.github/workflows/fuzz.yml` (30s/target on PRs, 10m/target nightly).
+3. **Fuzz targets** — native `go test -fuzz` targets in `fuzz_test.go` files, covering the parser (strict, fault-tolerant and format-preserving), the lexer and byte scanner, the formatter and minifier round-trips, the JSON decoder, the **evaluator** (`lisp.FuzzEval`), the **debugger's** eval path (`lisp/x/debugger.FuzzDebugEval` — attaching a debugger disables tail-call optimisation globally and stamps `MacroExpansionInfo` onto every macro-expanded node, so it is a materially different path from `FuzzEval`'s), the **language server** as a whole session (`lsp.FuzzLSPSession` — a fuzzer-chosen sequence of LSP requests against fuzzer-chosen documents and *cursor positions*; still the only whole-session LSP target), and **semantic analysis and lint** directly (`analysis.FuzzAnalyzeSource`, `analysis.FuzzWorkspaceScan`, and `lint.FuzzLintSource` — fuzzer-chosen source, workspace trees, and analyzer sets, so `analysis/` and `lint/` no longer rely on the LSP session for their fuzz coverage). Seeds come from `internal/fuzzseed` (the repository's real `.lisp` sources plus hand-written adversarial input); each package's `testdata/fuzz/<Target>/` holds the crashers found so far, which plain `go test` replays as regression cases. Run with `make fuzz`; CI runs `.github/workflows/fuzz.yml` (30s/target on PRs, 10m/target nightly).
 
    **Evaluator fuzzing is different** and the differences are load-bearing (see the header comment on `lisp/eval_fuzz_test.go`):
    - Evaluation is Turing-complete and `go test -fuzz` has no per-input deadline, so every evaluation runs under an explicit budget (`WithMaxSteps`, `WithMaxTailIterations`, `WithMaximumPhysicalStackHeight`, `WithMaxAlloc`, plus a context deadline) *and* a 30s watchdog goroutine, so "it terminates" is an assertion rather than an assumption.
    - `env.eval` recovers every Go panic into an ordinary-looking `*LVal`, so "the process survived" proves nothing. The assertion with teeth is **`lisp.IsInternalPanic(result) == false`** — the non-forgeable marker keyed off the recovered Go-stack snapshot.
    - Eval seeds are *executed*, so `internal/fuzzseed/evalseed.go` is hand-written and deliberately does **not** seed from `LispSources()`: a large `.lisp` test suite as a seed throttles every generation descended from it. Seeds are split into `EvalRunaway` (must error) and `EvalTerminating` (must complete without error) — an infinite ELPS loop the budget stops is CORRECT; a bounded program that trips a budget is a defect, and both directions are asserted.
 
-   **LSP fuzzing has the same hazard in a different shape** (see the header on `lsp/lsp_fuzz_test.go`): the server installs four blanket `recover()`s, so "it did not crash" is again worth nothing until they are accounted for. Two are neutralised by running the timer closures inline instead of letting them fire asynchronously; `buildWorkspaceIndex`'s is *detected* (it has no early return, so a nil `analysisCfg` afterwards means it panicked); `EnvMacroExpander.ExpandMacro`'s is not covered and the header says so. Everything else runs with no recover between the handler and the fuzz function, so a plain Go panic is the assertion.
+   **LSP fuzzing has the same hazard in a different shape** (see the header on `lsp/lsp_fuzz_test.go`): the server installs four blanket `recover()`s, so "it did not crash" is again worth nothing until they are accounted for. Two are neutralised by running the timer closures inline instead of letting them fire asynchronously; `buildWorkspaceIndex`'s is *detected* (it has no early return, so a nil `analysisCfg` afterwards means it panicked); `EnvMacroExpander.ExpandMacro`'s is now also *detected* — the expander counts every call that aborts before its normal return (a frame-local flag in `analysis/expander.go` that only a completed expansion sets, feeding a monotonic counter exposed via `analysis.PanicReporter`), and the harness's `checkExpansionPanics` fails the run on any swallowed expansion panic. Everything else runs with no recover between the handler and the fuzz function, so a plain Go panic is the assertion.
 
 Go test suites typically use `elpstest.TestSuite` with `TestSequence` entries that define `{expression, expected-result, expected-output}` triples.
```
</details>

<details>
<summary>Patch 2/3 — lint: pin lintPreamble's head list to scanFileFull's switch</summary>

```patch
diff --git a/lint/lint_fuzz_guard_test.go b/lint/lint_fuzz_guard_test.go
index 939bd2e..96d54d6 100644
--- a/lint/lint_fuzz_guard_test.go
+++ b/lint/lint_fuzz_guard_test.go
@@ -4,6 +4,9 @@ package lint
 
 import (
 	"bytes"
+	"os"
+	"path/filepath"
+	"regexp"
 	"strings"
 	"testing"
 
@@ -120,6 +123,32 @@ func TestLintFuzzDeterminismCheckIsLive(t *testing.T) {
 	assert.Equal(t, first, second)
 }
 
+// TestLintFuzzPreambleHeadsMatchScan pins the harness's copy of the preamble
+// head list to the switch it was copied from, the same way analysis's
+// TestFuzzPreambleHeadsMatchScan pins that package's copy. If scanFileFull
+// starts collecting a new form and this list does not, lintPreamble replays a
+// SUBSET of what production replays and the difference is invisible.
+func TestLintFuzzPreambleHeadsMatchScan(t *testing.T) {
+	src, err := os.ReadFile(filepath.Join("..", "analysis", "workspace.go"))
+	require.NoError(t, err)
+
+	// The `case "in-package", "use-package", ...:` arm inside scanFileFull.
+	re := regexp.MustCompile(`(?s)switch astutil\.HeadSymbol\(expr\) \{\s*case ([^:]+):\s*preamble = append`)
+	m := re.FindSubmatch(src)
+	require.NotNil(t, m, "could not locate scanFileFull's preamble switch in analysis/workspace.go;"+
+		" this guard has gone stale and must be repointed, not deleted")
+
+	found := map[string]bool{}
+	for _, lit := range strings.Split(string(m[1]), ",") {
+		if s := strings.Trim(strings.TrimSpace(lit), `"`); s != "" {
+			found[s] = true
+		}
+	}
+	assert.Equal(t, found, lintPreambleHeads,
+		"the lint fuzz harness's lintPreambleHeads no longer matches scanFileFull's"+
+			" switch; lintPreamble would replay a different preamble than production does")
+}
+
 func parseForGuard(t *testing.T, src string) []*lisp.LVal {
 	t.Helper()
 	return rdparser.New(token.NewScanner(lintFileA, bytes.NewReader([]byte(src)))).
diff --git a/lint/lint_fuzz_test.go b/lint/lint_fuzz_test.go
index c7dca14..17afda6 100644
--- a/lint/lint_fuzz_test.go
+++ b/lint/lint_fuzz_test.go
@@ -508,19 +508,24 @@ func walkSurface(src []byte) {
 	_ = UserDefined(res.Exprs)
 }
 
+// lintPreambleHeads are the top-level forms scanFileFull collects into
+// Prescan.Preamble and LoadWorkspaceMacros then evaluates. Kept in step with
+// the switch in analysis/workspace.go; TestLintFuzzPreambleHeadsMatchScan
+// asserts it.
+var lintPreambleHeads = map[string]bool{
+	"in-package": true, "use-package": true, "export": true,
+	"defmacro": true, "defun": true, "set": true,
+}
+
 // lintPreamble extracts the forms a workspace scan would replay into the env.
 func lintPreamble(src []byte) []*lisp.LVal {
 	res := rdparser.New(token.NewScanner(lintFileB, bytes.NewReader(src))).ParseProgramFaultTolerant()
-	heads := map[string]bool{
-		"in-package": true, "use-package": true, "export": true,
-		"defmacro": true, "defun": true, "set": true,
-	}
 	var out []*lisp.LVal
 	for _, expr := range res.Exprs {
 		if expr == nil || expr.Type != lisp.LSExpr || expr.Quoted || len(expr.Cells) == 0 {
 			continue
 		}
-		if heads[HeadSymbol(expr)] {
+		if lintPreambleHeads[HeadSymbol(expr)] {
 			out = append(out, expr)
 		}
 	}
```
</details>

<details>
<summary>Patch 3/3 — analysis, lint: nil-guard LastExpansionPanic in harness error paths</summary>

```patch
diff --git a/analysis/analysis_fuzz_test.go b/analysis/analysis_fuzz_test.go
index d93e40d..af1d83d 100644
--- a/analysis/analysis_fuzz_test.go
+++ b/analysis/analysis_fuzz_test.go
@@ -422,6 +422,9 @@ func runAnalyze(src, wsSrc, scriptBytes []byte) (analyzeResult, error) {
 	var reporter PanicReporter = exp
 	if n := reporter.ExpansionPanics(); n > 0 {
 		rec := exp.inner.LastExpansionPanic()
+		if rec == nil { // recordAbort counts before it stores, so nil is observable
+			rec = &ExpansionPanic{}
+		}
 		return stats, fmt.Errorf(
 			"EnvMacroExpander swallowed %d panic(s) during analysis-time macro expansion."+
 				" ExpandMacro's recover() returns nil, which is also the ordinary"+
diff --git a/analysis/workspace_fuzz_test.go b/analysis/workspace_fuzz_test.go
index 0140064..f7e349c 100644
--- a/analysis/workspace_fuzz_test.go
+++ b/analysis/workspace_fuzz_test.go
@@ -291,6 +291,9 @@ func runWorkspaceScan(root string, a, b, scriptBytes []byte) (scanStats, error)
 
 	if n := exp.ExpansionPanics(); n > 0 {
 		rec := exp.inner.LastExpansionPanic()
+		if rec == nil { // recordAbort counts before it stores, so nil is observable
+			rec = &ExpansionPanic{}
+		}
 		return stats, fmt.Errorf(
 			"EnvMacroExpander swallowed %d panic(s) during the workspace scan"+
 				"\n  macro:   %s\n  package: %s\n  value:   %v\n  stack:\n%s",
diff --git a/lint/lint_fuzz_test.go b/lint/lint_fuzz_test.go
index 17afda6..c7aa925 100644
--- a/lint/lint_fuzz_test.go
+++ b/lint/lint_fuzz_test.go
@@ -409,6 +409,9 @@ func runLint(root string, src, wsSrc, scriptBytes []byte) (lintStats, error) {
 	// have shown this.
 	if n := exp.ExpansionPanics(); n > 0 {
 		rec := exp.LastExpansionPanic()
+		if rec == nil { // recordAbort counts before it stores, so nil is observable
+			rec = &analysis.ExpansionPanic{}
+		}
 		return stats, fmt.Errorf(
 			"EnvMacroExpander swallowed %d panic(s) while lint ran semantic analysis"+
 				"\n  macro:   %s\n  package: %s\n  value:   %v\n  stack:\n%s",
```
</details>

---

PR #359, opened 2026-08-10
Honour Config.MacroExpander in AnalyzeFile, and stop astutil nil-dereferencing

### LINKED ISSUE #353: analysis.AnalyzeFile silently drops Config.MacroExpander

sam-at-luther:
Fixes #353 and #354, in two independently revertable commits.

## #354 — `astutil` accessors nil-deref on the nil parent `Walk` documents

`Walk` documents "parent is nil for top-level expressions"; the accessors exported alongside it dereferenced unconditionally. Each SIGSEGV'd before the fix, verified individually.

`SourceOf` was not alone — `HeadSymbol` and `ArgCount` have the same shape and the same natural pairing with `Walk`. All three now return their zero value (`nil`, `""`, `0`) and document it. The remaining helpers (`Walk`, `WalkSExprs`, `UserDefined`, `CollectFormals`, `PackageNameArg`) already guarded.

## #353 — `AnalyzeFile` dropped `Config.MacroExpander`

**Option (a): carry the field through.** The evidence that the omission was never deliberate:

- The field-by-field rebuild was written 2026-02-28 (`c0c6a20`). `MacroExpander` was added to `Config` 2026-04-01 (`e5e6417`), and that commit never touched `analysis/workspace.go`. There is no revision where the field existed and was consciously skipped.
- The twin rebuild in `lsp/server.go:getAnalysisConfig` **does** carry it — same pattern, same author.
- This exact site had already dropped fields once before: `3c6a860` added `DefForms` and `PackageImports` back after they went missing, leaving a regression test behind.

**The cost premise for "skip expansion deliberately" does not hold.** Measured on a synthetic 200-file / 2400-definition workspace running full `ScanWorkspaceRefs`: 350ms with the expander against 395ms without. Expansion is gated on `isMacro || sym == nil`, `EnvMacroExpander` caches negative lookups per (pkg, name), and an expanded form is analysed *instead of* the opaque walk rather than in addition.

The rebuild is now a whole-struct copy (`ConfigForFile`), applied at all three sites that specialise a workspace `Config` for one file — `analysis/workspace.go`, `lsp/server.go`, `mcpserver/service.go`. The latter two were complete but carried the same latent risk. A reflection-based test asserts every `Config` field except `Filename` survives the copy, so the next field added cannot go missing here again.

## The part reviewers should look at: wiring it through naively would have corrupted cross-file rename

Macro expansion splices nodes from the **macro's definition site** into the calling file's AST, and those nodes carry the *other* file's `Source`. `ExtractFileRefs` stamped them with the calling file, producing a reference with `File=b.lisp` and `Source=a.lisp:3:40`.

`lsp/rename.go:111` takes the edit URI from `File` and the edit range from `Source`. So cross-file rename would have written edits into `b.lisp` at coordinates belonging to `a.lisp`, and `lsp/references.go:54` would have reported a bogus location. The per-document path was safe, because it uses `ref.Source.File` for the URI.

`ExtractFileRefs` now skips references whose text lives in another file; they are captured correctly when the defining file is itself scanned. `TestScanWorkspaceRefs_MacroExpansionKeepsRefsInOwnFile` pins it.

I verified the test actually detects the regression rather than merely passing: removing the guard fails it with `reference recorded in .../b.lisp but sourced in .../a.lisp`.

## Other blast radius

- A **third** `ScanWorkspaceRefs` caller exists that #353 did not list: `lint/lint.go:431`, which also sets `MacroExpander`. CLI lint's cross-file `unused-function` detection now sees macro-reachable references.
- `mcpserver/service.go:692` calls `ScanWorkspaceRefs` *before* `LoadWorkspaceMacros` at :700, so its expander is typically nil at scan time. Unaffected here, but worth a separate look.
- With expansion on, a macro call site records one reference instead of two — the opaque path recorded the head as a macro reference and again as a walked child. That removes duplicate rename edits at the same range.
- `ScanWorkspaceRefs` fans out over `NumCPU` workers against the single mutex-serialised `EnvMacroExpander`. Documented thread-safe; the timing above is measured with that contention.

## Verification

`go build ./...` clean, full `go test ./...` green, `gofmt -l` clean on touched files, `golangci-lint run` over `analysis/ lint/ astutil/ lsp/ mcpserver/` reports 0 issues. The pre-existing `analysis/analyzer.go` gofmt diff on main is untouched.

---

### DIFF
--- a/analysis/analysis_fuzz_test.go
+++ b/analysis/analysis_fuzz_test.go
@@ -6,6 +6,7 @@ import (
 	"bytes"
 	"errors"
 	"fmt"
+	"sync/atomic"
 	"testing"
 	"time"
 
@@ -133,17 +134,22 @@ func (discardWriter) Write(p []byte) (int, error) { return len(p), nil }
 //
 // It also forwards ExpansionPanics, so the target's assertion goes through the
 // exported PanicReporter interface exactly as an embedder's would.
+// The counters are atomic because ScanWorkspaceRefs analyses files
+// concurrently and shares one Config -- and therefore one expander -- across
+// those goroutines. That only became reachable when AnalyzeFile started
+// honouring Config.MacroExpander; before that this expander was never called
+// from the parallel scan, so plain ints raced with nobody.
 type countingExpander struct {
 	inner    *EnvMacroExpander
-	expanded int
-	asked    int
+	expanded atomic.Int64
+	asked    atomic.Int64
 }
 
 func (c *countingExpander) ExpandMacro(form *lisp.LVal, pkg string) *lisp.LVal {
-	c.asked++
+	c.asked.Add(1)
 	v := c.inner.ExpandMacro(form, pkg)
 	if v != nil {
-		c.expanded++
+		c.expanded.Add(1)
 	}
 	return v
 }
@@ -412,7 +418,7 @@ func runAnalyze(src, wsSrc, scriptBytes []byte) (analyzeResult, error) {
 	// config copy and the strict parser.
 	_ = AnalyzeFile(src, fuzzAnalysisFile, cfg)
 
-	stats.asked, stats.expanded = exp.asked, exp.expanded
+	stats.asked, stats.expanded = int(exp.asked.Load()), int(exp.expanded.Load())
 	stats.symbols, stats.refs = len(result.Symbols), len(result.References)
 
 	probeResult(sc, result, src)
--- a/analysis/samesourcefile_test.go
+++ b/analysis/samesourcefile_test.go
@@ -1,0 +1,70 @@
+package analysis
+
+import (
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// sameSourceFile decides whether a reference's text lives in the file being
+// indexed. ExtractFileRefs pairs File (the URI rename edits) with Source (the
+// range rename edits), so a wrong "true" here aims an edit at the wrong file.
+// These cases pin each branch against the two call sites that exist.
+func TestSameSourceFile(t *testing.T) {
+	t.Run("identical is the LSP call site's fast path", func(t *testing.T) {
+		// lsp/server.go passes one filePath to both AnalyzeFile (which sets
+		// Source.File) and ExtractFileRefs, so ordinary nodes compare equal
+		// without ever reaching filepath.Abs.
+		assert.True(t, sameSourceFile("/ws/a.lisp", "/ws/a.lisp"))
+		assert.True(t, sameSourceFile("a.lisp", "a.lisp"))
+	})
+
+	t.Run("Abs carries the workspace-scan call site", func(t *testing.T) {
+		// workspace.go analyses with a possibly-relative path but extracts
+		// against its absolute form. Without normalisation every reference
+		// from a relative path would be dropped and rename would silently
+		// miss those files -- so this is load-bearing, not cosmetic.
+		abs, err := filepath.Abs("a.lisp")
+		require.NoError(t, err)
+		assert.True(t, sameSourceFile("a.lisp", abs))
+		assert.True(t, sameSourceFile(abs, "a.lisp"))
+		assert.True(t, sameSourceFile("./a.lisp", "a.lisp"))
+		assert.True(t, sameSourceFile("b/../a.lisp", "a.lisp"))
+	})
+
+	t.Run("distinct files are rejected", func(t *testing.T) {
+		assert.False(t, sameSourceFile("a.lisp", "b.lisp"))
+		assert.False(t, sameSourceFile("/ws/a.lisp", "/other/a.lisp"))
+		// The macro-expansion case this guard was added for: a node spliced
+		// in from the macro's defining file carries that file's location.
+		assert.False(t, sameSourceFile("macros.lisp", "consumer.lisp"))
+	})
+
+	t.Run("empty source is deliberately permissive", func(t *testing.T) {
+		// FAIL-OPEN, and the only branch that can produce a false positive.
+		// Nodes without a recorded file are attributed to the file being
+		// indexed; failing closed instead would drop them and break rename
+		// for ordinary code. The cost is that if macro expansion ever yields
+		// a node with an EMPTY Source.File rather than the defining file's
+		// path, this guard cannot see it and rename could aim at the wrong
+		// file. Anything that starts populating Source.File differently
+		// should revisit this case first.
+		assert.True(t, sameSourceFile("", "consumer.lisp"))
+		assert.True(t, sameSourceFile("", "/anywhere/at/all.lisp"))
+	})
+
+	t.Run("known limits, recorded rather than asserted as correct", func(t *testing.T) {
+		// Both are false negatives -- they drop a real reference, so rename
+		// under-applies rather than corrupting an unrelated file. Neither is
+		// reachable from the current call sites, because both paths derive
+		// from the same string. They would become reachable if a caller ever
+		// compared an editor-supplied path against an independently resolved
+		// one.
+		assert.False(t, sameSourceFile("/tmp/a.lisp", "/private/tmp/a.lisp"),
+			"symlinked paths to one file are not resolved")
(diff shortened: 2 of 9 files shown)


iamsamwood (commented):
Independent verification review at head `62672f5`, run in an isolated worktree and trial-merged against current main `663c0f3`. Verdict from this pass: **approve** — both fixes are real, both red-tested, and the riskiest part of the change is the part that's best defended.

## Verification

**#354 (astutil nil-deref).** Restoring main's `astutil/walk.go` and keeping the PR's tests: `TestSourceOf_Nil` fails with `SIGSEGV: segmentation violation`, and `TestHeadSymbol_Nil` / `TestArgCount_Nil` fail identically — so guarding the sibling helpers was fixing real crashes, not decoration. `TestWalk_NilParentAccessors` (the issue's exact `Walk`-pairing reproducer) fails via `assert.NotPanics`. Restored: green. I also confirmed the PR's claim that `Walk`, `WalkSExprs`, `UserDefined`, `CollectFormals` and `PackageNameArg` were already nil-safe.

**#353 (AnalyzeFile drops MacroExpander).** Reverting just the `AnalyzeFile` body to main's field-by-field rebuild: `TestAnalyzeFile_ForwardsMacroExpander` fails with `expected: 1, actual: 0` — the expander consulted once by `Analyze` and zero times by `AnalyzeFile`, exactly the issue's symptom.

**The `ExtractFileRefs` guard is the load-bearing part, and it's red-tested too.** With the #353 fix in place but `sameSourceFile` disabled: `TestScanWorkspaceRefs_MacroExpansionKeepsRefsInOwnFile` fails with a reference recorded under `b.lisp` whose range belongs to `a.lisp:3:40` — precisely the coordinates `lsp/rename.go` would then apply to the wrong file. Forwarding the expander naively would have corrupted cross-file renames; the guard is not optional and the PR is right to treat it as part of the fix.

One nuance worth knowing (not a defect): that invariant test **passes** with the #353 fix reverted, because without expansion no foreign nodes are spliced and the invariant holds vacuously. `TestAnalyzeFile_ForwardsMacroExpander` carries the #353 red signal; the invariant test's red condition is the guard. The division of labour is correct.

Full `go test ./...` passes at head; trial merge onto `663c0f3` auto-merges clean (only overlap is appended tests in `analysis/workspace_test.go`) and passes; CI 19/19 green. The nil-`MacroExpander` default is unchanged for existing callers — the expansion gate at `analysis/analyzer.go:1297` requires non-nil — so the behaviour change is confined to callers that already set the field and were being silently ignored.

## Findings

1. **`mcpserver/service.go:682-692`** — `buildWorkspaceState` calls `ScanWorkspaceRefs` before `state.cfg.MacroExpander` is set, so mcpserver's workspace index is still built without expansion while its per-document analysis now expands: the #353 disagreement surviving in one caller. I agree it doesn't belong in this PR (the fix needs `LoadWorkspaceMacros` hoisted above the scan), so I've filed it as #403 rather than asking you to grow the diff.

2. **Interaction with #396 — worth knowing before merge.** This PR widens where `EnvMacroExpander.ExpandMacro` runs: previously per-document analysis only, now also inside `ScanWorkspaceRefs` (fanned out over `NumCPU` workers, for lsp `buildWorkspaceIndex`, `lint/lint.go:431`, and eventually mcpserver) and `updateFileRefs`. #396 reports that `ExpandMacro` aliases the form it expands (`args := lisp.SExpr(form.Cells[1:])`, `analysis/expander.go:96`), so a destructive macro writes through into the analyzed AST. Workspace scans parse fresh ASTs per call so the corruption isn't persistent there — but a mid-scan mutation can now put corrupted ranges into the long-lived `workspaceRefs` index, which it couldn't before this PR. Nothing to change here; it raises #396's priority, and #396 currently has a fix on `claude/elpsvet-ci` with no open PR.

3. **`sameSourceFile` (`analysis/workspace.go:675-685`)** — nit. `filepath.Abs` normalizes relative-vs-absolute but not symlinked spellings. If the prescan walk and an editor URI reach the same file through different spellings, a same-file macro-spliced ref is misclassified as foreign and dropped — a silent missing reference, not corruption. `filepath.EvalSymlinks` (errors tolerated) would close it. Fine to leave.

4. **PR body, "Option (a)" paragraph** — nit: "350ms with the expander against 395ms without" reads backwards. If the numbers are swapped, worth fixing so the cost claim reads correctly; if they're genuinely in that order, one clause on why would spare the next reader the double-take.

## Correction (verified, attached)

Finding: `analysis/workspace_test.go:1937`'s reflection test calls `sv.Field(i).Interface()`, which panics on unexported fields. `Config` has none today, so this is latent — but if one is ever added the test dies with an opaque reflect panic instead of a readable failure. Patch below adds the guard, in the test's existing assertion style.

Verified on top of `62672f5` merged with `663c0f3`: `go test ./analysis/... ./astutil/... ./lint/... ./lsp/... ./mcpserver/...` all pass, `gofmt`/`go vet` clean.

Apply with `git merge origin/main` (clean, verified) then `git am`:

<details>
<summary>Patch — analysis: fail readably if Config ever gains an unexported field</summary>

```patch
diff --git a/analysis/workspace_test.go b/analysis/workspace_test.go
index bd854f7..3d774be 100644
--- a/analysis/workspace_test.go
+++ b/analysis/workspace_test.go
@@ -1934,6 +1934,10 @@ func TestConfigForFile_CarriesEveryField(t *testing.T) {
 		if name == "Filename" {
 			continue
 		}
+		// Interface() panics on unexported fields; fail readably instead if
+		// Config ever gains one.
+		require.True(t, sv.Type().Field(i).IsExported(),
+			"unexported Config field %s cannot be checked by this test — update it", name)
 		assert.Equal(t, sv.Field(i).Interface(), gv.Field(i).Interface(),
 			"ConfigForFile dropped Config.%s", name)
 	}
```
</details>

Scope note: the `ConfigForFile` consolidation into `lsp/` and `mcpserver/` is beyond the two stated fixes but verified behaviour-preserving (both of main's rebuilds already carried all 8 fields plus `Filename`, pinned by the new reflection test), and it converts two latent copies of the bug class into one audited implementation. In scope, in my view.

---

Issue #99, opened 2026-02-12
libhelp.LoadPackage does not restore previous package


sam-at-luther:
## Bug

`libhelp.LoadPackage(env)` calls `env.InPackage(name)` to switch to the `help` package (line 123 in libhelp.go) but never restores the previous package. After calling `LoadPackage`, the environment is left in the `help` package instead of the caller's package (typically `user`).

This means any code evaluated after `LoadPackage` runs in the `help` package, which doesn't `use` the `lang` package. Core builtins like `in-package`, `defun`, `set`, etc. become unresolvable.

## Reproduction

```go
env, _ := shirocore.NewEnv(ctx, libcc.AllowMockLibraries)
// env.Runtime.Package.Name == "user" ✓

libhelp.LoadPackage(env)
// env.Runtime.Package.Name == "help" ✗ — should still be "user"

// Now any phylum load fails:
// main.lisp:1:2: cc-context:load-chaincode: unbound symbol: in-package
```

## Expected behavior

`LoadPackage` should save and restore the current package, like `env.Load` does:

```go
func LoadPackage(env *lisp.LEnv) *lisp.LVal {
    prevPkg := env.Runtime.Package.Name
    defer env.InPackage(lisp.Symbol(prevPkg))
    // ... rest of function
}
```

## Impact

Any embedder that calls `libhelp.LoadPackage(env)` before loading user code will see `unbound symbol` errors for all `lang` builtins. This affects the `shirotester repl` and `shirotester doc` commands in substrate.

## Workaround

Save and restore the package manually:
```go
prevPkg := env.Runtime.Package.Name
libhelp.LoadPackage(env)
env.InPackage(lisp.Symbol(prevPkg))
```

PR #360, opened 2026-08-10
Restore the caller's package across elpsutil loads

### LINKED ISSUE #352: elpsutil.PackageLoader does not restore the caller's package after PackageInit
elps already treats "a package load restores the previous package" as an invariant for its own stdlib. #99 fixed the `LoadPackage` functions for violating it, and `TestLoadLibrary_EachPackageRestoresPrevious` pins it.

That guarantee does not extend through `elpsutil` to embedders, which is the one path elps ships specifically *for* embedders.

## The behaviour

`PackageLoader` does not re-establish its package after `PackageInit` returns. An embedder whose `PackageInit` switches packages — directly, or indirectly by calling a nested `elpsutil.Load` — gets every subsequent builtin, special op and macro registered in the *other* package, and exported from there. Silently. No error.

This is not hypothetical, and it is not a corner case: `elpsutil.Loader`'s own doc comment tells embedders to do the thing that triggers it.

> A Loader should first define and switch into the package(s) it loads.

An embedder following that convention inside `PackageInit` gets misplaced symbols.

## It has already been hit downstream

`luthersystems/substrate` discovered this and compensates by hand, in `internal/substrate/shiro/libcc/libcc.go`:

```go
lerr := elpsutil.Load(env, elpsutil.LibraryLoader(subs...))
if lerr.Type == lisp.LError {
    return lerr
}
// The call to elpsutil.Load put us into the user package, but we started
// in p.name.
lerr = env.InPackage(lisp.Symbol(p.name))
```

So substrate is not broken today. The cost is that the workaround is load-bearing, undocumented on the elps side, and every embedder has to rediscover the same thing independently.

## Suggested direction

Make `PackageLoader` save the active package before calling `PackageInit` and restore it afterwards, matching what #99 established for the stdlib. `Load`/`LoadAll` should be reviewed for the same property.

Two things to decide:

- **Is this a breaking change?** An embedder that currently relies on `PackageInit` leaving them in a different package would see behaviour change. That seems unlikely to be deliberate anywhere, but it is worth a search before assuming.
- **Should the guarantee be tested the way the stdlib's is?** `TestLoadLibrary_EachPackageRestoresPrevious` is the model. An equivalent over `elpsutil.PackageLoader` would stop this regressing, and would have caught it originally.

Once fixed, substrate's manual `InPackage` restoration can be removed — worth a companion issue there rather than leaving dead compensation in place.

Found during the `elpsutil` fuzzing work for #318, which took the package from 6.2% to 96.9% statement coverage; `Load`, `LoadAll`, `LibraryLoader` and `PackageLoader` were all at 0% beforehand, which is why this had never surfaced in elps's own tests.

sam-at-luther:
Fixes #352.

elps already treats "a package load restores the previous package" as an invariant for its own stdlib — #99 fixed the `LoadPackage` functions for violating it, and `TestLoadLibrary_EachPackageRestoresPrevious` pins it. That guarantee did not extend through `elpsutil`, which is the one path elps ships specifically *for* embedders.

## Pre-fix failing tests, written before the fix

Run against unmodified `main`, every new test failed except the two pinning behaviour that was already correct:

- `TestPackageLoader_EachPackageRestoresPrevious/{plain,stationary}` — the loader left the env in the loaded package instead of the caller's.
- `.../switcher`, where `PackageInit` does a nested `elpsutil.Load` — `switcher-fn`, `switcher-op` and `switcher-macro` were bound **and exported in `user`, not in `switcher`**, silently, with a nil return. That is the defect verbatim.
- `.../wanderer` — symbols landed in an unrelated package.
- `TestLoad_RestoresPrevious`, `TestLoadAll_RestoresPrevious`, `TestLibraryLoader_RestoresPrevious` and the `PackageLoader` error/doc cases.

## The fix

`PackageLoader` saves the active package on entry, restores it with `defer` (the idiom #99 put in the stdlib loaders, so error paths restore too), and — the actual bug — re-establishes the loaded package after `PackageInit` returns, before `AddBuiltins`/`AddSpecialOps`/`AddMacros`.

**`Load` and `LoadAll` shared it.** They hard-switched to `lisp.DefaultUserPackage` rather than restoring the caller, so a *nested* load silently relocated its caller — precisely what substrate's comment describes. Both now restore the package active when they were called. For the conventional top-level call from `user`, `prev == user`, so behaviour is unchanged; pinned by `TestLoad_FromUserPackage`. `LibraryLoader` needed no change, since it goes through `Load`.

## Breaking-change assessment

Low risk, checked rather than assumed. A code search for `elpsutil.PackageLoader` / `Load` / `LoadAll` / `LibraryLoader` / `PackageInit` across accessible Go code returns only `luthersystems/substrate` and `luthersystems/svc/libhandlebars`. elps itself never calls any of them, consistent with the 0% coverage #352 notes.

Behaviour differs only when a load starts from a non-`user` package. Every substrate entry point starts in `user`, so `prev == user` and their loader chains are bit-identical. No code was found anywhere that deliberately relies on `PackageInit`/`Load` leaving it elsewhere; the one place that observes the old behaviour documents it as a nuisance and compensates.

## substrate

Confirmed both directions with a test reproducing `ShiroPackage.PackageInit` (nested `elpsutil.Load`, then `InPackage`, then `UsePackage` + `Externals` re-export):

- **With** substrate's manual restore: still correct. It becomes a redundant no-op, not a double switch. **Nothing in substrate needs to change to adopt this.**
- **Without** it: also correct now — which was *not* true of a `PackageLoader`-only fix, because substrate keeps doing package-scoped work after the nested `Load` and inside `PackageInit`. That work only lands in the right package because `Load` now restores the caller.

So substrate can drop its `InPackage(p.name)` workaround once it requires an elps containing this, but only then. No substrate changes are in this PR.

## Verification

`go build ./...` clean, `go test ./... -count=1` all packages pass, `gofmt -l` clean on both touched files, `golangci-lint run ./elpsutil/...` reports 0 issues.

---

### DIFF
--- a/elpsutil/elpsutil.go
+++ b/elpsutil/elpsutil.go
@@ -32,9 +32,12 @@ func (fun *Builtin) Eval(env *lisp.LEnv, args *lisp.LVal) *lisp.LVal {
 }
 
 // Loader is a generic function to initialize/load an LEnv.  A Loader should
-// first define and switch into the package(s) it loads.  Typically, after a
-// Loader executes LEnv.InPackage() is executed to switch back into the user
-// package.
+// first define and switch into the package(s) it loads.  The helpers in this
+// package (Load, LoadAll, LibraryLoader and PackageLoader) restore the package
+// that was active when they were called, so that symbols defined after a load
+// end up where the caller expects them.  A load performed from the user
+// package -- the conventional top-level usage -- therefore ends in the user
+// package, as it always has.
 //
 // A chain of loaders may be formed to load a library.
 type Loader = lisp.Loader
@@ -112,32 +115,29 @@ func packageMacros(p Package) []lisp.LBuiltinDef {
 	return _p.Macros()
 }
 
-// Load loads an elps package implemented in Go.
+// Load loads an elps package implemented in Go.  Load restores the package
+// that was active when it was called so that further defined symbols end up in
+// that package by default.
 func Load(env *lisp.LEnv, fn Loader) *lisp.LVal {
+	prevPkg := env.Runtime.Package.Name
 	lerr := fn(env)
 	if lerr.Type == lisp.LError {
 		return lerr
 	}
-	// Switch back to the user package so that further defined symbols end up
-	// in that package by default.
-	lerr = env.InPackage(lisp.String(lisp.DefaultUserPackage))
+	lerr = env.InPackage(lisp.String(prevPkg))
 	if lerr.Type == lisp.LError {
 		return lerr
 	}
 	return lisp.Nil()
 }
 
-// LoadAll loads multiple elps files implemented in Go.
+// LoadAll loads multiple elps files implemented in Go.  Like Load, LoadAll
+// restores the package that was active when it was called, and runs each
+// loader from that package.
 func LoadAll(fn ...Loader) Loader {
 	return func(env *lisp.LEnv) *lisp.LVal {
 		for _, fn := range fn {
-			lerr := fn(env)
-			if lerr.Type == lisp.LError {
-				return lerr
-			}
-			// Switch back to the user package so that further defined symbols end up
-			// in that package by default.
-			lerr = env.InPackage(lisp.String(lisp.DefaultUserPackage))
+			lerr := Load(env, fn)
 			if lerr.Type == lisp.LError {
 				return lerr
 			}
@@ -160,8 +160,19 @@ func LibraryLoader(ps ...Package) Loader {
 }
 
 // PackageLoader loads an elps package implemented in Go.
+//
+// PackageLoader restores the package that was active when it was called, the
+// same invariant the stdlib LoadPackage functions honor
+// (https://github.com/luthersystems/elps/issues/99).  It also re-establishes
+// the loaded package after PackageInit returns, so that a PackageInit which
+// switches packages -- directly, or indirectly through a nested Load -- does
+// not cause the package's builtins, special operators and macros to be
+// registered and exported somewhere else
+// (https://github.com/luthersystems/elps/issues/352).
 func PackageLoader(p Package) Loader {
 	return func(env *lisp.LEnv) *lisp.LVal {
+		prevPkg := env.Runtime.Package.Name
+		defer env.InPackage(lisp.Symbol(prevPkg))
 		name := lisp.Symbol(p.PackageName())
 		e := env.DefinePackage(name)
 		if !e.IsNil() {
@@ -177,6 +188,13 @@ func PackageLoader(p Package) Loader {
 		initLoader := packageInit(p)
 		e = initLoader(env)
 		if e.Type == lisp.LError {
+			return e
+		}
+		// PackageInit may have switched packages -- the Loader convention
+		// documented above tells embedders to do exactly that.  Re-establish
+		// the package being loaded so the definitions below land in it.
+		e = env.InPackage(name)
+		if !e.IsNil() {
 			return e
 		}
 		for _, fn := range packageBuiltins(p) {
--- a/elpsutil/elpsutil_test.go
+++ b/elpsutil/elpsutil_test.go
@@ -1,0 +1,372 @@
+// Copyright © 2026 The ELPS authors
+
+package elpsutil_test
+
+import (
+	"testing"
+
+	"github.com/luthersystems/elps/elpsutil"
+	"github.com/luthersystems/elps/lisp"
+	"github.com/luthersystems/elps/parser"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// testPackage is a configurable elpsutil.Package used to exercise the package
+// restoration invariant.
+type testPackage struct {
+	name     string
+	doc      string
+	init     func(env *lisp.LEnv) *lisp.LVal
+	builtins []lisp.LBuiltinDef
+	ops      []lisp.LBuiltinDef
+	macros   []lisp.LBuiltinDef
+}
+
+var (
+	_ elpsutil.PackageInit       = (*testPackage)(nil)
+	_ elpsutil.PackageDocumented = (*testPackage)(nil)
+	_ elpsutil.PackageBuiltins   = (*testPackage)(nil)
+	_ elpsutil.PackageSpecialOps = (*testPackage)(nil)
+	_ elpsutil.PackageMacros     = (*testPackage)(nil)
+)
+
+func (p *testPackage) PackageName() string { return p.name }
+
+func (p *testPackage) PackageDoc() string { return p.doc }
+
+func (p *testPackage) PackageInit(env *lisp.LEnv) *lisp.LVal {
+	if p.init == nil {
+		return lisp.Nil()
+	}
+	return p.init(env)
+}
+
+func (p *testPackage) Builtins() []lisp.LBuiltinDef { return p.builtins }
+
+func (p *testPackage) SpecialOps() []lisp.LBuiltinDef { return p.ops }
+
+func (p *testPackage) Macros() []lisp.LBuiltinDef { return p.macros }
+
+// nilFun is a trivial builtin implementation used by test packages.
+func nilFun(env *lisp.LEnv, args *lisp.LVal) *lisp.LVal { return lisp.Nil() }
+
+func testBuiltin(name string) lisp.LBuiltinDef {
+	return elpsutil.Function(name, lisp.Formals(), nilFun)
+}
+
+func newTestEnv(t *testing.T) *lisp.LEnv {
+	t.Helper()
+	env := lisp.NewEnv(nil)
+	env.Runtime.Reader = parser.NewReader()
+	env.Runtime.Library = &lisp.RelativeFileSystemLibrary{}
+	rc := lisp.InitializeUserEnv(env)
+	require.True(t, rc.IsNil(), "InitializeUserEnv: %v", rc)
+	require.Equal(t, lisp.DefaultUserPackage, env.Runtime.Package.Name)
+	return env
+}
+
+// assertBoundIn asserts that sym is bound and exported in package pkg, and
+// that it is not present in any other package.
+func assertBoundIn(t *testing.T, env *lisp.LEnv, pkg string, sym string) {
+	t.Helper()
+	for name, p := range env.Runtime.Registry.Packages {
+		v := p.Get(lisp.Symbol(sym))
+		if name == pkg {
+			assert.NotEqualf(t, lisp.LError, v.Type,
+				"symbol %q should be bound in package %q", sym, pkg)
+			assert.Containsf(t, p.Externals, sym,
+				"symbol %q should be exported from package %q", sym, pkg)
+			continue
+		}
+		assert.Equalf(t, lisp.LError, v.Type,
+			"symbol %q should not be bound in package %q (expected it in %q)", sym, name, pkg)
+	}
+}
+
+// TestPackageLoader_EachPackageRestoresPrevious mirrors
+// lisplib.TestLoadLibrary_EachPackageRestoresPrevious (the test pinning the fix
+// for https://github.com/luthersystems/elps/issues/99) for the embedder-facing
+// elpsutil.PackageLoader.  See
+// https://github.com/luthersystems/elps/issues/352.
+func TestPackageLoader_EachPackageRestoresPrevious(t *testing.T) {
+	env := newTestEnv(t)
+
+	pkgs := []*testPackage{
+		// A package with no PackageInit at all.
+		{name: "plain", builtins: []lisp.LBuiltinDef{testBuiltin("plain-fn")}},
+		// A package whose PackageInit does not move.
+		{
+			name:     "stationary",
+			init:     func(env *lisp.LEnv) *lisp.LVal { return lisp.Nil() },
+			builtins: []lisp.LBuiltinDef{testBuiltin("stationary-fn")},
+		},
+		// A package whose PackageInit switches package directly, as
+		// elpsutil.Loader's documentation tells embedders to do.
+		{
+			name: "switcher",
+			init: func(env *lisp.LEnv) *lisp.LVal {
+				return elpsutil.Load(env, elpsutil.PackageLoader(&testPackage{
+					name:     "switcher-dep",
+					builtins: []lisp.LBuiltinDef{testBuiltin("switcher-dep-fn")},
+				}))
+			},
+			builtins: []lisp.LBuiltinDef{testBuiltin("switcher-fn")},
+			ops:      []lisp.LBuiltinDef{testBuiltin("switcher-op")},
+			macros:   []lisp.LBuiltinDef{testBuiltin("switcher-macro")},
+		},
+		// A package whose PackageInit lands in an unrelated package and never
+		// comes back.
+		{
+			name: "wanderer",
+			init: func(env *lisp.LEnv) *lisp.LVal {
+				name := lisp.Symbol("elsewhere")
+				if e := env.DefinePackage(name); !e.IsNil() {
+					return e
+				}
+				return env.InPackage(name)
+			},
+			builtins: []lisp.LBuiltinDef{testBuiltin("wanderer-fn")},
+			ops:      []lisp.LBuiltinDef{testBuiltin("wanderer-op")},
+			macros:   []lisp.LBuiltinDef{testBuiltin("wanderer-macro")},
+		},
+	}
+
+	for _, p := range pkgs {
+		t.Run(p.name, func(t *testing.T) {
+			before := env.Runtime.Package.Name
+			rc := elpsutil.PackageLoader(p)(env)
+			require.Truef(t, rc.IsNil(), "PackageLoader(%s) failed: %v", p.name, rc)
+			assert.Equalf(t, before, env.Runtime.Package.Name,
+				"PackageLoader(%s) should restore the previous package", p.name)
+			for _, fn := range p.builtins {
+				assertBoundIn(t, env, p.name, fn.Name())
+			}
+			for _, fn := range p.ops {
+				assertBoundIn(t, env, p.name, fn.Name())
+			}
+			for _, fn := range p.macros {
+				assertBoundIn(t, env, p.name, fn.Name())
+			}
+		})
+	}
+}
+
+// TestPackageLoader_RestoresNonUserCaller checks that the package restored is
+// the caller's package, not simply the user package.
+func TestPackageLoader_RestoresNonUserCaller(t *testing.T) {
+	env := newTestEnv(t)
+	caller := lisp.Symbol("caller")
+	require.True(t, env.DefinePackage(caller).IsNil())
+	require.True(t, env.InPackage(caller).IsNil())
+
+	p := &testPackage{
+		name: "callee",
+		init: func(env *lisp.LEnv) *lisp.LVal {
+			return elpsutil.Load(env, elpsutil.PackageLoader(&testPackage{name: "callee-dep"}))
+		},
+		builtins: []lisp.LBuiltinDef{testBuiltin("callee-fn")},
+	}
+	rc := elpsutil.PackageLoader(p)(env)
+	require.True(t, rc.IsNil(), "PackageLoader: %v", rc)
+	assert.Equal(t, "caller", env.Runtime.Package.Name,
+		"PackageLoader should restore the caller's package, not the user package")
+	assertBoundIn(t, env, "callee", "callee-fn")
+}
+
+// TestPackageLoader_PackageDocNotStolen checks that a PackageInit which
+// switches packages cannot cause the package documentation of the loaded
+// package to be applied elsewhere, and that documentation lands on the loaded
+// package.
+func TestPackageLoader_PackageDoc(t *testing.T) {
+	env := newTestEnv(t)
+	userDoc := env.Runtime.Registry.Packages[lisp.DefaultUserPackage].Doc
+	p := &testPackage{
+		name: "documented",
+		doc:  "a documented package",
+		init: func(env *lisp.LEnv) *lisp.LVal {
+			return elpsutil.Load(env, elpsutil.PackageLoader(&testPackage{name: "documented-dep"}))
+		},
+	}
+	require.True(t, elpsutil.PackageLoader(p)(env).IsNil())
+	assert.Equal(t, "a documented package", env.Runtime.Registry.Packages["documented"].Doc)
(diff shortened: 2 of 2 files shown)


iamsamwood (commented):
Independent verification review at head `19f6bb7`, isolated worktree, trial-merged and test-run against current main `663c0f3`. Verdict from this pass: **approve** — the fix is real and correctly idiomatic. Four findings, none blocking; corrections for all four are authored, verified, and attached as patches below (this session can't push to the branch).

## Verification

**Red test.** Reverting only `elpsutil/elpsutil.go` to main's version (byte-identical to the PR base) while keeping the new tests: **7 of 9 fail**, showing builtins/ops/macros registered and exported in the wrong package with a nil success return — the defect verbatim (`sub-a-fn should be re-exported into the cc package … does not contain "sub-a-fn"`, `cc-fn should not leak into the user package`). Restored: all 9 pass. The fix copies the established #99 idiom from the stdlib loaders (`libtime.LoadPackage` uses the identical `prevPkg` + `defer env.InPackage(...)` shape).

**Merge and fuzz interaction.** Merge onto `663c0f3` is textually clean (main's new `elpsutil_fuzz_test.go` is disjoint from the PR's `elpsutil_test.go`). The merged tree passes the full elpsutil suite including the harness's pinned tests, plus 45s of live `FuzzElpsutilEmbed` (148k execs, no crashers). The reason it stays sound: `reserveName` pre-reserves names in *both* the target package and `initSwitchPackage`, so the collision pre-check holds under either placement semantics.

**One correction to the PR description:** it lists the "doc case" among the pre-fix failures, but `TestPackageLoader_PackageDoc` **passes on unmodified main** — `SetPackageDoc` runs before `PackageInit` even in the old code, so the doc was never misplaced. It's a pinning test like `TestLoad_FromUserPackage`, and the description should say so to keep the red-test claim exact.

## Findings and corrections

1. **New panic site on an uninitialized env** (`elpsutil.go:122`, `:174`). `prevPkg := env.Runtime.Package.Name` is a fresh nil-deref when `PackageLoader`/`Load` is called on an env that never ran `InitializeUserEnv` (`Runtime.Package` is nil until `lisp/env.go:28`). Verified empirically: on main, `PackageLoader(p)(env)` on a bare `lisp.NewEnv(nil)` returns success; on this branch it panics. Contract-violating usage, and the stdlib loaders share the hazard — but PR #361 exists precisely to convert elpsutil panics into errors on this surface, so shipping a *new* panic here would be an odd result. **Patch 3 adds the guard** (returns a proper error LVal, tests on a bare env). Note for coordination: #361 edits these same functions; whichever lands second must reconcile.

2. **Three stale comments in main's fuzz harness.** The merge makes them false: the `initTable` comment ("PackageLoader does not re-establish the package after init returns"), the `strategyLabels` comment ("returning to `user` is Load's job"), and `reserveName`'s rationale for checking `initSwitchPackage` (now belt-and-braces rather than a live requirement). This repo's fuzz files pin their own header claims precisely so they can't go stale silently — the first merge after #357 shouldn't be the exception. **Patch 1 fixes these**, plus four more comments in the same file falsified by the same change (`FuzzElpsutilEmbed`'s assertion item 3, `check()`'s quote of the pre-#360 Loader doc, and the two `init/drop-user` comments attributing a separate `InPackage(user)` call to `LoadAll`). Comments only — no assertion weakened, no harness test changed.

3. **`Load`/`LoadAll` don't restore on the error path.** `Load` returns an erroring loader's result *before* the restore, so a Loader that switches packages and then errors still leaves the caller relocated — the same defect #352 fixes for the success path, surviving on the error path. The updated `Loader` doc promises restoration unqualified, which currently only `PackageLoader` (via its defer) delivers. **Patch 2 restores unconditionally before the error check**, preferring the loader's error as the return value, with three regression tests (two of which fail on the parent commit).

4. **Comment/name mismatch** at `elpsutil_test.go:177` (`TestPackageLoader_PackageDocNotStolen` vs `TestPackageLoader_PackageDoc`). **Patch 4.**

**Verification of the corrections** (final state, merged tree): `go test ./elpsutil/... ./lisp/...` all pass, including `TestKnownNilContractViolations`, `TestCollisionPanicIsReal`, `TestElpsutilFuzzReachesEveryFeature`, `TestElpsutilSeedsTerminate`; 20s live fuzz → 79,759 execs, 0 crashers; `gofmt`/`go vet` clean. Teeth confirmed: the two error-path tests fail on the parent commit, and the two uninitialized-env tests panic without the guard.

Apply with `git merge 663c0f3` (verified clean) then `git am` patches 1–4 in order.

<details>
<summary>Patch 1/4 — unstale the fuzz-harness comments that predate the restore fix</summary>

```patch
diff --git a/elpsutil/elpsutil_fuzz_test.go b/elpsutil/elpsutil_fuzz_test.go
index bc4cf5c..4df8e1d 100644
--- a/elpsutil/elpsutil_fuzz_test.go
+++ b/elpsutil/elpsutil_fuzz_test.go
@@ -454,9 +454,10 @@ var bodyTable = []struct {
 
 // initTable is every PackageInit body. `init/error` is the only way to reach
 // PackageLoader's `if e.Type == lisp.LError { return e }` after the init call,
-// and `init/switch` is the only way to reach the state where the package the
-// builtins land in is NOT the package PackageLoader selected -- PackageLoader
-// does not re-establish the package after init returns.
+// and `init/switch` is the only way to reach PackageLoader's re-establishment
+// of the selected package after PackageInit returns (PR #360): an init that
+// leaves the interpreter elsewhere exercises the restore path, and the
+// builtins that follow land in the package PackageLoader selected.
 var initTable = []struct {
 	label string
 	fn    func(*lisp.LEnv) *lisp.LVal
@@ -486,9 +487,10 @@ var initTable = []struct {
 }
 
 // strategyLabels are the four ways the generated packages get installed. Every
-// one goes through elpsutil.Load, because PackageLoader on its own leaves the
-// interpreter inside the package it just defined -- returning to `user` is
-// Load's job, and asserting that postcondition (see install) requires it.
+// one goes through elpsutil.Load: both PackageLoader and Load restore the
+// caller's package (PR #360), and routing every strategy through Load is what
+// lets the "back in `user`" postcondition (see install) be asserted uniformly
+// across them.
 var strategyLabels = []string{
 	"strategy/load-each",
 	"strategy/library-loader",
@@ -586,10 +588,11 @@ func nameAvailable(pkg *lisp.Package, name string) bool {
 
 // reserveName decides whether a def may keep the name the spec gave it.
 //
-// It checks the target package AND initSwitchPackage, because a package whose
-// PackageInit switches packages installs its builtins somewhere other than the
-// package PackageLoader selected, and the harness does not know at reservation
-// time which init body the spec will pick.
+// It checks the target package AND initSwitchPackage. PackageLoader
+// re-establishes the selected package after PackageInit returns (PR #360), so
+// no current path registers a def into initSwitchPackage; the extra check is
+// belt-and-braces against that restore ever being reverted, kept because its
+// only cost is reserving a few names the install would not collide on today.
 func (in *install) reserveName(pkgName, name string) bool {
 	if in.claimed[pkgName] == nil {
 		in.claimed[pkgName] = map[string]bool{}
@@ -941,11 +944,11 @@ func (in *install) check(t fatalf, out *outcome, spec, src []byte) {
 		_ = rc.String()
 	}
 
-	// Load's documented postcondition: "after a Loader executes
-	// LEnv.InPackage() is executed to switch back into the user package", so
-	// symbols defined next land in `user` rather than in whichever package the
-	// last loader happened to leave selected. Only claimed on success --
-	// Load returns early on an error and makes no promise then.
+	// Load's documented postcondition: it restores the package that was
+	// active when it was called, so symbols defined next land in `user`
+	// rather than in whichever package the last loader happened to leave
+	// selected. Only claimed on success -- Load returns early on an error
+	// and makes no promise then.
 	if out.allLoadsOK && out.pkgAfterLoad != lisp.DefaultUserPackage {
 		t.Fatalf("every elpsutil.Load succeeded but the interpreter was left in package %q,"+
 			" not %q. Load exists to restore the user package; code loaded after this"+
@@ -1114,7 +1117,8 @@ func buildSpec(strategy int, pkgs ...seedPkg) []byte {
 //  2. Every elpsutil.Load returns a non-nil *LVal that is either nil or an
 //     LError, and never an internal-panic error.
 //  3. When every load succeeded, the interpreter is back in the `user`
-//     package. This is Load's entire reason to exist over PackageLoader.
+//     package -- the restore contract Load and PackageLoader both honor
+//     since PR #360.
 //  4. Evaluating src against the installed builtins terminates (the watchdog),
 //     returns a non-nil *LVal, never recovers a Go panic (IsInternalPanic),
 //     and produces a renderable result.
@@ -1257,8 +1261,9 @@ func FuzzElpsutilEmbed(f *testing.F) {
 		}), `(debug-print "loaded")`)
 	}
 
-	// An init that leaves the interpreter in another package: the builtins
-	// that follow land there rather than in the package PackageLoader chose.
+	// An init that leaves the interpreter in another package: PackageLoader
+	// re-establishes the package it chose (PR #360), so fz-a must land in
+	// fuzzpkg, where src calls it.
 	add(buildSpec(0, seedPkg{
 		typ: "type/full", name: "fuzzpkg", init: "init/switch",
 		builtins: []seedDef{{"fz-a", "unary", "echo"}},
```
</details>

<details>
<summary>Patch 2/4 — restore the caller's package on Load's error path too</summary>

```patch
diff --git a/elpsutil/elpsutil.go b/elpsutil/elpsutil.go
index 9712736..1fdcb6b 100644
--- a/elpsutil/elpsutil.go
+++ b/elpsutil/elpsutil.go
@@ -34,10 +34,10 @@ func (fun *Builtin) Eval(env *lisp.LEnv, args *lisp.LVal) *lisp.LVal {
 // Loader is a generic function to initialize/load an LEnv.  A Loader should
 // first define and switch into the package(s) it loads.  The helpers in this
 // package (Load, LoadAll, LibraryLoader and PackageLoader) restore the package
-// that was active when they were called, so that symbols defined after a load
-// end up where the caller expects them.  A load performed from the user
-// package -- the conventional top-level usage -- therefore ends in the user
-// package, as it always has.
+// that was active when they were called, whether the load succeeds or fails,
+// so that symbols defined after a load end up where the caller expects them.
+// A load performed from the user package -- the conventional top-level usage
+// -- therefore ends in the user package, as it always has.
 //
 // A chain of loaders may be formed to load a library.
 type Loader = lisp.Loader
@@ -116,24 +116,25 @@ func packageMacros(p Package) []lisp.LBuiltinDef {
 }
 
 // Load loads an elps package implemented in Go.  Load restores the package
-// that was active when it was called so that further defined symbols end up in
-// that package by default.
+// that was active when it was called, whether or not fn succeeds, so that
+// further defined symbols end up in that package by default.  When fn returns
+// an error that error is returned, even if the restore itself also fails.
 func Load(env *lisp.LEnv, fn Loader) *lisp.LVal {
 	prevPkg := env.Runtime.Package.Name
 	lerr := fn(env)
+	rerr := env.InPackage(lisp.String(prevPkg))
 	if lerr.Type == lisp.LError {
 		return lerr
 	}
-	lerr = env.InPackage(lisp.String(prevPkg))
-	if lerr.Type == lisp.LError {
-		return lerr
+	if rerr.Type == lisp.LError {
+		return rerr
 	}
 	return lisp.Nil()
 }
 
 // LoadAll loads multiple elps files implemented in Go.  Like Load, LoadAll
-// restores the package that was active when it was called, and runs each
-// loader from that package.
+// restores the package that was active when it was called, whether or not the
+// loaders succeed, and runs each loader from that package.
 func LoadAll(fn ...Loader) Loader {
 	return func(env *lisp.LEnv) *lisp.LVal {
 		for _, fn := range fn {
diff --git a/elpsutil/elpsutil_fuzz_test.go b/elpsutil/elpsutil_fuzz_test.go
index 4df8e1d..b7b708f 100644
--- a/elpsutil/elpsutil_fuzz_test.go
+++ b/elpsutil/elpsutil_fuzz_test.go
@@ -474,8 +474,8 @@ var initTable = []struct {
 		return lisp.Nil()
 	}},
 	// Removes the user package from the registry. This is the ONLY way to
-	// reach the `if lerr.Type == lisp.LError { return lerr }` guarding
-	// InPackage(user) in Load and LoadAll -- every other argument elpsutil
+	// make the InPackage(prevPkg) restore in Load fail (LoadAll inherits the
+	// error through its delegation to Load) -- every other argument elpsutil
 	// passes to DefinePackage/InPackage it constructed itself and cannot be
 	// wrong. Registry.Packages is an exported map, so a loader really can do
 	// this; whether an embedder should is a different question from whether
@@ -947,8 +947,10 @@ func (in *install) check(t fatalf, out *outcome, spec, src []byte) {
 	// Load's documented postcondition: it restores the package that was
 	// active when it was called, so symbols defined next land in `user`
 	// rather than in whichever package the last loader happened to leave
-	// selected. Only claimed on success -- Load returns early on an error
-	// and makes no promise then.
+	// selected. Only claimed on success -- Load attempts the restore even
+	// when a loader errors (preferring the loader's error), but the restore
+	// itself can fail (init/drop-user deletes the package to restore into),
+	// so success is the only case with an unconditional claim.
 	if out.allLoadsOK && out.pkgAfterLoad != lisp.DefaultUserPackage {
 		t.Fatalf("every elpsutil.Load succeeded but the interpreter was left in package %q,"+
 			" not %q. Load exists to restore the user package; code loaded after this"+
@@ -1216,9 +1218,10 @@ func FuzzElpsutilEmbed(f *testing.F) {
 		), `(fuzzpkg2:fz-b 1)`)
 	}
 
-	// An init that removes the user package, under each strategy. Load and
-	// LoadAll each guard their own InPackage(user) call, and only the strategy
-	// that routes through LoadAll reaches LoadAll's copy of that guard.
+	// An init that removes the user package, under each strategy. This makes
+	// Load's InPackage(prevPkg) restore fail; the strategies that route
+	// through LoadAll and LibraryLoader exercise the propagation of that
+	// error through their delegation to Load.
 	for i := range strategyLabels {
 		add(buildSpec(i,
 			seedPkg{typ: "type/full", name: "fuzzpkg", init: "init/drop-user",
diff --git a/elpsutil/elpsutil_test.go b/elpsutil/elpsutil_test.go
index c3133d4..070f142 100644
--- a/elpsutil/elpsutil_test.go
+++ b/elpsutil/elpsutil_test.go
@@ -244,6 +244,70 @@ func TestLoad_FromUserPackage(t *testing.T) {
 	assert.Equal(t, lisp.DefaultUserPackage, env.Runtime.Package.Name)
 }
 
+// TestLoad_ErrorPathRestoresPrevious checks that Load restores the caller's
+// package even when the loader fails after switching packages, and that the
+// loader's error is the one returned.
+func TestLoad_ErrorPathRestoresPrevious(t *testing.T) {
+	env := newTestEnv(t)
+	rc := elpsutil.Load(env, func(env *lisp.LEnv) *lisp.LVal {
+		name := lisp.Symbol("load-error-elsewhere")
+		if e := env.DefinePackage(name); !e.IsNil() {
+			return e
+		}
+		if e := env.InPackage(name); !e.IsNil() {
+			return e
+		}
+		return env.Errorf("loader failed")
+	})
+	require.Equal(t, lisp.LError, rc.Type, "Load should return the loader's error")
+	assert.Contains(t, rc.String(), "loader failed")
+	assert.Equal(t, lisp.DefaultUserPackage, env.Runtime.Package.Name,
+		"Load should restore the caller's package when the loader fails")
+}
+
+// TestLoad_LoaderErrorPreferredOverRestoreError checks that when the restore
+// itself fails after a loader error, the loader's error is still the one
+// returned.
+func TestLoad_LoaderErrorPreferredOverRestoreError(t *testing.T) {
+	env := newTestEnv(t)
+	rc := elpsutil.Load(env, func(env *lisp.LEnv) *lisp.LVal {
+		delete(env.Runtime.Registry.Packages, lisp.DefaultUserPackage)
+		return env.Errorf("loader failed")
+	})
+	require.Equal(t, lisp.LError, rc.Type)
+	assert.Contains(t, rc.String(), "loader failed",
+		"the loader's error should take precedence over the restore error")
+}
+
+// TestLoadAll_ErrorPathRestoresPrevious checks that LoadAll restores the
+// caller's package when a loader fails, and stops running loaders at the
+// failure.
+func TestLoadAll_ErrorPathRestoresPrevious(t *testing.T) {
+	env := newTestEnv(t)
+	ranAfter := false
+	rc := elpsutil.LoadAll(
+		func(env *lisp.LEnv) *lisp.LVal {
+			name := lisp.Symbol("loadall-error-elsewhere")
+			if e := env.DefinePackage(name); !e.IsNil() {
+				return e
+			}
+			if e := env.InPackage(name); !e.IsNil() {
+				return e
+			}
+			return env.Errorf("loader failed")
+		},
+		func(env *lisp.LEnv) *lisp.LVal {
+			ranAfter = true
+			return lisp.Nil()
+		},
+	)(env)
+	require.Equal(t, lisp.LError, rc.Type, "LoadAll should propagate the loader's error")
+	assert.Contains(t, rc.String(), "loader failed")
+	assert.False(t, ranAfter, "LoadAll should stop at the first failing loader")
+	assert.Equal(t, lisp.DefaultUserPackage, env.Runtime.Package.Name,
+		"LoadAll should restore the caller's package when a loader fails")
+}
+
 // TestLoadAll_RestoresPrevious checks elpsutil.LoadAll restores the caller's
 // package, both between and after the loaders it runs.
 func TestLoadAll_RestoresPrevious(t *testing.T) {
```
</details>

<details>
<summary>Patch 3/4 — error instead of panicking when the env has no current package</summary>

```patch
diff --git a/elpsutil/elpsutil.go b/elpsutil/elpsutil.go
index 1fdcb6b..f0387e6 100644
--- a/elpsutil/elpsutil.go
+++ b/elpsutil/elpsutil.go
@@ -115,13 +115,27 @@ func packageMacros(p Package) []lisp.LBuiltinDef {
 	return _p.Macros()
 }
 
+// currentPackageName returns the name of env's current package, or an error
+// when there is no current package to restore into.  An LEnv that has not
+// been through lisp.InitializeUserEnv has a nil Runtime.Package (and a
+// hand-constructed one may have a nil Runtime); dereferencing either panics.
+func currentPackageName(env *lisp.LEnv) (string, *lisp.LVal) {
+	if env.Runtime == nil || env.Runtime.Package == nil {
+		return "", lisp.Errorf("environment has no current package -- initialize it with lisp.InitializeUserEnv before loading")
+	}
+	return env.Runtime.Package.Name, nil
+}
+
 // Load loads an elps package implemented in Go.  Load restores the package
 // that was active when it was called, whether or not fn succeeds, so that
 // further defined symbols end up in that package by default.  When fn returns
 // an error that error is returned, even if the restore itself also fails.
 func Load(env *lisp.LEnv, fn Loader) *lisp.LVal {
-	prevPkg := env.Runtime.Package.Name
-	lerr := fn(env)
+	prevPkg, lerr := currentPackageName(env)
+	if lerr != nil {
+		return lerr
+	}
+	lerr = fn(env)
 	rerr := env.InPackage(lisp.String(prevPkg))
 	if lerr.Type == lisp.LError {
 		return lerr
@@ -172,7 +186,10 @@ func LibraryLoader(ps ...Package) Loader {
 // (https://github.com/luthersystems/elps/issues/352).
 func PackageLoader(p Package) Loader {
 	return func(env *lisp.LEnv) *lisp.LVal {
-		prevPkg := env.Runtime.Package.Name
+		prevPkg, lerr := currentPackageName(env)
+		if lerr != nil {
+			return lerr
+		}
 		defer env.InPackage(lisp.Symbol(prevPkg))
 		name := lisp.Symbol(p.PackageName())
 		e := env.DefinePackage(name)
diff --git a/elpsutil/elpsutil_test.go b/elpsutil/elpsutil_test.go
index 070f142..aa75f86 100644
--- a/elpsutil/elpsutil_test.go
+++ b/elpsutil/elpsutil_test.go
@@ -365,6 +365,32 @@ func TestLibraryLoader_RestoresPrevious(t *testing.T) {
 	assertBoundIn(t, env, "lib-b", "lib-b-fn")
 }
 
+// TestLoad_UninitializedEnv checks that Load on an environment that never ran
+// lisp.InitializeUserEnv returns an error instead of panicking on the nil
+// current package (Runtime.Package is nil until lisp.InitializeUserEnv).
+func TestLoad_UninitializedEnv(t *testing.T) {
+	env := lisp.NewEnv(nil)
+	called := false
+	rc := elpsutil.Load(env, func(env *lisp.LEnv) *lisp.LVal {
+		called = true
+		return lisp.Nil()
+	})
+	require.NotNil(t, rc)
+	assert.Equal(t, lisp.LError, rc.Type,
+		"Load on an uninitialized environment should return an error")
+	assert.False(t, called, "the loader should not run without a current package")
+}
+
+// TestPackageLoader_UninitializedEnv checks the same for a PackageLoader
+// invoked directly, without going through Load.
+func TestPackageLoader_UninitializedEnv(t *testing.T) {
+	env := lisp.NewEnv(nil)
+	rc := elpsutil.PackageLoader(&testPackage{name: "direct"})(env)
+	require.NotNil(t, rc)
+	assert.Equal(t, lisp.LError, rc.Type,
+		"PackageLoader on an uninitialized environment should return an error")
+}
+
 // TestPackageLoader_SubstrateShape reproduces the shape of
 // luthersystems/substrate's ShiroPackage.PackageInit, which loads a nested
 // library and then explicitly switches back into its own package before
```
</details>

<details>
<summary>Patch 4/4 — align TestPackageLoader_PackageDoc's comment with its name</summary>

```patch
diff --git a/elpsutil/elpsutil_test.go b/elpsutil/elpsutil_test.go
index aa75f86..ffe22ce 100644
--- a/elpsutil/elpsutil_test.go
+++ b/elpsutil/elpsutil_test.go
@@ -174,10 +174,9 @@ func TestPackageLoader_RestoresNonUserCaller(t *testing.T) {
 	assertBoundIn(t, env, "callee", "callee-fn")
 }
 
-// TestPackageLoader_PackageDocNotStolen checks that a PackageInit which
-// switches packages cannot cause the package documentation of the loaded
-// package to be applied elsewhere, and that documentation lands on the loaded
-// package.
+// TestPackageLoader_PackageDoc checks that a PackageInit which switches
+// packages cannot cause the package documentation of the loaded package to be
+// applied elsewhere, and that documentation lands on the loaded package.
 func TestPackageLoader_PackageDoc(t *testing.T) {
 	env := newTestEnv(t)
 	userDoc := env.Runtime.Registry.Packages[lisp.DefaultUserPackage].Doc
```
</details>

Scope note: the `Load`/`LoadAll` semantic change (restore caller vs. hard-switch to `user`) is beyond the minimal `PackageLoader` fix, but #352's "Suggested direction" asks for it, it's behaviour-identical for every top-level call from `user` (pinned by `TestLoad_FromUserPackage`), and it's required for the substrate `PackageInit` shape to work without the workaround. In scope.

---

Issue #367, opened 2026-08-11
An elps program must never panic the host: lisp-reachable failures raise catchable conditions


sam-at-luther:
Scope clarification from the maintainer: "elps should never panic" means an **elps program** should never panic. The criterion for this sweep is lisp-reachability, not the raw count of `panic(` sites in the Go code.

## The invariant

No program text evaluated by the interpreter, over any data, may take down the host process with a Go panic. Every failure a lisp program can provoke - bad types reaching a builtin, malformed shapes, arity mistakes, resource limits - must surface as a catchable condition that `handler-bind` can handle, the way `CondMissingArgument` does.

Panics that only embedder Go code can trigger (API misuse at integration time, violated internal invariants unreachable from lisp source, and the deliberate elpscheck detectors) are ordinary Go practice and stay panics. Registration-time embedder mistakes are tracked separately in #351.

## The precedent to follow

#355 (merged as 86db2e4) is the model. `BuiltinSleep` began reading `args.Cells[1]` for an `&key` argument, which is safe under the evaluator's calling convention and unsafe under an embedder binding its own formals, so every `cc:sleep` call in a downstream embedder panicked with an index-out-of-range, surfaced as an opaque `internal-panic` that `handler-bind` is documented not to catch.

The fix was not a bounds check at the call site. It was `LVal.KeyArg`/`ReqArg` plus `CondMissingArgument` (`lisp/conditions.go`), then a registry-walking test (`TestOptionalArgBuiltinsTolerateShortArgLists`) that finds the next instance without anyone remembering the failure mode. The elpspath list-rejection fix for luthersystems/substrate#378 followed the same shape: a catchable error where an uncatchable panic (with prior state corruption) used to be.

That is the shape this sweep should take: a condition for the caller, plus a walking test or fuzz target that finds the next one.

## Method

1. Classify each panic site by **lisp-reachability**: can any evaluated program, over any data, reach it? Type-assertion accessors used by builtins (`lisp/lisp.go:768,779,786,888,897` "not an error:/not a function:/not bytes:/not sorted-map:", `lisp/builtins.go:2740` toFloat, sequence/array shape checks at `lisp/lisp.go:1430,1434`) are the likely in-scope group. Internal invariant panics (`lisp/env.go:1390` invalid mark, `lisp/lisp.go:718,725,732` marker-tail-recursion, `lisp/stack.go:294`) are presumed out of scope, but each needs a one-line reachability argument recorded in a comment, not an assumption.
[...]

Issue #364, opened 2026-08-11
libschema.NewValidator returns a value that is not safe to bind into more than one Runtime


sam-at-luther:
`libschema.NewValidator` is the exported extension point for Go-implemented schema constraints (`lisp/lisplib/libschema/libschema.go:474`):

```go
func NewValidator(formals *lisp.LVal, fn lisp.LBuiltin) *lisp.LVal {
	return newValidator(formals, fn)
}
```

The value it returns is an LFun carrying per-value mutable state: `Cells` (formals, docstring, and the validator marker appended by `markValidator`) plus its `FunData`. Binding one such value into two independent `LEnv`/`Runtime` pairs shares that state across runtimes.

Nothing in the signature, the doc comment, or the package documentation says the result is single-runtime. An embedder that builds its constraint set once at process start and installs it into every environment it creates, which is the natural reading of an extension point, gets silent cross-runtime sharing with no diagnostic.

## How it was found

An experimental `elpscheck`-gated ownership checker (a process-wide `*LVal -> *Runtime` side table that panics when one value is seen under two runtimes) flagged this on its first full-suite run.

The in-tree offender was this repository's own test. `TestNewValidatorIsAcceptedAsAConstraint` (`lisp/lisplib/libschema/foreign_constraint_test.go:194-227` on main, 86db2e4) mints one validator and binds it into `newSchemaEnv(t)` and then into one further fresh environment per table case. `newSchemaEnv` calls `lisp.NewEnv(nil)` (`lisp/lisplib/libschema/libschema_test.go:84-96`), so each is a separate `Runtime`: six runtimes sharing one validator.

That the repository's own demonstration of the extension point uses it this way is the evidence that the constraint is not discoverable from the API.

## Status

The test was corrected on the unmerged branch `claude/lval-hermetic-full`, in commit [1f3835a](https://github.com/luthersystems/elps/commit/1f3835ab7ca53d80af7ede88dbbfab08fe6c3298), which now mints one validator per environment and states the rule in a comment.

The exported API is unchanged and is what this issue tracks. `NewValidator` should either document the single-runtime requirement or return something that is safe to share. Filing so the finding is not lost if that branch stalls.

## Related

- #363, builtin formals shared across every `LEnv` in the process: the same hazard class reached through a different door.
- #362, the shared native source `Location`.

PR #361, opened 2026-08-10
Return errors instead of panicking on embedder mistakes in elpsutil

sam-at-luther:
Fixes #351. Three embedder mistakes that panicked — in the embedder's process, or deferred to call time — now return an `*LVal` error naming what was wrong and which thing was wrong.

## Nothing under `lisp/` changed behaviour

The only file added there is a test. The panics in `LEnv.bind`, `AddBuiltins` and `Package.get` are untouched, deliberately. A panic in `lisp` means "this is a bug in the interpreter, or in code that had no business calling this"; weakening it blunts a real signal to serve a case the boundary serves better — and serves it with a far more useful message, because elpsutil knows the package name, the definition kind and the definition name, where `lisp` at that point knows only a symbol.

## What an embedder now sees

1. **nil formals** — `package "acme": builtin "do-thing": formals are nil (use lisp.Formals() to declare a function that takes no arguments)`. Also covers formals that are an `LError` (what `lisp.Formals("&rest","a","b")` returns), a non-list, a nil cell, and a non-symbol cell.
2. **nil loader result** — `loader github.com/acme/pkg.MyLoader (pkg.go:41) returned a nil *LVal (a Loader must return lisp.Nil() on success)`, identified via `runtime.FuncForPC`; in a chain, `loader 2 of 5, …`; from a package, `package "acme": PackageInit returned a nil *LVal …`.
3. **reserved / taken name** — `package "acme": builtin "true": "true" is a reserved constant and can never be bound`, and for a real collision `package "lisp": builtin "car": symbol is already defined in package "lisp"`.

Registration routes through `addDefs`, which mirrors the collision rules of `AddBuiltins`/`AddSpecialOps`/`AddMacros` exactly (`AddBuiltins` rejects any existing binding; the other two tolerate overwriting a nil-valued one), so a definition `lisp` accepts today is still accepted.

## The silent `Builtins()` trap, addressed

#351 records a harder related problem: a `Package` whose `Builtins()` signature does not exactly match `Builtins() []lisp.LBuiltinDef` fails the type assertion in `packageBuiltins`, which returns nil — so **every builtin silently vanishes** and `Load` returns success. I hit this myself while writing #351's reproductions.

Go cannot detect an interface someone *meant* to implement, but it can detect a **near miss**. `checkPackageMethods` uses reflect: for each optional interface, if the type does not satisfy it but does have a method of that name, it reports have/want; if the method exists only on the pointer type and the package was passed by value, it says so.

```
[...]

PR #374, opened 2026-08-11
LVal hermetic sealing: zero-cost stack + seal/copy-on-write + verification tooling

### LINKED ISSUE #362: nativeSource() hands out a shared mutable Location that no guard covers
`nativeSource()` returns `defaultSourceLocation` — a package-level `*token.Location` shared by every natively-constructed `LVal`:

```go
var defaultSourceLocation = &token.Location{
	File: "<native code>",
	Pos:  -1,
}

// TODO(elps2): make the LVal.Source "immutable" (possibly an interface or a
// string) so it won't matter that nativeSource returns a shared reference.
func nativeSource() *token.Location {
	return defaultSourceLocation
}
```

Any in-place write to one of its fields — `v.Source.Pos = 7` on a value from `lisp.Int(1)`, say — corrupts it for **every value in the process**, and the damage surfaces somewhere entirely unrelated to the write.

## Not currently reachable in production, and worth saying so plainly

I checked before filing. Parsing does **not** corrupt it:

```
before parse: {File:<native code> Pos:-1 ...}
'()          -> shared loc unchanged
'(a b c)     -> shared loc unchanged
#'car        -> shared loc unchanged
(quote x)    -> shared loc unchanged
```

`parser/rdparser/parser.go` does write `Source` fields in place (`inheritEndPos`, `applyPrefixLocation`, and the end-position fixups), but those operate on Locations `tokenLVal` allocates per node, never on the shared one. So this is a latent trap, not a live defect, and it does not warrant an urgent fix.

## It is a trap people actually fall into

This was found the hard way. `internal/fuzzfp`'s test suite (added in #356) had a case named "shared synthetic edited in place" that took a value straight from `lisp.Int(1)` and wrote `v.Source.Pos = 7` into it. That permanently corrupted `defaultSourceLocation` for the rest of the test binary, and `TestGuardPermitsEveryReplacementOfASyntheticLocation` then failed its own precondition — "expected a synthetic Source on a constructed value" — whenever it happened to run afterwards. Both tests call `t.Parallel()`, so which one won was a race: **2 failures in 8 runs**, and CI had gone green on luck.

Two things make that instructive rather than just embarrassing:

- The write looked completely innocuous at the point it was written, and the failure appeared in a different test asserting something unrelated.
- It happened *inside the package whose entire purpose is detecting mutation of shared state*.

Fixed on that branch by giving the case its own synthetic location, but that fixes the instance, not the hazard.

## Nothing guards it

`TakeSingletonSnapshot` covers exactly three singletons:

```go
return SingletonSnapshot{
	nilV:   *singletonNil,
	trueV:  *singletonTrue,
	falseV: *singletonFalse,
}
```

`defaultSourceLocation` is a fourth shared mutable singleton with no equivalent protection, and `checkSingleton` does not cover it either. elps has been through this class twice already — #334 stopped `lambdaVars` mutating the shared nil singleton and made such writes detectable, and #333 covered the write-same-value race — so the pattern is established; the source location was simply missed.

## Options

Roughly in increasing order of cost:

1. **Extend the singleton guard** to snapshot `*defaultSourceLocation` alongside the three `LVal` singletons. Cheapest, and it turns a silent action-at-a-distance flake into a named failure. Does not prevent the write.
2. **Hand out a fresh Location per call.** Removes the hazard entirely at the cost of an allocation per natively-constructed value — probably unacceptable on that path, but worth measuring rather than assuming.
3. **Make `LVal.Source` immutable**, which is what the existing TODO proposes (an interface, or a string). The real fix, and the largest.

I would suggest (1) now regardless of which of (2)/(3) is eventually chosen, since it is small and converts the failure mode from "mysterious flake in an unrelated test" into "the guard names the offender".

### LINKED ISSUE #363: Builtin formals LVals are shared across every LEnv in the process
Sibling of #362, found while sweeping for other instances of that class. This one is broader: the shared state crosses `LEnv` boundaries, so a single write would corrupt **every environment in the process**, concurrently.

## The finding

Nine packages register builtins from a package-level table initialised once at package load:

```
lisp/op.go              var langSpecialOps = []*langBuiltin{...}
lisp/macro.go           var langMacros     = []*langBuiltin{...}
lisp/lisplib/libtime    var builtins = []*libutil.Builtin{...}
lisp/lisplib/libregexp  ""
lisp/lisplib/libhelp    var ops      = []*libutil.Builtin{...}
lisp/lisplib/libschema  ""
lisp/lisplib/libmath    ""
lisp/lisplib/libbase64  ""
lisp/lisplib/libstring  ""
```

Each entry holds `formals *lisp.LVal`. That `*LVal` is constructed once and handed to every `LEnv` that loads the package, so the formals are shared process-wide.

Verified empirically with two independently constructed environments, each with the full stdlib loaded:

```
time:sleep         LFun shared=false  formals LVal shared=true
json:dump-string   LFun shared=false  formals LVal shared=false
math:atan          LFun shared=false  formals LVal shared=true
lisp:map           LFun shared=false  formals LVal shared=true
```

The `LFun` wrappers are per-environment. The formals `LVal` inside them is the **same pointer**.

`json:dump-string` is the control that proves the pattern is the cause: libjson builds its table through `Builtins(s *Serializer)`, a function called per load, so it gets fresh formals. The packages using a package-level `var` do not.

## Why this is worse than #362

`defaultSourceLocation` is shared between values in one process. This is shared between **environments**, and elps embedders run many concurrently:

- `luthersystems/substrate` caps its MCP environment pool at 64 live environments, and phyla evaluate concurrently across goroutines.
- A single in-place write to a formals `LVal` — `formals.Cells[0].Str = ...`, `formals.Cells = append(...)` — would therefore be both a cross-environment correctness bug and a **data race**, with the damage appearing in an environment that never executed the offending code.

Nothing covers it. `TakeSingletonSnapshot` guards `Nil()`, `Bool(true)` and `Bool(false)`. The argument-mutation guard added in #356 watches a builtin's *arguments*, not its formals.

## Reachability: latent, as far as I can tell

I checked, and did not find a live path:

- `lisp/op.go:280` (`formals.Cells[0] = Symbol("%")`) operates on `formals := SExpr(nil)`, constructed fresh inside `opExpr`. Not shared.
- `lisp/env.go:1544` and `lisp/builtins.go:1067` read `fun.Cells[0]`; they do not write to it.
- `lisp/lisplib/libschema/libschema.go:448` (`fun.Cells = append(fun.Cells, validatorMarker)`) appends to the per-environment `LFun`'s own cell slice, which `Fun()` allocates per registration. It does not reach the shared formals.

So this is a trap rather than an active defect — the same status as #362, and worth the same treatment: fix the hazard, not just wait for someone to fall in.

## Why "latent" should not be reassuring here

#362 was found because someone *did* fall into the equivalent trap: a test wrote `v.Source.Pos = 7` on a natively-constructed value, corrupted a process-wide singleton, and surfaced as a 25% flake in an unrelated assertion. That write looked entirely innocuous where it was written.

The formals case has the same property — `formals.Cells[...] = ...` reads as local surgery on one function — but a worse blast radius, since the damage crosses environments and races.

## Options

1. **Detection**: extend the singleton guard to fingerprint the registered formals, so a write is named rather than surfacing elsewhere. Cheapest; does not prevent.
2. **Fresh formals per registration**: have `PackageLoader`/`libutil` deep-copy the formals when binding, or convert the remaining packages to the `Builtins(...)` function form libjson already uses. Cost is one copy per builtin per environment load — bounded by package size, paid at load rather than in the hot path. Probably the right trade, but should be measured rather than assumed.
3. **Immutable formals**, the same direction #362 points for `Source`.

(2) is worth measuring first: unlike #362, where per-construction allocation cost ~23% of interpreter throughput, this cost is paid once per environment load rather than per value.

### LINKED ISSUE #365: elpstest.RunBenchmark evaluates one parsed AST in a fresh Runtime per iteration
`elpstest.RunBenchmark` parses the source once, outside the loop, and then creates a new `Runtime` on every `b.N` iteration while evaluating the same `[]*lisp.LVal` (`elpstest/lisptest.go:386-411` on main, 86db2e4):

```go
func RunBenchmark(b *testing.B, source string) {
	b.StopTimer()
	p := parser.NewReader()
	exprs, err := p.Read("benchmark", strings.NewReader(source))
	...
	for range b.N {
		env := lisp.NewEnv(nil)
		...
		b.StartTimer()
		for i, expr := range exprs {
			lerr := env.Eval(expr)
```

Every iteration therefore evaluates AST nodes that previous iterations already evaluated under a different `Runtime`. This is exactly the sharing `lisp.TextLoader` defends against: `TextLoader` calls `expr.Copy()` per load for this reason (`lisp/loader.go`).

## Why it matters beyond tidiness

`RunBenchmark` is the standard harness for this repository's lisp benchmarks, so every historical number produced through it was measured on cross-runtime-shared ASTs rather than on the per-iteration-fresh trees the benchmark is meant to model. Two consequences:

- Any evaluator behaviour that depends on AST node state being fresh (macro expansion stamping, source stamping, tail-recursion markers) is measured in a state the benchmark did not intend.
- If a future change makes eval write to an AST node, the effect accumulates across iterations instead of being isolated to one, which is a misleading benchmark rather than a failing test.

## Status

Fixed on the unmerged branch `claude/lval-hermetic-full`, in commit [1f3835a](https://github.com/luthersystems/elps/commit/1f3835ab7ca53d80af7ede88dbbfab08fe6c3298). The fix copies the expressions per iteration and does so outside the timed region, so `ns/op` stays comparable with historical numbers:

```go
iterExprs := make([]*lisp.LVal, len(exprs))
for i, expr := range exprs {
	iterExprs[i] = expr.Copy()
}
b.StartTimer()
```

Found by an experimental `elpscheck`-gated ownership checker on the same branch. Filing so the fix is not lost if that branch stalls.

## Related

- #363 and #362, the static half of the same hazard class.

### LINKED ISSUE #366: ErrorAssociate stores env.Loc by reference into the error it associates
`LEnv.ErrorAssociate` stamps the environment's current source location onto an error value by storing the pointer, not a copy (`lisp/env.go:877-893` on main, 86db2e4):

```go
if lerr.Source == nil || lerr.Source.Pos < 0 {
	lerr.Source = env.Loc
}
```

`env.Loc` is a `*token.Location` the evaluator keeps using and rebinding after the error value has escaped to the caller. The error and the evaluator therefore alias one `Location`. Any in-place write through `env.Loc` retroactively changes the position already reported by an error that was raised earlier, and the change surfaces at a point unrelated to the write.

This is distinct from #362. #362 concerns the process-wide `defaultSourceLocation` singleton handed out by `nativeSource()`. This one is about the per-environment current location, which is a different object with the same aliasing shape, and it survives even if `nativeSource()` is fixed.

## Severity

Latent rather than live: this needs an in-place write to a `Location` that an already-raised error is holding, and the in-tree writers (`parser/rdparser`'s `inheritEndPos` and `applyPrefixLocation`, the end-position fixups) operate on locations the parser allocates per node. It is filed as a correctness trap in an error path, where a wrong reported position is particularly expensive to debug, rather than as an urgent defect.

## Status

Fixed on the unmerged branches `claude/lval-hermetic-full` and `claude/lval-hermetic-zerocost`, in commit [d922290](https://github.com/luthersystems/elps/commit/d922290375aa9d050316903afbab88df783c50fb): the location is stored through `copyLocation`, which preserves nil so the `<native code>` convention is unchanged.

The commit adds `TestErrorAssociateCopiesLocation` (`lisp/eval_safety_test.go`), which asserts three things: the error's location is pointer-independent of `env.Loc`, mutating `env.Loc` in place afterwards does not move the error's recorded position, and a nil `env.Loc` still yields a nil error source. It fails on the unfixed code and passes on the fixed code.

Found by the freshness audit on that branch. Filing so the fix is not lost if the branches stall.

### LINKED ISSUE #369: Quoted program literals can be corrupted in place from pure lisp (stable-sort; append 'vector via slice-retained capacity)
Two lisp-reachable mechanisms mutate the backing of a quoted list literal on current main. Because evaluating a quote hands out the parsed AST node itself, and embedders can share parsed ASTs across environments (parse caches, luthersystems/substrate#375), this is the same corruption class as luthersystems/substrate#378, reachable without any embedder Go code.

## Mechanism 1: stable-sort is documented in-place, and lists can be literals

```lisp
(defun s () (let ([q '(1 2 3)]) (debug-print q) (stable-sort > q)))
(s) ; prints (1 2 3)
(s) ; prints (3 2 1)  -- the literal itself was sorted in place
```

This falsifies the assumption that all mutating builtins type-error on lists: stable-sort mutates lists by design, so the "literals are never mutation targets" invariant does not hold on main.

## Mechanism 2: append 'vector writes through capacity retained by slice

```lisp
(set 'lit '(1 2 3))
(append 'vector (slice 'list lit 0 1) 99)
lit ; => (1 99 3)
```

`slice 'list` shares the source backing array including spare capacity; `append 'vector` then writes into that capacity. Notably the eager leak-point-copy design (branch claude/exp-ast-leakpoints) does NOT stop this shape for the per-environment value copy; the seal/copy-on-write design does.

## Fix

Both are fixed on branch `claude/exp-cow-seal` (parser-set seal flag + copy-on-write at the mutation sites), with regression tests `TestStableSortLiteralPristine` and `TestAppendVectorSliceBackingPristine`, red-proofed with the seal disabled. stable-sort on a sealed list sorts a fresh-backing shallow copy (element identity preserved); append/slice copy before writing into shared backing.

## Why the checkers missed it

Both writes are laundered through a local slice alias (`cells := seqCells(list); cells[i] = ...`), which the elpsvet freshness rule does not currently flag. Tracked separately.

### LINKED ISSUE #370: stampMacroExpansion can write metadata onto shared parser AST nodes (cross-env write, data race)
`stampMacroExpansion` stamps call-site locations onto expanded nodes that carry synthetic (Pos&lt;0) source locations. The parser itself emits synthetic locations in places (the `lisp:function` head of a funref, the `#^` head-symbol nil-source case), so the stamp is reachable on parser-produced AST nodes. When an embedder shares a parsed AST across environments (parse cache), that is a cross-environment metadata write, and a data race under concurrent environments.

Severity is lower than value corruption (the write is `.source`/`.MacroExpansion` metadata, not cells), but it is an unsynchronized write to shared memory.

Fixed on branch `claude/exp-cow-seal` by skipping sealed subtrees in the stamp walk. Behavior divergence is error-message-only: sealed edge-case heads keep native/absent source instead of a call-site stamp. Related history: #274 added the identity-based singleton guard to this same walk.

### LINKED ISSUE #371: elpsvet freshness rule: close the local-slice-alias laundering gap
The elpsvet freshness rule (branch claude/exp-freshness-vet, now part of the hermetic stacks) flags LVal field writes on values not constructed in the same function. It does not flag writes laundered through a local slice alias:

```go
cells := seqCells(list)
cells[i] = x // mutates list.Cells[i], invisible to the current rule
```

This gap is exactly why the two literal-corruption bugs in #369 were invisible to the 48-flag census. The rule should track slice aliases of `.Cells` (and bytes/array backings) at least one assignment deep, or flag index writes through any slice derived from an LVal accessor.

### LINKED ISSUE #372: Seal verification tooling: fuzz corruption oracle, checked-mode AST inspector, -race seal watchdog
With the seal/copy-on-write design (branch claude/exp-cow-seal), shared program AST nodes are marked in the data itself, which makes the core invariant directly checkable: **sealed bytes never change after parse**. This issue tracks the verification tooling that checks that property mechanism-blind, so corruption is detected no matter which code path causes it. Motivated by #369 and #370, which shipped past both a manual audit and the elpsvet census (#371 tracks the static-rule gap).

## Deliverables (all elps-side, kernel-owned)

1. **Fuzz corruption oracle.** Extend the eval fuzz targets: fingerprint the sealed AST of each parsed input, assert the fingerprint is unchanged after every execution. Reuses the existing fingerprint-walk approach. Acceptance: with the stable-sort copy-on-write fix reverted, the oracle finds the corruption from a seeded corpus quickly; with the fix in place, a bounded run is clean.

2. **Checked-mode AST inspector.** Under the elpscheck build tag: record a fingerprint of every sealed parse (parser exit or Program creation), re-verify at environment teardown and at test end. Zero cost when the tag is off (empty inlined hooks; off-tag binaries must be verified unchanged). Embedders get full detection by running their suites with the tag; no embedder-side checker required.

3. **-race seal watchdog.** Generalize the singleton write watchdog from #334: test-only machinery that keeps an unsynchronized read of sealed nodes live, so under -race any unsynchronized write to a sealed node is reported with a stack naming the offender, same-value writes included. Acceptance: with the stampMacroExpansion sealed-skip guard (#370) reverted, the watchdog reports the write deterministically under the existing race job.

4. **Red-proofs for all three**, plus perf confirmation: untagged build behavior unchanged (byte-identical binaries or interleaved benchstat showing zero delta), tagged overhead measured and reported as CI-only cost, fuzz exec-rate overhead reported.

## Non-goals

- No checker or analyzer runs in embedder repos; embedders consume the tooling by running checked-mode CI (and optionally adding their own fuzz targets that reuse the oracle).
- Static-rule improvements are tracked separately in #371.

Related: #369, #370, #371, luthersystems/substrate#378.

### LINKED ISSUE #382: Unexport the LVal metadata/flag fields (Quoted, Spliced, Meta, MacroExpansion) in the same release break
Maintainer directive: break the API once, not whack-a-mole across releases. The current release already breaks `LVal.Source`; the remaining metadata/flag fields join it now.

## Census (downstream production usage)

- `Spliced`: 0 uses — unexport outright.
- `MacroExpansion`: 0 production uses — unexport (debugger machinery is in-repo).
- `Meta`: 0 downstream uses — unexport; elps's own formatter/minifier/format-preserving parser need an in-repo bridge (the `internal/astraw` two-stage hook is the existing pattern) or in-package accessors.
- `Quoted`: 1 downstream read + tests — unexport the field, add a read accessor (`Quoted()` method or `IsQuoted()`), keep `Quote`/`shallowUnquote`-style construction as the only write paths.

Staying exported: `Native` (55 downstream production uses — native handles legitimately live there), and the data-read surface `Type/Str/Int/Float/Cells` (read-accessor migration priced at ~3,000 sites and rejected; writes are covered by the seal, elpsvet rules, and checked-mode verification).

## Why this set

Every historical metadata corruption went through these fields: the #333/#334 singleton race was a `Quoted` write; #370's stamp wrote `MacroExpansion`/source metadata onto shared parser nodes; the post-seal leak fixes (macroexpand &amp;rest / format Meta) were `Meta`-adjacent writes. Unexporting closes the external write channel for the class at compile time, complementing the seal (runtime) and elpsvet (static, in-repo).

## Notes

- elps-internal packages outside `lisp/` (parser, formatter, minifier, elpstest, repl, lsp) get accessors or internal bridges as needed — census and price per package in the implementation.
- Downstream migration: one read-site swap to the accessor + test updates; patch alongside the existing Source-accessor migration.
- Lands on the current release train (PR #374 branch) so the API breaks once.

sam-at-luther:
## Summary

The cumulative elps side of the LVal shared-mutation hardening work. **Design doc: [`docs/sealed-ast.md`](https://github.com/luthersystems/elps/blob/claude/exp-seal-tooling/docs/sealed-ast.md)** — the IsSealed mechanism, the threat model (all six shipped corruption vectors), what each verification layer prevents and its blind spots, the footguns, and the embedder checklist.

Layers, by commit range:

1. **Zero-cost core** (`main..d922290`, review fixes through `d47b58d`): formals boundary copy, `LVal.Source` unexport + source-location singleton deletion, in-kernel `detach()` deep-copy machinery (unexported — zero consumers today; re-exporting is additive and waits for a real embedder consumer), elpscheck ownership tag, sealed registry/package accessors, opaque `Program` type, elpsvet freshness rule, error-location copy fixes. Breaking API changes are compile-time only; substrate migration verified end-to-end. *(The formals boundary copy was later withdrawn — see layer 4's `2110ac5`, seal-and-share.)*
2. **Seal / copy-on-write** (`d922290..4c7649a` + leak fixes `8d18071`): parser-set `sealed` bit in an existing LVal padding byte (struct stays 112 bytes); quote/macro/lambda hand sealed nodes out by reference (no copies); mutation sites copy-on-write. Kills the cross-env AST corruption class (substrate#378) including the pure-lisp mechanisms (#369) and metadata writes (#370). Warm-transaction cost: geomean −0.11% vs no protection (statistically zero; allocations bit-identical).
3. **Verification tooling** (#372): shared sealed-AST fingerprint walker; fuzz corruption oracle; elpscheck-gated inspector; `-race` seal watchdog. All red-proofed against the real shipped bugs; untagged builds link zero seal symbols; tagged overhead ~+3% suite wall (CI-only).
4. **Assembly** (`2b74f29..HEAD`): merges of the reviewed base (`d47b58d`), the elpspath package (`95bbd0c`, #172/#380 line), and the vet slice-alias rule (`9443169`, #371); the formals **seal-and-share** replacement for layer 1's boundary copy (`2110ac5`, after the CI benchmark gate flagged the copy at +9.32% B/op on libjson `Package/$load`); elpsvet now exits binary-clean (50 `//elps:mutates` + 19 `//elpsvet:allow`, individually justified); design doc.
5. **Field privatization** (#382, `5819cdc..HEAD`): the compile-time half of the seal. Every historical metadata corruption went through an exported field, so `LVal`'s `Quoted`/`Spliced`/`Meta`/`MacroExpansion` and `LEnv`'s `Scope`/`FunName`/`Parent`/`Loc` are unexported, with mediated reads (`IsQuoted()`, `MacroExpansion()` snapshot, `Bindings()`, `Parent()`, `Source()` copies), in-repo `internal/` hook bridges for the format tooling, the debugger and the profiler, and `MapData`'s backing fixed at construction. Census-driven: the kept surface (`Native`/`Str`/`Cells`/`Type`/`Int`/`Float`/`FunType` at ~3,000 downstream sites, `Runtime`, `CallStack`) is argued in the design doc as explicitly as the breaks. Downstream cost is 7 mechanical read swaps; interleaved benchstat n=8 shows the accessor indirection free (geomean +0.61%, no row significant). `TestLValFieldSeal`/`TestLEnvFieldSeal` are the regression guards; the latter was red-proofed by re-exporting the scope map (fails, reverted clean).
6. **libelpspath copying-op corruption fix** (`0442c52`): the non-mutating `?set`/`?del`/`?nil` wrote the ARRAY memory layout into LIST copies — short lists panicked uncatchably, longer lists came back structurally mangled with no error at all. The defect is in code this PR moves upstream (layer 4's elpspath package), so the fix ships here rather than in the stacked optimization PR that found it. `storeCells` dispatches on the container layout; `TestCopyOpsOnList` is the regression guard, red-proofed by forcing the array branch. Found by #387's `FuzzPathEngine`.

## Review status — all three independent adversarial review passes returned READY

- **Core stack** (`main..d922290`): READY after 4 confirmed fixes (error-constructor location aliasing, Detach empty-map aliasing, allowlist correction, profiler attribution) — all with regression tests.
- **elpspath move + vet rule**: READY after fixes `95bbd0c` (two ported-engine crash bugs) and `9443169` (two taint-tracking gaps).
- **Seal + tooling layer**: READY at `dee7317` — laundering battery (40 lisp programs) could not construct a corruption; red-proofs verified non-vacuous at this head; untagged binaries carry zero seal symbols. Two documented residuals (format-preserving reader contract, unbounded sealAST depth on hand-built embedder chains) are design boundaries recorded in the doc, not defects.

Sign-off items for the maintainer: the debugger native-expr hook behavior change (deliberate, documented); the elpspath default-load decision (the positional-arg elpspath API ships in default envs); descendant substrate pin tracks this head.

Closes #362, Closes #363, Closes #365, Closes #366, Closes #369, Closes #370, Closes #371, Closes #372, Closes #382 — each verified against this branch, not against intent.

**#364 is deliberately NOT on that list.** The branch fixes only this repository's own multi-runtime test (`foreign_constraint_test.go` now mints one validator per environment); `libschema.NewValidator`'s signature and doc comment are byte-identical to `main`, and that exported API is what the issue tracks. Also not closed here, by design: #367 (never-panic sweep), #378 (a lisp-level `copy` builtin — `detach()` stays unexported until it lands), #379–#381, #383.

## Relationship to open PRs

Verdicts are against this branch's content at `0442c52`: each queue PR was trial-merged into the branch and the merged tree compiled.

### Pre-existing fix queue (#356–#361)

**Status update — this section was written when all six were open on `main`. Since then #356 (663c0f3) and #357 (f5d3d9e) have MERGED to `main`, alongside #384 (b1ab958) and #391 (050ade5); `main` is now 663c0f3.** #358–#361 remain open. The per-PR verdicts below still describe the work correctly, but for #356 and #357 read them as *what the forward-merge of `main` into this branch must resolve*, not as advice about a merge order that has already been decided.

| PR | Verdict | Evidence |
|---|---|---|
| #356 builtins do not mutate their arguments + stdlib fuzz budget — **merged to `main`** | **Conflicting** | `lisp/lisplib/fuzz_test.go` conflicts. The corruption-oracle work (`388e222`, `7dda246`) rebuilt `FuzzApplyStdlib` around the sealed-AST fingerprint (`lisp/sealfp.go`), which is strictly stronger than `internal/fuzzfp` for sealed arguments; and `internal/fuzzfp`'s whole `Source` premise — one process-wide `nativeSource()` `Location` writable through an exported field — is deleted by `9edf260` + `69be094`, so that package no longer compiles here. Its LNative-contents check, `FuzzApplyThinStdlib` and the `.github/workflows/fuzz.yml` change are **not** on this branch and are still wanted. |
| #357 fuzz `elpsutil` (6.2% coverage) — **merged to `main`** | **Independent** | Touches only `elpsutil/elpsutil_fuzz_test.go` and `.github/workflows/fuzz.yml`; this branch touches neither `elpsutil/` nor `.github/`. Merges clean. One mechanical fixup after: the new target reaches `env.Runtime.Registry.Packages`, unexported here → `Registry.Package()` / `PackageNames()`. |
| #358 observable macro-expansion panics + fuzz analysis/lint | **Independent** | `analysis.PanicReporter` / `ExpansionPanics` are absent here; this branch's only `analysis/expander.go` change is a one-line `v.Quoted` → `v.IsQuoted()`. Merges clean. Fixups after: `expr.Quoted` → `IsQuoted()`, and `SourceOf(x).Source` → `astutil.SourceLoc(x)` in the two new fuzz files. Note #358 now also carries the **sixth fuzz shard** (`5ccc6d4`). |
| #359 `Config.MacroExpander` in `AnalyzeFile`; `astutil` nil-deref | **Conflicting** | `astutil/walk.go` conflicts on `SourceOf`: this branch rewrote the exact predicate (`v.Source != nil && v.Source.Line > 0` → the `v.Source()` accessor) while #359 inserts a `v == nil` guard at the same line. Neither fix exists here — no `ConfigForFile` (#353), no nil guards (#354). |
| #360 restore the caller's package across `elpsutil` loads | **Independent** | `elpsutil/` is untouched by this branch; merges clean. Fixup after: the new test reads `Registry.Packages` and `pkg.Externals` → `Package()` / `PackageNames()` / `Externals()`. |
| #361 errors instead of panics on embedder mistakes in `elpsutil` | **Independent** | Same; merges clean. Note this one needs a **non-test** fixup: `elpsutil/validate.go` reads `pkg.Symbols` → `Package.Symbol()` / `SymbolNames()`. |

Nothing in the queue is fully subsumed, so nothing there should be closed on account of this PR.

### Merge order — what has happened, and what is left

1. ~~**#384 first.**~~ **Done.** Merged as `b1ab958`; this branch merged `main` forward at `ee25bbd` to pick it up, so the confidentiality guard now runs over this branch's ~12k added lines.
2. ~~**#356–#361 next, also on `main`.**~~ **Partly done.** #356 (`663c0f3`) and #357 (`f5d3d9e`) merged, and #391 (`050ade5`, cyclic-value process kill) merged alongside them. #358–#361 are still open and still `main`-based; none of them needs the sealed API to merge to `main`.
3. **The forward-merge of `main` into `claude/exp-seal-tooling` is still owed.** `ee25bbd` merged `main` only as far as `b1ab958`, which predates #356/#357/#391 — so the resolutions listed here have not been performed yet, and they are now against `main` at `663c0f3`: `astutil/walk.go` for #359 when it lands (keep *both* — the nil guard and the accessor form); `lisp/lisplib/fuzz_test.go` for #356 (keep this branch's fingerprint oracle, port fuzzfp's LNative check, `FuzzApplyThinStdlib` and the fuzz.yml change on top, drop fuzzfp's `Source` half since the field and the shared singleton are gone); `elpsutil/` for #357 plus the mechanical accessor swaps listed above — the same migration substrate already priced at 7 read-site swaps.
4. **#374** once that forward-merge is green and re-benchstatted.
5. **Stacked, post-release.** The stack is a **linear chain**, not four siblings on this branch: **#374 → #377 → #388 → #389 → #387 → #386 → #385**. Only #377 has `base = claude/exp-seal-tooling`; each of the others is based on its predecessor's head.

   | PR | base branch | subject |
   |---|---|---|
   | #377 | `claude/exp-seal-tooling` | elpsvet escape-aliasing rule + singleton sealing (#375, #376) |
   | #388 | `claude/exp-vet-esc` | `copy` builtin + checked-mode CoW counter (#378) |
   | #389 | `claude/exp-copy-builtin` | two-env and container-family corruption fuzz targets |
   | #387 | `claude/exp-corruption-fuzz` | seal-enabled optimizations (#379) |
   | #386 | `claude/exp-seal-opts` | template-environment `Fork` (#380) |
   | #385 | `claude/exp-env-fork` | per-callsite macro-expansion caching (#381) |

   They must merge bottom-up in that order; rebasing them onto whatever #374 becomes has to preserve the chain.

## Test plan

- [x] `go test ./...` and `go test -tags elpscheck ./...` green at every layer and at HEAD
- [x] golangci-lint 0 issues; elpsvet exit 0 (binary-clean, all suppressions justified)
- [x] Corruption kill tests + red-proofs (#378-class, #369 both mechanisms, #370 watchdog) — re-verified at assembled head
- [x] Interleaved benchstat vs main and vs minus-seal control (house methodology, n≥10); untagged binaries symbol-verified seal-free
- [x] Full downstream end-to-end: substrate + 71k-line production phylum suite PASS (both build modes)
- [x] Three independent adversarial review passes: READY

---


sam-at-luther:
CI follow-up at `2110ac5` — two commits pushed to turn the two red required checks green. One of them is a **functional change** that post-dates the "READY at dee7317" review status, so flagging it here explicitly.

## 1. Lint & Test (`93938777601`): formatting gate

`./elps fmt -l ./...` flagged `lisp/lisplib/libelpspath/libelpspath_test.lisp` (trailing blank line). Fixed by `elps fmt -w` (`93bfa08`, whitespace-only).

## 2. Benchmark Comparison (`93938788594`): the gate was RIGHT — fixed the regression instead of the gate

Gate report: exactly one at-or-above-gate row — **libjson `Package/$load` B/op +9.32%** (949.1Ki → 1037.5Ki, alloc gate 5%; allocs/op +4.56% just under). This was *not* the known libelpspath load-time delta: alloc-space profiling attributes the entire increase to the issue-#363 **formals boundary copy** — `AddBuiltins`/`AddMacros`/`AddSpecialOps` deep-copied every definition's formals per environment, ~90KiB and >1000 allocs per `LoadLibrary` env.

**Fix (`2110ac5`, functional):** replace the eager per-env copy with the seal design's own lazy pattern. The shared definition tables' formals are sealed once at construction (`sealDefaultFormals` in `lisp/builtins.go`, the `libutil` constructors, `libgolang`'s init, `RegisterDefault*`), and registration (`registrationFormals`, formerly `copyFormals`) aliases the sealed list into each env — the identical treatment lambdas already get (their formals are sealed parser output aliased into every closure), covered by the same CoW guards and, in checked builds, the fingerprint verifier. Third-party `LBuiltinDef`s with unsealed formals still get the defensive deep copy. `TestNoCrossEnvironmentLValSharing` now asserts the #363 property in sealed form: cross-env shared pointers must be singletons or sealed (anti-vacuity floor: ≥100 sealed shared, so the walk provably still reaches the formals graph). Doc: `docs/sealed-ast.md` §2.6.

**Measured (interleaved vs origin/main, libjson `Package/$load`):** B/op +9.33% → **+0.70%**, allocs/op +4.57% → **+0.44%** — remainder is the honest libelpspath package registration at library load. `BenchmarkEnvInit` drops ~74KiB/op and ~1050 allocs/op vs dee7317.

Local gates at `2110ac5`: `go test ./...` and `-tags elpscheck` green, `make race` green, golangci-lint 0 issues (both passes), `elpsvet -test=false` exit 0, `elps fmt -l` clean, `elps lint`/`elps doc -m` green, `_examples` green.

**Maintainer note:** the descendant substrate PR #382 pins `dee7317` by pseudo-version; since `2110ac5` is a functional change to registration, that pin needs a bump to the new head once this is accepted (not done here).

---

sam-at-luther:
Heads-up for whoever picks this branch back up: #384 removes a customer name from test comments across `analysis/` and `lint/`, and adds a CI guard (`scripts/confidentiality-guard.sh`, wired into the `lint-and-test` job) that fails the build on any case-insensitive word-boundary match of that term in the tracked tree.

This branch predates that scrub and still carries the term in its test comments, so **once #384 merges, this PR will need `main` merged forward** before CI can go green — the guard will fail on the pre-scrub comments otherwise.

I have deliberately **not** done that merge here, since this PR has an active owner. Flagging it so the red isn't a surprise.

---

sam-at-luther:
## Adversarial review — the #382 privatization layer (`5819cdc..HEAD`)

**Verdict: READY**, after one confirmed fix (`4cab7aa`, below). Every other candidate I chased resolved to no-drift, and I verified rather than reasoned wherever a repro was possible.

### Findings

| # | Area | Status | Outcome |
|---|---|---|---|
| 1 | `funData` fields reachable by reflection through `LVal.Native` | **CONFIRMED** | Fixed in `4cab7aa` with a regression guard |
| 2 | Guard vacuity — `TestLValFieldSeal` / `TestLEnvFieldSeal` | **Verified non-vacuous** | 4 mutation probes, all red |
| 3 | `internal/` bridges as a seal bypass | **Not a hole** | All 7 packages compiler-blocked from an external module |
| 4 | Format-preserving path byte-identity | **Confirmed identical** | 40 inputs, 0 byte diffs vs `5819cdc` |
| 5 | `Bindings()` — snapshot vs live, mutation during iteration | **No drift** | Live-iterates the same map the exported field exposed |
| 6 | `Source()` copies — identity reliance | **No drift** | No caller compares or mutates through it |
| 7 | `MacroExpansion()` snapshot — debugger correctness | **No drift** | One condition tightened; the state it drops is unconstructible |
| 8 | `funraw.Env` lifetime for in-repo tooling | **No drift** | Same live pointer the old `fun.Env()` returned |
| 9 | Debugger / profiler / LSP consumer swaps | **No breakage** | Suites green plain, `-race`, and `-tags elpscheck` |

---

### 1. CONFIRMED — the captured environment was still reachable, by reflection (fixed)

Unexporting a *type* is not the same as sealing it. `LFun` values carry their `funData` in `LVal.Native`, which is exported and deliberately stays exported. `reflect` **can** read an *exported* field of an unexported struct reached that way and hand back a fully usable value — `reflect.Value.Interface` only refuses values obtained from *unexported* fields. So while `funData.Env` was exported, this compiled in any embedder, with no `unsafe`:

```go
fd := reflect.ValueOf(fn.Native).Elem()
captured := fd.FieldByName("Env").Interface().(*lisp.LEnv)
captured.Put(lisp.Symbol("n"), lisp.Int(9999))
```

Run against this branch from a separate module: a closure built as `(mk 41)` returned **9999** afterwards. That is exactly the captured-environment aliasing channel `internal/funraw`'s doc comment says an embedder "cannot reach at all", still open through the one privatized field that was an *exported field of an unexported struct* rather than an unexported field.

I probed each privatized field individually. Every other one is immune, because each is an unexported **field** — `LVal.source`/`meta`/`macroExpansion`/`quoted`/`spliced`/`sealed`, `LEnv.scope`/`funName`/`parent`/`loc`, and `MapData`'s aliased `mapBacking` embed all make `reflect.Value.Interface` panic. `funData`'s four were the sole asymmetry.

**Fixed in `4cab7aa`**: the payload's fields are unexported (`builtin`/`env`/`fid`/`pkg`). No behaviour change — they were only ever read in-package (the `FID`/`Package`/`Builtin` accessors, `funEnv`, `Copy`, `detach`'s builtin check) and by in-repo tooling through `internal/funraw`, which is injected from package `lisp` and unaffected. The repro now reports "no Env field reachable". `TestFunDataPayloadFieldSeal` pins it, asserting the payload behind `Native` exports no fields at all and that the mediated identity reads still work; red-proofed by re-exporting `Env` (guard fires by name, reverted clean).

### 2. The two field-seal guards are not vacuous

Both branches of both guards were mutation-tested, not read:

| Probe | Result |
|---|---|
| Add an exported field to `LVal` | red — names `ProbeMeta` |
| Add an exported field to `LEnv` | red — names `ProbeScope` |
| Rename away `(*LVal).MacroExpansion` | red — "accessor missing" |
| Rename away `(*LEnv).Bindings` | red — "accessor missing" |

The allowlist-removal branch was already red-proofed by the author (`8b05c3b`). Worth noting the accessor-presence half only fires once the package still *compiles* without the accessor — deleting one outright fails the build first, which is a stronger signal, so the guard's real job is catching a *rename* that in-repo callers followed.

### 3. The internal bridges do not defeat the seal

Compiled a throwaway external module against this tree and tried to import each bridge directly:

```
BLOCKED: internal/funraw       BLOCKED: internal/funraw/hook
BLOCKED: internal/fmtraw       BLOCKED: internal/fmtraw/hook
BLOCKED: internal/fmtmeta      BLOCKED: internal/astraw
BLOCKED: internal/macroexp
```

All seven refused by the compiler's `internal/` rule — including the `hook` slots, which matters, since those are untyped `any` vars an external package could otherwise *overwrite* to swap the injected accessor. And no exported symbol in `lisp` returns `*funData` or `*fmtmeta.Meta`. The one remaining route was finding #1, now closed.

On the write side: `fmtraw.SetMeta` has no seal guard, which is correct but load-bearing on an invariant — every `EnsureMeta` call site in `rdparser` is `preserveFormat`-gated, and format parses are never sealed (`TestFormatParseTrailingCommentNotSealedWrite` pins that). Worth keeping those two facts adjacent in the doc, since the guard's absence is only safe while both hold.

### 4. Format-preserving output is byte-identical

Built `elps` at `5819cdc` and at HEAD and compared `elps fmt` output byte-for-byte: **34 repo `.lisp` files + 6 hand-built stress cases** (leading/inner/trailing comments, comment-only gaps, quasiquote with unquote-splicing, bracket-let, strings containing `;`, prefix forms) — 0 differences in stdout, stderr, or exit status. The `fmtmeta`/`fmtraw` move is a pure relocation.

### 5–8. Candidates that resolved to no drift

- **`Bindings()` live-iterates**, it does not snapshot: the returned `iter.Seq2` ranges `env.scope` at iteration time, so a debugger pane sees current values, matching the old exported map. Mutation-during-iteration keeps Go's map semantics exactly (the write happens inside `yield`, inside the range) and concurrent writes still fatal — identical hazard, not a new one. I also checked the one way a stored `Seq2` could diverge from a stored map reference: a live env's `scope` is never *replaced* after construction (the only `cp.scope = make(...)` is on the copy in `LEnv.Copy`), so the two are equivalent even when held across time.
- **`Source()` copies** are strictly safer and no caller relies on identity; `InspectFunction`'s `FunctionInfo.Source` (`8b05c3b`) has no in-repo production consumer at all.
- **`MacroExpansion()`**: `engine.go`'s condition tightened from `expr.MacroExpansion != nil` to "info **and** context non-nil". Reachable only if a `macroExpansionInfo` could exist with a nil context — it cannot: `stampMacroExpansion` constructs one only under `if ctx != nil`, and the `macroexp.Attach` test bridge always supplies one. Equivalent in practice. This is also why `TestInspectMacroExpansion_NilContext` was deleted rather than migrated; the state is now unconstructible even in-repo. Defensible, but that is a deliberate coverage retirement rather than a mechanical port, and it is the one place in the layer where a test went away.
- **`funraw.Env`** returns the same live `*LEnv` the old `fun.Env()` did — no lifetime change for in-repo tooling.
- **Consumer swaps** (`minifier` ×14, `libschema`, `libjson`, `lsp`, `repl`, debugger, profiler) are mechanical, and several are *safer*: `IsQuoted()` is nil-receiver-safe where `.Quoted` would have nil-dereferenced. `isValidator`'s `v.Builtin()` and `applyConstraint`'s `constraint.Builtin()(env, input)` both sit behind an `LFun` type check, so the accessor's panic-on-non-function is unreachable. Dropped `funData == nil` early returns in the profiler and debugger were already dead code — the old `FunData()` type-asserted unchecked too.
- I also checked whether the DAP `setVariable` path could write through a `MacroExpansionMeta.Args` node (the snapshot shares the original nodes, which are typically sealed). It cannot: `onSetVariable` handles only package and local scope; the macro scope falls through to "not supported".

### Gates on the fix

`go test ./...` and `-tags elpscheck ./...` pass; `go test -race ./lisp/ ./lisp/x/... ./lsp/` passes; golangci-lint 0 issues on both passes; `elpsvet -test=false ./...` exit 0; `elps fmt -l` clean.

**READY** for the privatization layer. The seal is now uniform: every field #382 privatized is unreachable from outside the module without `unsafe`, and both guards fail if that changes.

---

sam-at-luther:
Superseded by #517, the same core rebuilt as one clean commit on current main (post-v1.50.0, #512, #513, #515), with the rebase resolutions documented there and in-tree. The full replacement stack is #517 → #518 → #519. This branch is preserved for reference.

---

PR #377, opened 2026-08-11
elpsvet escape-aliasing rule + singleton sealing (stacked on #374)

### LINKED ISSUE #375: elpsvet: escape-aliasing rule — flag runtime-owned pointers stored uncopied into escaping values
The freshness rule checks the mutation direction (writes to values that are not fresh). The ErrorCondition/ErrorConditionf bug found in review was the mirror image: a shared runtime-owned pointer (`env.Loc`) stored uncopied into a freshly constructed error that escapes the function. The write target was fresh, so the freshness rule was structurally blind to it.

New rule: taint runtime-owned pointer state — `env.Loc`, `v.source`, and other `*token.Location` reads off runtime/LVal state — and flag any store of a tainted pointer into a field of a value that escapes (returned, stored into an LVal, put into a registry), unless it passes through `copyLocation`/an explicit copy. Escape hatch remains `//elps:mutates`-style annotation for the deliberate in-runtime aliases (`TaggedValue`, `Lambda`, parser token locations).

Acceptance:
- Retro-catches pre-fix `ErrorCondition`/`ErrorConditionf` (fixed in ac0a326) and the original ErrorAssociate aliasing (fixed in d922290), verbatim diagnostics in the report.
- False-positive census over the tree with per-flag dispositions; deliberate aliases annotated with justification.
- Analyzer unit tests; vet wall time stays in the ~0.6s class.

Related: #371 (the mutation-direction alias gap, closed), the review findings on PR #374.

### LINKED ISSUE #376: Seal the Nil/True/False singletons and register them as permanent inspector roots
The seal machinery currently skips the process-wide singletons. Closing that gives them the same layered protection as program AST:

1. Mark `singletonNil`/`singletonTrue`/`singletonFalse` sealed at init: `IsSealed()` then communicates the do-not-mutate contract to embedders, and the copy-on-write mutation-site guards refuse/copy container ops on them.
2. Register them as permanent roots with the checked-mode AST inspector so their bytes are re-verified continuously at every load/teardown in CI, not only at snapshot points.
3. Document the layering: freshness rule flags writes to constructor returns statically (#333 shape); elpscheck snapshot catches value drift; the #334 watchdog catches any unsynchronized write under -race including same-value writes; the untagged-production same-value write remains -race-only by design (mprotect approaches evaluated and rejected in #334).

Acceptance: red-proof that a deliberate singleton field write is caught by the inspector under the tag (and remains caught by the watchdog under -race); untagged builds unaffected (no new symbols, benchstat flat); all suites green both tags.

Related: #333/#334, #372, the shared-formals allowlist fix (a2341f1).

sam-at-luther:
## Summary

Stacked on #374; chain: #374 → **#377** → #388 → #389 → #387 → #386 → #385. Base branch is `claude/exp-seal-tooling` (#374's head); #388 is stacked directly on this branch.

Closes #375 and closes #376 — the two gaps the #374 review surfaced:

1. **`elpsescape` vet rule (#375):** the freshness/alias rules check the mutation direction; this checks the mirror image — runtime-owned `*token.Location` pointers (`env.Loc`, `v.source`, producer-fixup token locations) stored uncopied into escaping values. Retro-catches all three historical bugs verbatim (pre-fix `ErrorCondition`, `ErrorConditionf`, `ErrorAssociate`). Census: 13 flags — 3 fixed with `copyLocation` (including a real latent exposure: `InspectFunction` handed embedders a mutable pointer into a possibly-sealed node's location), 9 annotated `//elps:aliases` with individual justification, 1 false positive retired by proof rather than annotation (below). Vet wall time 0.58s.
2. **Singleton sealing + permanent inspector roots (#376):** `Nil`/`True`/`False` are now sealed (IsSealed() communicates the contract; CoW guards cover container ops — edge shapes tested) and registered as permanent inspector roots re-verified at every load and teardown under `elpscheck`. Red-proofs: a deliberate singleton write is caught by the inspector (fingerprints printed) and by the watchdog under `-race` even for a same-value write. No hot-path changes were needed.

## Cross-package freshness facts (`cmd/elpsvet/locfact.go`)

The rule's first cut treated *every* `*token.Location`-returning method on a receiver the caller did not construct as a taint source. That is right for `rdparser.Parser.Location`, which hands out the parser's own token pointer, and wrong for `lisp.LEnv.Source`, which returns `copyLocation(env.loc)` precisely so callers may keep the result. Annotating the call site was the wrong answer: it pushes the cost onto every embedder using the sanctioned API, and spends an audited annotation on something that is not a claim but a provable fact.

The analyzer now computes, per function whose sole result is a `*token.Location`, the summary "every value this can return is freshly allocated", and exports it as a `go/analysis` fact. Facts travel the import graph, so `LEnv.Source` is clean in `lsp/`, in `mcpserver/` and in any embedder with nothing written at the call site, while `LocRef`-shaped by-reference accessors of identical signature stay flagged. A result qualifies when it is `nil`, a `&token.Location{...}` literal or `new(token.Location)`, `&local` for a value-typed `token.Location` variable or parameter (the deref-copy idiom), or a call to something already proven. Anything else leaves the summary false and the call site conservatively flagged, so the fact can only ever retire a proven false positive.

Two fixpoints keep it honest: functions are proven by least fixpoint from nothing, so a recursion cycle cannot bootstrap itself, and location locals are resolved flow-**insensitively** — every assignment must be fresh — because a source-order walk would call a loop-carried local fresh when a `return loc` textually precedes the `loc = tok.Source` that feeds it on the next iteration.

Consequences: the `copyLocation` name check is gone (the cleanser is now recognised by behaviour, not spelling); the `parser/token/scanner.go` annotation is retired, since `LocStart` mints a fresh `Location` per token and the fact sees it. Annotation count 10 → 9, and every survivor was re-checked by removing it and confirming the diagnostic returns.

`analysistest` checks facts as strictly as diagnostics, so a declaration with **no** `// want Name:"freshLocation"` asserts the summary was not recorded. Pinned in both directions: `LEnv.Source` clean vs `LEnv.LocRef` flagged, the one-leaking-return accessor, the deref-copy accessor, the transitive wrapper, the parameter passthrough, the recursion cycle, the bare-return named result, the loop-rebound local, and `(*LVal).Source` / `ErrorVal.Source`, whose by-value results need no fact at all.

## Performance

Untagged binary: `go tool nm` diff vs base = **0 lines**. Benchstat interleaved n=10: allocs/op and B/op identical on all 19 benchmarks; no significant sec/op row.

## Status

Draft: this branch has **not** yet had the independent adversarial review pass that #374's layers received. Evidence comments: [#375](https://github.com/luthersystems/elps/issues/375#issuecomment-5259621792), [#376](https://github.com/luthersystems/elps/issues/376#issuecomment-5259627078).

## Test plan

- [x] `go test ./...` plain and `-tags elpscheck`; `-race -tags elpscheck ./lisp/` clean
- [x] golangci-lint 0 issues (both passes); elpsvet exit 0
- [x] Retro-catch fixtures + live-tree reverted-fix diagnostics
- [x] Red-proofs for both inspector hooks with restore-and-clean verification
- [x] `analysistest` strictness verified by deleting a `// want` (diagnostic and fact) and confirming failure, then restoring
- [ ] Independent adversarial review

---

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,9 @@
-elps
+# Anchored: only the built binary at the repo root.  A bare `elps`
+# pattern matches ANY path component named elps -- including
+# cmd/elpsvet/testdata/src/github.com/luthersystems/elps/..., which
+# silently kept new analysistest stub packages out of the repository
+# (they existed locally, so tests passed locally and failed in CI).
+/elps
 .firecrawl/
 .claude/settings.local.json
 .claude/worktrees/
--- a/cmd/elpsvet/alias.go
+++ b/cmd/elpsvet/alias.go
@@ -103,7 +103,7 @@ func newAliasTracker(pass *analysis.Pass, fresh map[types.Object]bool, ann map[i
 
 func (t *aliasTracker) report(pos token.Pos, op string) {
 	line := t.pass.Fset.Position(pos).Line
-	if t.ann[line] || t.ann[line-1] {
+	if t.ann[line] {
 		return
 	}
 	t.pass.Reportf(pos,
--- a/cmd/elpsvet/escape.go
+++ b/cmd/elpsvet/escape.go
@@ -36,11 +36,18 @@
 //   - a field write rooted at a package-level variable — the
 //     registry-reachable shape.
 //
-// CLEAN: a location routed through copyLocation (or any func(*Location)
-// *Location named copyLocation), a &token.Location{...} literal, and
-// &local where local is a value-typed variable (the parser's deref-copy
-// idiom: src := *loc; v.SetSource(&src)) — all of these give the escaping
-// value its own memory.
+// CLEAN: a &token.Location{...} literal, &local where local is a
+// value-typed variable (the parser's deref-copy idiom: src := *loc;
+// v.SetSource(&src)), and — the cross-package part — the result of ANY
+// function or method PROVEN to allocate the location it returns.  That
+// proof is the freshLocation fact (locfact.go): the analyzer summarises
+// each location-returning function in the package it is analysing and
+// exports the summary along the import graph, so copyLocation,
+// lisp.LEnv.Source (which returns copyLocation(env.loc)) and
+// token.Scanner.LocStart (which mints a &Location{...} per token) are clean
+// at every call site, in every package, WITHOUT an annotation.  An embedder
+// using the sanctioned copying accessor must never have to annotate; only a
+// deliberate alias should.
 //
 // Out of scope (documented limitations, consistent with the freshness
 // rule's intraprocedural approximation): function parameters are not taint
@@ -49,7 +56,10 @@
 // (stampMacroExpansion receives callSite by parameter); whole-struct copies
 // (*cp = *v) alias the source field implicitly and remain the freshness
 // rule's and the seal design's concern; index reads (locs[i]) are not taint
-// sources.
+// sources.  A callee with NO freshness fact — an un-analysed package, an
+// interface method, a function whose body has one leaking return — keeps
+// the conservative treatment, so the fact can only ever retire a proven
+// false positive, never open a hole.
 //
 // Escape hatch: //elps:aliases trailing the flagged line, standing alone on
 // the line directly above it, or in the enclosing function's doc comment.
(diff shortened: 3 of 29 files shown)


sam-at-luther:
Superseded by #518 (collapsed with #388 and #389 into the verification layer of the rebuilt stack #517 → #518 → #519). Branch preserved for reference.

---

Issue #379, opened 2026-08-12
Seal-enabled optimizations: defensive-copy audit, bench-harness sharing, interning, path-compile cache


sam-at-luther:
The formals seal-and-share change (PR #374, docs/sealed-ast.md §2.6) demonstrated the general pattern: the seal converts "copy defensively because sharing is unsafe" into "prove immutable, then share" (env-init −18% time / −20% allocs). This tracks the systematic follow-through. None of these are breaking; all can land in ordinary releases after the current PR set.

1. **Defensive-copy audit**: sweep every `Copy()` call site (elps and downstream) asking "does this copy exist only because sharing used to be unsafe?" The formals copy was found by a CI gate; find the rest on purpose. Each hit gets the formals treatment: seal at construction, share, cover with the property test.
2. **`elpstest.RunBenchmark`**: the per-iteration AST copy (added to avoid cross-runtime sharing) is unnecessary for sealed programs — share the sealed AST across iterations/runtimes. Faster, lower-noise benches.
3. **Symbol interning**: an intern table of sealed symbols for runtime-constructed symbols; safe by construction now. Modest allocation win; profile before committing.
4. **Compiled-path cache in libelpspath v1**: `get-path` and friends regex-parse the path string per call; a small LRU keyed by path string (strings immutable; independent of the seal but adjacent) removes the per-call parse. The v2 ops already avoid parsing entirely.
5. **Candidate (profile-first): share whole builtin function objects, not just formals.** Registration still allocates an LFun header per builtin per env, and writes the docstring into `fn.Cells[1]` per env — moving the doc into the shared sealed definition would let the entire function object be shared, taking another large bite out of env-init. Natural completion of the formals change.
6. **Candidate (profile-first): libjson allocation profile.** The lisplib benches show ~1.27MiB / 20k allocs per op on dump/load shapes and JSON dominates real transaction cost; buffer reuse / streaming paths are the classic wins. Independent of the seal; listed because the same perf effort should look here first for request-path gains.

House methodology applies throughout: interleaved benchstat, red-proofs where a guard is removed, no gate weakening.

Related: #374 (the pattern's first instance), #372 (verification layers that make sharing provable).

sam-at-luther:
[...]

PR #385, opened 2026-08-12
Per-callsite macro-expansion caching keyed by sealed AST nodes (POC, default off)

### LINKED ISSUE #381: Design: per-callsite macro-expansion caching keyed by sealed AST nodes
Sealed AST nodes are valid cache keys — stable pointer identity plus immutable content — which makes memoizing macro expansions per call site coherent for the first time. Macro-heavy request paths (routing wrappers, handler-bind) currently re-expand on every call; caching the expansion per callsite attacks request-path latency directly.

Hazards the design must resolve before any implementation:
1. **Impure macros**: expansion may depend on runtime state; caching changes semantics. Options: opt-in cacheability, purity analysis, or conservative whitelisting of kernel macros only.
2. **Gensym uniqueness**: a cached expansion must not reuse gensyms across calls where uniqueness matters (or must prove the expansion is evaluated fresh each time so gensym identity inside the template is safe — the expansion is code, evaluated per call; analyze whether reuse of the expansion *tree* is observationally different from re-expansion, including through quote/closure capture of expansion nodes).
3. **Invalidation**: macro redefinition must invalidate cached expansions (epoch counter on the macro binding).
4. **Cross-env sharing**: cached expansions contain sealed input nodes; the expansion tree itself must be sealed before sharing across envs, or the cache kept per-env.

Deliverables for a POC: a whitelist-scoped prototype (kernel macros only), warm-transaction benchmark on a macro-heavy endpoint, red-proof that redefinition invalidates, semantics-equivalence tests including gensym-sensitive shapes.

API note: purely internal to the evaluator; no API change; does not block the current release train. Expensed to the request path (that is the point).

sam-at-luther:
&gt; **Stacked on #386; chain: #374 → #377 → #388 → #389 → #387 → #386 → **#385**. Post-release.** Base branch is `claude/exp-env-fork`, not `main` — review the [comparison against that base](https://github.com/luthersystems/elps/compare/claude/exp-env-fork...claude/exp-macro-cache). This PR sits at the TIP of the chain deliberately: it is the most speculative branch in the stack, and the tip is the only position from which it blocks nothing beneath it. This is a POC/decision packet, not a merge-now change: the cache ships **default-off** and should land after the sealed-AST stack releases.

Closes #381. Full decision packet with the per-macro distributions: [#381 comment](https://github.com/luthersystems/elps/issues/381#issuecomment-5262942075).

Per-callsite memoization of macro expansions, keyed by sealed callsite node identity. Sealed AST nodes have stable pointer identity and immutable content, which is exactly what a callsite cache key needs. Macro-heavy request paths re-expand the same callsites on every transaction; this attacks that cost.

Default `MacroCacheOff` — nothing changes unless a host opts in via `SetMacroCacheMode` or `ELPS_MACRO_CACHE=runtime|shared` (`ELPS_MACRO_CACHE_CAP` bounds either table).

## Decision packet

**Speed** — warm-transaction benchmark against a production-scale phylum, n=10 interleaved, benchstat:

| endpoint | off | shared | runtime |
|---|---|---|---|
| healthcheck | 623.6µs | ~ (p=0.28) | ~ (p=0.12) |
| read-light | 3.774ms | ~ (p=0.22) | ~ (p=0.22) |
| read-heavy | 2.640ms | **−7.08%** (p=0.000) | **−5.81%** (p=0.000) |
| write-heavy | 11.89ms | **−5.80%** (p=0.019) | **−9.04%** (p=0.002) |
| **geomean** | 2.932ms | **−4.30%** | **−4.82%** |

Allocations are the deterministic column: B/op geomean **−5.44%**, allocs/op **−6.37%**, p=0.000 everywhere, bit-identical between the two modes (same expansion work saved).

**Memory** — warm-pool topology, where the cache is useful:

| config | entries | cache bytes | hit rate |
|---|---:|---:|---:|
| shared, unbounded | 154 | 110 KB | 95.9% |
| shared, LRU 1024 / 256 | 154 | 110 KB | 95.9% (0 evictions) |
| shared, LRU 64 | 64 | 46 KB | 51.6% (thrash) |
| per-runtime | 154 | 110 KB × envs | 95.9% |

Hundreds of entries at ~730 B/expansion. **But** in a parse-churn topology the sealed callsite keys pin entire dead parse trees: the same phylum's unit-test suite under unbounded shared mode reached 15.6 GB RSS and was OOM-killed. With `ELPS_MACRO_CACHE_CAP=8192` the same suite serves the same reuse (81,584) in 27 MB and passes.

**Per-runtime mode's bound is the runtime's LIFETIME, not immunity.** An earlier revision of this body claimed per-runtime mode was "structurally immune" to that failure mode. It is not, and the claim is withdrawn: a long-lived runtime that keeps evaluating *fresh parses* (a REPL, a host that hot-reloads programs) pins dead parse trees through its callsite keys exactly like an uncapped shared table. What makes per-runtime mode safe in the warm-pool topology is that program identity there is stable, not that the map was bounded. `SetMacroCacheCap` / `ELPS_MACRO_CACHE_CAP` now bound **both** tables. The per-runtime bound is a wholesale **drop** rather than an LRU — the table is an unsynchronized hot-path map, and in the only topology where it grows without limit every old key is a dead parse, so recency carries no information worth paying for. Correctness is unaffected: a dropped callsite simply re-expands.

**Where the reuse is** — 786,432 dispatches / 704,538 callsites / 28 macros on the real corpus. Four user template macros (`when`/`unless`/`default`-class plus one domain wrapper) carry **99.05%** of all repeat opportunity. Kernel `defun` is 80.6% of dispatches but has **40** repeats across 633,590 callsites — admitting it would cost 0.46 GB (at the warm topology's 730 B/expansion) to 2.1 GB (at the 3,281 B/entry that run actually measured) to buy 0.05% more reuse, before counting the dead parse trees each key would pin. The whitelist boundary was argued from semantics (closure-embedding expansions cannot be reused); the measurement shows it is also the dominant memory decision. Implementation serves **99.62% of all achievable reuse**.

**The main design conclusion:** this issue offered "conservative whitelisting of kernel macros only" as one way to resolve the impurity hazard, and the measurement rules it out. The admissible *native* macros account for **695 repeat-dispatches — 0.85%** of the available reuse. The other **99.15%** lives in embedder-defined `defmacro` template macros, which a kernel-only whitelist cannot reach by construction. The user-macro purity prover is not a refinement of the whitelist approach; it is the only tier that captures anything.

## Design

Full write-up in [`docs/design/macro-expansion-cache.md`](https://github.com/luthersystems/elps/blob/claude/exp-macro-cache/docs/design/macro-expansion-cache.md).

- **Admission**, three tiers: hand-audited native macros (`get-default`, `trace`, `curry-function`, `testing:` family); user `defmacro` macros certified by a conservative syntactic prover (`macrocache_purity.go`) as pure quasiquote template instantiation; everything else bypasses. `defun`/`defmacro`/`deftype` are excluded by design — their expansions embed a closure allocated at expansion time, so reuse would share one closure's environment.
- **Invalidation** by exact macro identity, no epoch counter: package+FID **plus the registration id** for natives, sealed formals-node pointer for user macros, validated on every hit against the macro the callsite resolves to now. Redefinition necessarily mints fresh identity, so a stale entry cannot be served. The registration id matters because the FID is derived from the registration *name*: two environments registering different implementations under one qualified name were otherwise indistinguishable to the process-shared table, and the second was served the first's expansion.
- **Gensym reuse** is safe for admitted macros because their gensyms only appear in binding positions; the prover rejects gensym-under-quote, the one shape where identity is observable.
- **Expansion trees are sealed before publication** to the shared table, mirroring the shared-formals precedent.
- **Scope**: per-`Runtime` map (unsynchronized, dies with the runtime) or process-shared `sync.Map` with optional LRU. Both are capped by `SetMacroCacheCap`.

### Soundness: what the prover resolves, and what remains an assumption

An earlier revision of this body said the prover matched its structural operators *by name*, "assuming they resolve to the kernel bindings", and called that assumption the reason the branch was not ready. **That assumption is gone.** Every one of those shadows was demonstrated to return a *wrong answer* under caching — not merely a misclassification — so the admission test is now split in two:

- **The syntactic verdict** (`macroPurity.pure`) is a function of the macro's sealed, immutable formals and body nodes and *nothing else*. That, and only that, is memoized process-wide on the formals node.
- **The name-resolution obligations** (`defRefs` / `callRefs`) are the operator spellings the proof interpreted, each of which must resolve to the kernel binding before any expansion is cached. They are environment dependent, so they are **re-checked on every dispatch** and never memoized — which is also what makes a `(set 'if ...)` applied *after* the macro was proven take effect immediately.

Which obligation is checked where mirrors where the code actually runs. `if`, `progn`, `let`, `let*`, `quasiquote` and `gensym` from the **body** resolve in the macro's **defining** environment — and specifically fall through to the macro's *own package*, because `funCall` switches the runtime into the function's package for the duration of a call, so a shadow installed there is invisible from the caller. The template's **binder syntax** (`let`/`let*`/`labels`/`flet`/`lambda`, the one place the prover interprets output code) resolves at the **callsite**, and only when the macro actually mints a gensym, the sole case where that reading matters. Package-qualified spellings (`lisp:if`) name their package explicitly and are resolved as written. A spelling that does not resolve to the kernel binding simply makes the macro uncacheable; refusing to cache is always sound.

`gensym` deserves a separate note because it was the one expression the grammar admitted *without proving it from its own structure*. The review asked for the binding **value** to be purity-checked. What shipped is stronger and not equivalent: the head must resolve to the **kernel gensym builtin** in the defining environment and the call must take no arguments. `(gensym)` is not pure — that is the point of it — so a value-purity check could only ever reject it and delete the whole gensym tier; an *identity* check keeps the tier while refusing every substitute, including a deterministic user replacement that a purity check would happily admit (`TestMacroCacheGensymShadowedByPureFunctionNotAdmitted`).

**What genuinely remains, stated precisely:**

1. **Template content is not covered, and does not need to be.** `(quasiquote (if ...))` is output code, inert to the prover and reused verbatim, so whatever `if` means at the callsite it means identically with and without the cache. The exception — binder syntax — is exactly what `callRefs` covers.
2. **Three readings carry no obligation, each on a ground now asserted rather than argued.** `quote` in template position can only *raise* the quote depth, so mis-reading a shadowed `quote` can only reject a macro, never admit one. `unquote`/`unquote-splicing` are not bindings — quasiquote matches them by name on an `LSymbol` head and ignores any binding of them, which `TestMacroCacheUnquoteIsNotABinding` now demonstrates. `true`/`false`, the only free symbol reads the grammar admits, are kernel constants that `set`, `defun`, `defmacro`, `let`, `let*`, `labels` and `flet` all refuse to rebind — every route asserted by `TestMacroCacheTrueFalseAreKernelConstants`.
3. **One reading is known to be wrong and is harmless.** Defining-environment obligations resolve in `funData.env`, while the body evaluates one frame *below* it, in a call environment carrying the macro's own **formals**. A macro whose formal is named `if` (or `let*`, or `gensym`) is admitted on the strength of a binding the body never consults. It cannot change an answer, because a macro formal is bound to an *unevaluated argument node* and an argument node is never callable: the body's `(if ...)` can only fail to call it, deterministically and identically in every cache mode. Pinned by `TestMacroCacheFormalShadowingAnOperatorAgrees`; if macro formals ever start carrying callable values, the obligations must move to the call environment.
4. **The native tier is still a hand audit, not a proof.** Identity now distinguishes two implementations sharing a registration name, but whether a given implementation deserves to be on the whitelist remains an assertion by whoever put it there.
5. **Shared mode's dead-AST anchoring is a memory hazard, not a soundness one** — see the memory section. It is the reason the recommendation below is per-runtime.

Cost of the narrowing: **zero measured reuse**. The committed synthetic corpus reproduces bit-identically before and after (706 entries, 5,000 hits, 98.0% hit rate), and no shape in the admitted tier acquires an obligation a normal program fails — a `when`-class macro records one (`quasiquote`), a `default`-class macro four.

### Load-path fix (worth reading on its own merits)

The first instrumented corpus run showed 108 hits in 655k dispatches. `lisp.TextLoader` evaluated a per-environment `Copy()` of its parse, and `Copy` clears the sealed flag — so every macro *defined by loader-loaded code* bound an unsealed formals node in every environment, while its callsites sat sealed. Every other load path (Reader/LoadString/Program/REPL) evaluates sealed trees. This branch makes the loader seal what it evaluates. Independent of caching, it closes a "loader-loaded code is invisibly second-class" gap in the sealed-AST model.

## Chain merge

Merging #386 in brought one conflict, in `lisp/loader.go`, and it is a real interaction rather than a textual one. Both sides changed what `TextLoader`'s returned closure does per expression, for different reasons:

- **#379/#380 (below)** removed the per-load deep copy when the parse is sealed *throughout*. Sharing a sealed tree across environments is the seal's purpose, and #365's rule narrows to "no MUTABLE cross-runtime sharing"; the copy is pure cost when there is nothing mutable to copy. An embedder `Reader` that does not seal still gets the copy.
- **#381 (here)** made the loader seal what it evaluates, per the load-path fix above.

These are complementary, and the `allSealed` answer the branch below already computes decides both. Sealed throughout: share as-is, no copy, and the sealing obligation is already met. Not sealed throughout: take #365's private copy **and** seal it — parser-shaped content, private to one environment, nothing mutates loader output in place, so the copy keeps the immutability contract of the parse it came from. Neither branch gives anything up.

The obligation *moved* in that merge, which needed a new test. `TestMacroCacheTextLoaderSealed` uses a sealing `Reader`, so after the merge it exercises the share path and passes whether or not the copy path seals — verified by deleting the `SealAST` and watching it stay green. Added `TestTextLoaderSealsEitherWay`, which asserts the loader-defined macro's formals are sealed through **both** readers and observes it directly on the formals rather than through the cache's admission counters. It goes red on the copy path with the `SealAST` removed.

Note that #386's `LEnv.Fork` shares sealed nodes between a template and its forks, and #387 shares one sealed parse across benchmark runtimes. Both widen the set of topologies in which a shared-mode cache entry can be reached, which is worth weighing alongside the shared-table caveats above.

## Test plan

All re-run at this head after the chain merges.

- `go test ./...` — green.
- `go test -tags elpscheck ./...` — green.
- `go test -race ./lisp/` — green (64s), including a concurrent hit/store hammer in shared mode.
- Whole suite re-run under `ELPS_MACRO_CACHE=shared` and `=runtime` — green (semantics equivalence at suite scale); `-tags elpscheck` additionally re-run under `ELPS_MACRO_CACHE=shared`, green.
- Cross-runtime shared-cache tests **execute** under `-tags elpscheck` rather than skipping: `TestMacroCacheCrossRuntimeSharedHit`, `TestMacroCacheCrossRuntimeGensymCollision` and the positive control `TestMacroCacheCrossRuntimeCheckerStillFires` all PASS with the tag set, and no macro-cache test carries a `t.Skip`. The control is what makes that meaningful — it shares the same two-runtime program *unsealed* and requires the ownership checker to panic.
- Production-scale phylum suite: PASS `shared`+cap 8192 and PASS under **`-tags elpscheck`** with shared mode on (456 subtests, 0 failures). Pre-merge-forward head additionally covered cache-off and `runtime` mode, both PASS.
- `golangci-lint` both tag configurations (0 issues), `elpsvet -test=false ./...` exit 0, `elps lint` exit 0, `elps fmt -l ./...` clean, `scripts/confidentiality-guard.sh` clean.
- `scripts/fuzz-budget-check.sh` at the chain tip, where all targets coexist: 23 targets / 6 shards / largest shard 4 / required 50 min / timeout 60 min / **headroom 10 min**.
- **Red-proofs.** Every guard in the soundness narrowing was deleted in a working tree, the named test run, and the guard restored — each one produces a *wrong answer*, not a lint failure:

  | guard removed | test that goes red | wrong answer produced |
  |---|---|---|
  | `needBuiltin` gensym obligation | `…ShadowedGensymNotCached` | `'(1 1 1)` vs cache-off `'(1 2 3)` |
  | `needSpecialOp` on `if` | `…ShadowedStructuralOpsNotCached` | `'(1 1 1)` vs `'(1 2 3)` |
  | defining-package fall-through | `…ShadowedInDefiningPackageNotCached` | `'(1 1 1)` vs `'(1 2 3)` |
  | quoted-head unquote recognition | `…QuotedUnquoteHeadNotCached` | `'('(1 1) '(1 2) 1)` vs `'('(5 1) '(6 2) 6)` |
  | same, for `'unquote-splicing` | `…QuotedUnquoteSplicingHeadNotCached` | `'('(1 1) '(1 2) '(1 3) 1)` vs `'('(1 1) '(2 2) '(3 3) 3)` |
  | `sameMacroBody` memo-key check | `…SharedFormalsTwoBodiesNotCached` | `(7 1) (7 1) (7 1)` vs `(7 1) (7 2) (7 3)` |
  | per-dispatch `defRefs` re-check | `…PurityMemoIsEnvironmentIndependent` | `'(1 1 1)` vs `'(1 2 3)` |
  | `callRefs` binder promotion | `…ShadowedBinderAtCallsiteNotCached` | `'gen…1 'gen…1` vs `'gen…1 'gen…2` |
  | per-runtime cap enforcement | `TestMacroCacheRuntimeCap` | "per-runtime cap=8 exceeded after 9 callsites" |
  | `elpscheck` control's unsealed arm | `…CrossRuntimeCheckerStillFires` | "ownership checker did NOT fire" |

  Earlier red-proofs still hold: drop the identity comparison → invalidation test fails in both modes (8 assertions); admit gensym-under-quote → prover and gensym-leak tests fail; skip expansion sealing → the sealed-entries walk fails. The sealing test had to be strengthened first — re-running the red-proof showed the original counter-only assertion could not fail.
- **`lisp/macrocache_boundary_test.go`** attacks the *narrowed* prover rather than re-testing its fixes: it pins the three readings that carry no obligation (`true`/`false` unrebindable through eight routes; `unquote` not a binding; the formal-shadow imprecision behaviourally equivalent in all three modes), gives the reviewer's second quoted-head defeat (`'unquote-splicing`) its own executed repro rather than leaving it to the shared switch arm, pins the gensym obligation as an identity rather than a purity check, and carries the memo-leak defeat into the shared **table** — two runtimes, one shared parse, the second shadowing the operator in its defining environment and at its callsite, where the entry's identity matches exactly and only the per-dispatch resolution can refuse it.
- `internal/synthphylum` — committed synthetic corpus generator built from shape statistics only (counts, depths, densities; no corpus content), with shape/determinism pins and cross-mode dispatch-equivalence tests. `TestCorpusHitRate` reproduces the headline without any private corpus: 98.0% hit rate, 706 entries, 2,188 B/entry.

## Recommendation

The soundness objection that held this branch back is **closed**: the prover no longer assumes its structural operators mean what they are spelled, it resolves them — in the macro's defining environment for its body, at the callsite for the binder syntax of its output — and re-resolves them on every dispatch. All four demonstrated defeats (quoted `'unquote`, quoted `'unquote-splicing`, shadowed `gensym`, shadowed `if`) are refused, each with a red-proof above.

**Ship as an opt-in, in per-runtime mode**, for request-path deployments: same speedup as shared mode, no cross-runtime sharing hazards, no cap policy needed in the warm-pool topology, memory released with the environment. Set `ELPS_MACRO_CACHE_CAP` if your runtimes outlive their programs — per-runtime is lifetime-scoped, not immune.

**Shared mode should stay experimental** until the table is weak-keyed (Go 1.24 `weak`), which would remove the dead-AST-anchor failure mode structurally. Uncapped shared is a footgun and arguably the code should refuse it outside tests.

**Staying in draft, and the reason has changed.** It is no longer "the prover rests on an assumption" — it is that this PR is six deep in an unmerged chain whose base is `claude/exp-env-fork`, and every other stacked branch in that chain (#377, #386, #387, #388, #389) is draft for the same reason. It should come out of draft when the chain beneath it lands and its base becomes reviewable on its own, not before; nothing in the code is now waiting on a soundness decision.


sam-at-luther:
Closing this PR. Not because the work was wrong — because a default-off POC is not a thing that can be merged, and leaving it open costs more than it carries.

**State as of `main` = `ed2538c`:** this branch forks from `663c0f3`, is **158 commits ahead and 37 behind**, and sits seventh in a chain (`#374 → #377 → #388 → #389 → #387 → #386 → #385`). Every commit under it is also unlanded, so nothing here can be reviewed against current `main` without first rebasing six other branches.

**Why closing loses nothing:** the design is tracked on #381, and the measurements and the POC's shape are preserved in this PR's history and body — a closed PR keeps its diff and its branch. If and when #381 is picked up, this is the reference implementation to start from, and it will need re-deriving against a `main` that has moved 37 commits regardless.

**Why leaving it open costs something:** two separate "merged on GitHub, absent from `main`" incidents happened in this repository today. The first was a four-PR chain that merged into its own branches rather than the default branch (rescued by #416). The second was a squash landing on a stale `main` pointer, which silently dropped a merged PR entirely (rescued by #457). Both were invisible from the PR list. A seven-deep stack at 37 commits of divergence is the configuration that produces exactly that, and every branch left open in it is one more chance to hit it.

The remaining six sealing PRs stay open — unlike this one they carry unlanded fixes and test infrastructure rather than a switched-off experiment. When that work restarts, the plan is to rebase #374 onto `main` alone, land it, and re-derive each successor one at a time, verifying by ancestry rather than by the API's `merged: true`.

Reopen freely if #381 gets scheduled and this is the fastest way back in.

---

Issue #383, opened 2026-08-12
Native contract primitives: typed handles, runtime affinity, clone and immutability protocols


sam-at-luther:
Design note (backlog, maintainer-requested). `LVal.Native` is `interface{}` — the one storage channel invisible to every protection layer: the seal cannot cover it, the ownership checker cannot attribute it (the #364 cross-runtime validator bug class), `Detach` must reject it, and the planned `copy` builtin can only share it by reference. Native stays exported (55 downstream production uses; it is the sanctioned home for embedder handles), but a set of optional, additive contracts would let natives opt into the machinery:

1. **Typed handles**: a generics-based accessor pair (`NativeValue[T](v) (T, bool)` / constructor) turning ad-hoc type switches into checked API with good errors.
2. **Runtime affinity**: `interface { BoundRuntime() *Runtime }` — checked mode asserts a bound native is not used from a foreign runtime, extending the ownership checker to the native layer.
3. **Clone protocol**: `interface { CloneNative() interface{} }` — makes deep copy extensible (copy/Detach handle declared-cloneable natives instead of reject-or-share).
4. **Declared immutability**: `interface { ImmutableNative() bool }` — opt-in inspector/fingerprint coverage for natives that promise not to change.

All additive; nothing blocks the current release. Related: #364, #372, #378 (copy semantics), #382 (field privatization).

sam-at-luther:
Item 3 now has a first shipped instance on main: 22d135a (#519) defines the `NativeCloner` interface (`CloneNative() interface{}`, explicitly shared with this design in its doc comment) which `Fork` honors for fork-time duplication of stateful native payloads, alongside the per-call `ForkWithNativeReplacer` hook (`lisp/fork.go`). Items 1 (typed handles), 2 (runtime affinity), and 4 (declared immutability) remain unimplemented, and `copy`/detach still share or reject natives rather than consulting the clone protocol — so this stays open as the umbrella design.

---
_Generated by [Claude Code](https://claude.ai/code)_

sam-at-luther:
Status refresh against current `main` (elps `6a6be2b`, substrate `cb43cf8`), now that the preheat/fork topology is the production default path.

**The premise is intact — verified against the code, not assumed:**
- `LVal.Native` is still `interface{}` (`lisp/lisp.go:329`)
- the seal still cannot cover natives (they sit in the necessarily-runtime-constructed class, `lisp/seal.go`)
[...]

PR #386, opened 2026-08-12
lisp: template-environment Fork (stacked on #387)

### LINKED ISSUE #380: Build (scoped): template-environment forking — pool-refill CPU/GC + fork-served test tooling
**Status: approved for build (maintainer), scoped per the POC decision packet.** Originally a design question; both feasibility passes and the production-scale POC are complete (see the two findings comments below for full data).

## Measured case (production-scale phylum)

- Fork **15.24ms vs 71.40ms** full load (4.7x, p=0.000, n=10 interleaved); **7.1x fewer allocations** (88.79k vs 631.53k); 9.47 vs 42.96 MiB/op churn.
- Sealed fraction at production scale: **82.5%** of 210,524 reachable LVals shared for free (59.3% at stdlib scale — big programs are mostly sealed code).
- **Not a memory feature**: resident 6.09 vs 7.96 MiB per env (~24%); the win is CPU/GC and latency.
- Correctness: transaction results byte-identical fork vs fresh; isolation fingerprint-verified under real writes; **full phylum suite fork-served: 28/28 files, identical case results, 20.1s vs 191.4s stock (9.5x)** — the fork-served test runner is a first-class deliverable, not a side effect.
- elpscheck clean end-to-end with the one-line sealed-value ownership exemption; forked envs are distinct Runtimes.

## Scope

1. Exported additive `Fork` API (est. ~400-600 lines lisp/fork.go + small clone helpers): memoized old→new env/value remap (closure capture graph, cycles), FID/env-counter continuation, template-quiescence assertion, sealed structure shared, mutable data hermetically copied.
2. **Native policy**: default share+document; optional per-native hooks for stateful natives (aligns with the clone-protocol contract in #383 — a `CloneNative`-style interface lets stateful handles opt into fork-time duplication).
3. Context rebind hook + option plumbing (the POC's productization gaps).
4. Property tests: bidirectional isolation, post-fork lambda FID uniqueness, ownership-checker behavior; benches pinning fork cost scaling with mutable fraction.
5. Embedder consumption (pool refill; fork-served test runner) lands downstream after a release, per the official-releases-only policy.

Out of scope: memory-sharing claims (measured ~24%, not the point); any change to the current release train (this is additive, post-release).

sam-at-luther:
**Stacked on #387; chain: #374 → #377 → #388 → #389 → #387 → **#386** → #385.** Base is `claude/exp-seal-opts`, so the diff shown here is this PR's own work only. Post-release.

Additive only — one new exported entry point plus two option constructors and one opt-in interface. No existing signature or behavior changes, and the evaluator hot path is untouched: `Fork` runs at pool-refill time, never on the request path.

Closes #380.

## What this adds

`LEnv.Fork` clones a fully loaded, quiescent *template* environment into an independent environment on a fresh `Runtime`, **sharing every sealed value** with the template and hermetically copying only the mutable remainder. It is a dividend of the sealed-AST work: `docs/sealed-ast.md` establishes that a sealed node's bytes never change after parsing, so a value that can never change is safe to hand to any number of runtimes — and at production scale sealed program structure is the overwhelming majority of a loaded environment.

```go
func (env *LEnv) Fork(opts ...ForkOption) (*LEnv, error)

func ForkWithContext(ctx context.Context) ForkOption
func ForkWithStderr(w io.Writer) ForkOption
func ForkWithNativeReplacer(fn func(payload interface{}) (interface{}, bool)) ForkOption

type NativeCloner interface{ CloneNative() interface{} }
```

Plus `elpstest.Runner.NewEnvFn`, the seam a fork-served test runner needs, and a one-line sealed-value exemption in the `elpscheck` ownership checker.

## Measured numbers

**Stdlib scale** — `BenchmarkEnvConstruction` (`lisp/lisplib/fork_test.go`), the in-repo pair CI can track. `benchstat -col /mode`, n=20:

```
          │     fork     │               fullload                │
          │    sec/op    │    sec/op     vs base                 │
EnvConst    366.8µ ± 17%   844.6µ ± 17%  +130.28% (p=0.000 n=20)
          │     B/op     │     B/op      vs base                 │
EnvConst    284.4Ki ± 0%   525.9Ki ± 0%   +84.94% (p=0.000 n=20)
          │  allocs/op   │  allocs/op    vs base                 │
EnvConst    1.257k ± 0%    5.354k ± 0%   +325.93% (p=0.000 n=20)
```

2.3× wall, 4.3× fewer allocations — and faster than the feasibility pass's exported-API prototype (471 µs / 1,367 allocs), because the in-package implementation clones symbol maps directly instead of replaying `Package.Put`.

**Embedder-runtime scale** — an embedder's own lisp runtime library (~11k lines) plus a loaded program, measured out-of-tree against this implementation with a local module `replace`: fork **8.7 ms / 36.8k allocs / 4.3 MiB churn** vs full load **50.9 ms / 440.9k allocs / 106 MiB churn** — 5.9× wall, 12.0× fewer allocations (n=20, p=0.000). The full-load arm carries the embedder's per-environment fixture setup, so read it as indicative rather than as a pure interpreter measurement.

**Production scale** — the POC's numbers stand as the production-scale evidence (15.2 ms vs 71.4 ms, 4.7× wall / 7.1× allocs on a ~72k-line production-scale phylum; 82.5% of 210,524 reachable values sealed and shared for free). That corpus is proprietary and not reproducible in-repo, which is why the two smaller scales are reported alongside.

The margin widens with program size, as predicted: a full load re-evaluates every top-level form, while a fork walks only the mutable and closure fraction, which shrinks proportionally as programs grow.

**Not a memory feature.** Resident memory per fork was ~24% below a fully loaded environment. Adopt Fork for CPU, GC pressure, and refill latency — not for RSS.

## The contract

- **Quiescence is asserted, with no bypass.** `Fork` errors on a non-empty call stack, an active eval entry, or pending condition handlers. A mid-evaluation environment holds torn state and a fork of it would too.
- **Forks are distinct `Runtime`s** — exactly the unit the `Runtime` doc comment already requires per goroutine. Template and forks may evaluate concurrently. `Fork` itself only reads the template, so concurrent forks of one quiescent template are safe.
- **Sealed shared, mutable copied**, with aliasing and cycles reproduced (memoized old→new env/value tables, seeded before descending — which is what makes `labels` mutual recursion and closure↔environment cycles fall out for free).
- **Env-ID and gensym counters continue** past the template's. Load-bearing: FIDs are `"_fun"`, so a restarted counter eventually mints FIDs colliding with inherited ones (corrupting `funNames` and tail-call `TerminalFID` matching), and a restarted gensym counter re-mints load-time names.
- **What does NOT travel:** the template's `context.Context` (rebind with `ForkWithContext`), `Profiler`, `Debugger`, call stack, condition stack, step accounting, macro-expansion debug metadata.
- **Native payloads are shared by default** — correct for the immutable handles that dominate real templates (compiled regexps, timestamps). Stateful payloads are the embedder's to handle, in order of preference: keep the state out of the template and run its hooks per fork; implement `NativeCloner`; or install `ForkWithNativeReplacer`. A shared stateful native is the one state leak no black-box isolation test can see — audit your template's native census when adopting this.

`NativeCloner` is deliberately one minimal interface with one method returning an independent payload, named to sit alongside the clone-protocol contract sketched in #383.

## Test plan

- `lisp/fork_test.go` — pairwise audit of the entire forked graph against the template: sealed pointer-shared, mutable pointer-distinct with equal content, template aliasing and cycles reproduced in the fork, with anti-vacuity floors on both the shared-sealed and copied counts. Plus quiescence rejection (idle and mid-evaluation), counter continuity, `NativeCloner`, `ForkWithNativeReplacer`, `ForkWithContext`, `ForkWithStderr`, runtime-config inheritance.
- `lisp/lisplib/fork_test.go` — behavioral tests over real parsed programs (closures, macros, `labels` mutual recursion, mutable globals): bidirectional isolation stated as observable mutations, *and* as a full-state fingerprint (structure-only hash of the env tree plus every package symbol, walking into captured environments via `internal/funraw`) — fork-side writes leave the template's hash bit-identical, a fresh fork reproduces it exactly, and the mutated fork's hash moves as the mutation control. Post-fork lambda FID uniqueness on both sides.
- `lisp/fork_ownership_elpscheck_test.go` — the checker model: sealed cross-runtime sharing sanctioned, mutable cross-runtime leaks still panic, a fork serves independently.
- `elpstest/fork_runner_test.go` — the fork-served runner end to end, including the stateful testing-suite rebind.

**Red-proof** (mutate the implementation, confirm the suite goes red, restore):

| Sabotage | Result |
|---|---|
| Widen the seal-sharing branch to share mutable arrays | `TestForkSharingContract` (pointer-shared mutable), `TestForkBidirectionalIsolation`, `TestForkIsolationFingerprint` all red |
| Remove the value-remap memo lookup | stack overflow on the self-referential value — cycles no longer terminate |
| Drop the env-ID counter carry | `TestForkCounterContinuity` + `TestForkLambdaFIDContinuity` red (FID collisions from the third lambda) |
| Drop the gensym counter carry | `TestForkCounterContinuity` red (counter + re-minted load-time gensyms) |

The last two red-proofs found two genuinely insensitive tests and fixed them (commit `e7378d3`): the gensym check compared `0 == 0` on a template that had never gensym'd, and the behavioral FID check minted a single lambda where the first collision only appears at the third.

**Gates** (all green at the chain head): `go test ./...`; `go test -tags elpscheck ./...`; `go test -race ./lisp/`; `golangci-lint run ./...` and `--build-tags elpscheck ./lisp/...` (0 issues both); `elpsvet -test=false ./...` (exit 0); `elps fmt -l ./...` clean; `scripts/confidentiality-guard.sh` clean; `make test-examples`.

## Chain merge

Merging #387 in brought one conflict, in `lisp/ownership_check_elpscheck.go`, and it is the same collision resolved one link down: **three** branches independently concluded that sealed nodes must be exempt from the runtime ownership checker, each from its own motivating topology — #372 from the parse cache, #379 item 2 from the shared benchmark program, and this branch (#380) from `LEnv.Fork`, which shares every sealed node between a template and its forks.

The resolution keeps all three. The Allowlist entry now names all three sanctioned sharing topologies, credits the three issues that arrived at the exemption independently (worth recording in itself — an exemption three people reach separately is a design, not an oversight), and carries #380's framing that immutability rather than per-runtime confinement is what protects a sealed node, by the same reasoning that exempts the singletons, with a pointer to `docs/sealed-ast.md` §3 where the enforcing layers are laid out together.

The code hunk was a false conflict: both sides had the identical `isSingleton(v) || v.sealed` short-circuit and differed only in whether the comment sat above the `if` or inside its body.

`lisp/fork_ownership_elpscheck_test.go` merged cleanly and is unchanged; with #389's and #387's tests it makes four independent red-proofs of the exemption — through `Put`, through `eval`, through `Copy` and through `Fork`.

## Earlier forward merge

Merged `origin/claude/exp-seal-tooling` in (commit `88f09f1`), which brought the third #382 privatization slice. No textual conflicts, but three semantic ones, all resolved in favor of the privatized accessors:

- `*LFunData` → `*funData`, `MapData.Map` → the fixed `mapBacking`, and the fork's fresh sorted map now built with `NewMapData`.
- In-package tests use `funData()` / `mapBacking` directly.
- The external isolation fingerprint uses `FID()`/`Package()` for identity and `internal/funraw.Env` for the captured environment — it has to walk in there, since a fingerprint stopping at the function header would not see closure state, which is exactly the state fork must not share.

## Downstream

Embedder consumption (pool refill, fork-served test runner) lands after a release, per the official-releases-only policy. Zero embedder change means zero regression and zero benefit — the entry point has to be called.

---


sam-at-luther:
Superseded by #519 (collapsed with #387 into the payoffs layer of the rebuilt stack #517 → #518 → #519), including the fork backing-array fix and the counter-continuity test sensitivity work. Branch preserved for reference.

---

PR #387, opened 2026-08-12
lisp: seal-enabled optimizations (stacked on #389)

### LINKED ISSUE #379: Seal-enabled optimizations: defensive-copy audit, bench-harness sharing, interning, path-compile cache

sam-at-luther:
Near-term seal-enabled optimizations from #379. Each item is its own commit with its own measurements.

**Stacked on #389; chain: #374 → #377 → #388 → #389 → **#387** → #386 → #385.** Base is `claude/exp-corruption-fuzz`, so the diff to review is this PR's own commits, not the whole seal stack.

This branch sits directly above #389 on purpose. Both branches independently added the same sealed-node exemption to the runtime ownership checker, and putting #387 on top resolves that collision **once, here**, rather than at merge time when neither author is looking. See "Chain merge" below.

The through-line is the pattern #374 established: the seal converts *"copy defensively because sharing is unsafe"* into *"prove immutable, then share."* Two of the four copy sites that existed for that reason are now shares, and the third — the one the issue named — turns out to live downstream.

---

## Results

| # | Item | Disposition | Headline |
|---|------|-------------|----------|
| 1 | `elpstest.RunBenchmark` sealed sharing | **Done** | −4.2 µs / −5.7 KB / −64 allocs per iteration (50 KB program: −3.2 ms / −1.36 MB / −14,500 allocs); ns/op unchanged |
| 2 | Compiled-path LRU in libelpspath v1 | **Dispositioned — code is downstream** | Prototyped and measured: −91% time, −93% allocs end-to-end. Must land in substrate; design + numbers below |
| 3 | Structure-aware elpspath fuzz target | **Done — found a live defect** | 120 s clean (746,833 execs); found an uncatchable panic + silent corruption in the copying ops, fixed here |
| 4 | Defensive-copy audit | **Done** | Full table below; one hit implemented: `TextLoader` −72.3% time, −61.9% allocs per load — **real, but no identified consumer today; see the framing note** |
| 5 | Symbol interning | **Skipped — profiled, below threshold** | `lisp.Symbol` is 0.83% of allocations, and 98.3% of those are registration-time, not runtime-constructed |

---

## 1. `elpstest.RunBenchmark`: share the sealed program

`RunBenchmark` deep-copied the parsed program once per iteration, because each iteration runs in a fresh `Runtime` and #365 established that an `LVal` shared by two Runtimes was unsafe under any circumstances.

The seal narrows that rule. Sealed nodes are frozen program-literal storage under copy-on-write protection, so #365 becomes *no **mutable** cross-runtime sharing* — the formals precedent from #374.

Three things make the share provable rather than hopeful:

- **`checkOwnership` exempts sealed nodes**, documented in the Allowlist section alongside the three singletons. `TestOwnershipCheck_SealedASTExempt` pins it, with a red-proof second half: an unsealed tree of identical shape crossing the same two runtimes must still panic. The exemption is "sealed", not "looks like an AST".
- **`VerifySealedASTs()` at the end of every benchmark.** In checked builds every sealed parse in the process is re-fingerprinted, so an in-place write to the shared tree fails the benchmark instead of passing silently. A free nil untagged.
- **An unsealed program still gets #365's per-iteration copy**, outside the timed region as before.

**#365's contamination test keeps its spirit.** Nothing mutable is shared across runtimes; the seal plus the fingerprint oracle is what proves it. Three tests that previously *skipped* their cross-runtime half under `elpscheck` (`cow_seal_test.go`, `path_mutate_list_test.go`, `seal_watchdog_evalpath_test.go`) now run that half in checked builds — strictly more coverage than before, and the `elpscheckEnabled` build-tag constant went away with the skips.

### The oracle had to be armed in CI (review follow-up)

The bullet above was true and inert. `VerifySealedASTs` is a nil call unless the `elpscheck` build tag is set, and `.github/workflows/benchmark.yml` measures with **no build tags** — so in CI every benchmarked path ran with its own safety net switched off, and the sharing was unprotected in exactly the place the sharing happens.

Nothing cheaper substitutes for the fingerprint. Corruption of a program literal is **deterministic**: every iteration produces the same wrong answer, so neither an iteration-to-iteration value differential nor a diff against an independent fresh parse diverges. The seal-time fingerprint is the only oracle that sees it.

`make bench-elpscheck-smoke` (`go test -tags elpscheck -bench=. -benchtime=1x -run='^$' ./...`) now runs in `.github/workflows/elps.yml`, immediately after `make test-elpscheck` so it inherits that step's warm tagged build cache. One iteration per benchmark is all a fingerprint check needs. **Measured marginal cost: 8.5–11.0 s** (n=3, warm, 4-core amd64), in a job that already runs the full suite, the race suite and the tagged suite.

The untagged measurement run is untouched, deliberately: `elpscheck` adds census bookkeeping and per-call integrity checks to the timed region, so tagging benchmark.yml's own `go test -bench` would have been the cheap fix and the wrong one — the numbers would stop being comparable with the base arm and with every historical run. Correctness is adjudicated by the smoke run, performance by the measurement run.

**An oracle only reports on the writes it is given**, and that is the half that would have made this ceremonial. Before `BenchmarkSealedLiteralCoW`, **no benchmarked program reached any copy-on-write-on-sealed site**, so deleting a guard changed nothing anywhere in the benchmark suite. The new benchmark reaches all three (`stable-sort`, `slice-vector`, `append-vector`) and is an ordinary benchmark of those builtins untagged. `TestSealedLiteralCoWBenchmarkCoversEverySite` pins that coverage and reads the site list **from the census** rather than restating it, so a fourth site added to `lisp/cow_sites.go` fails until a benchmark covers it.

**Red-proof**, with the `stable-sort` sealed CoW guard disabled in place:

| run | result |
|---|---|
| `go test -bench=. -benchmem -benchtime=100ms ./...` (benchmark.yml's own command) | **exit 0**, 49 benchmarks, green — the caution, reproduced |
| `make bench-elpscheck-smoke` (the new step) | **FAIL**, first at `BenchmarkSealedLiteralCoW`: `sealed AST verification failed after benchmark: 1 of 232 sealed roots were mutated in place` |

Guard restored; both green again.

`scripts/ci-gates-test.sh` gains four guards, each with its negative control run: the Makefile target exists; its recipe carries `-tags elpscheck`, `-bench` **and** `-benchtime` (dropping any one leaves a target that still runs and still passes while checking nothing); a workflow actually invokes it; and benchmark.yml still measures untagged, so the arming cannot be "fixed" later by tagging the measurement run. 125 → **132 passed, 0 failed**.

**Measurements** (interleaved, n=10, benchstat):

Work removed from the harness, per iteration:

| Program | Time | Bytes | Allocs |
|---|---|---|---|
| typical elps bench program | 4.2 µs | 5.7 KB | 64 |
| 50 KB program (phylum shape) | 3.2 ms | 1.36 MB | 14,500 |

Timed region, 6 `RunBenchmark` benchmarks: all `~` (p > 0.10) on sec/op, B/op and allocs/op; geomean −4.94% sec/op, inside this host's noise band. Sharing the sealed tree costs nothing on the eval path — evaluation triggers no COW copying — and the removed copy was already outside the timed region. This is a harness wall-clock and GC-pressure win, not an ns/op win, and it is reported as such.

---

## 2. Compiled-path LRU: the v1 grammar is not in this repo

The issue asks for an LRU at "the v1 parse step in libelpspath". That step is not here. The engine moved upstream, but **the v1 jq-string grammar deliberately stayed downstream** in `luthersystems/substrate` (`internal/substrate/shiro/libelpspath/parse.go`) — `libelpspath/query.go`'s package doc records the split. Upstream `libelpspath` has no string parsing at all: the `?`-family dispatches positional steps by type switch.

So there is nothing to cache in elps. Rather than skip the item, I prototyped and measured it where it lives (local worktree only — **no substrate commits**), so whoever lands it there starts from numbers instead of a guess.

**End-to-end through the v1 ops** (interleaved, n=10, benchstat; `.users[0].name`):

| | uncached | cached | delta |
|---|---|---|---|
| `get-path` sec/op | 3314 ns | 277 ns | **−91.65%** (p=0.000) |
| `set-path!` sec/op | 3229 ns | 316 ns | **−90.21%** (p=0.000) |
| `get-path` B/op | 1581 | 224 | −85.83% |
| `set-path!` B/op | 1599 | 240 | −84.99% |
| `get-path` allocs/op | 28 | 2 | −92.86% |
| `set-path!` allocs/op | 29 | 3 | −89.66% |

**Concurrency — this is the part that decides the design.** Two variants, same cache size, `RunParallel` on 4 procs:

| variant | serial hit | parallel hit |
|---|---|---|
| uncached | 2950 ns | 1850 ns |
| mutex + `container/list` LRU | 22.7 ns | 61.7 ns (36–95 ns, high variance) |
| `sync.Map` + bounded flush | 18.1 ns | **21.9 ns** (stable) |

A textbook LRU takes the write lock **on every hit** (`MoveToFront`), so recency bookkeeping serialises the read path: 2.8× worse under parallelism and visibly noisy. The `sync.Map` read path is lock-free and flat. Recommendation: **`sync.Map` keyed by the path string, bounded at 256 entries by a wholesale flush**, not a recency LRU. Path values are immutable after construction (every `Path` method reads its receiver only), so cache entries are safe to share across goroutines.

**Sizing, and the miss case.** 256 is a working-set bound, not a hit-rate target: substrate's core lisp passes paths as *variables*, so the live key set is whatever phylum code supplies at runtime — bounded, unknown, and exactly the shape that argues for a cap plus a cheap flush over an unbounded map. The cost when the cache never hits is real and measured: **2270 ns vs 1850 ns uncached (+23%) and 2 extra allocs** on a working set 4× the cache. That is the number to weigh, and it is why the bound belongs in the design rather than being assumed away.

**Not landed here.** It cannot be: the code is in another repo, and this PR is elps-only. Prototype is preserved and reproducible.

---

## 3. Structure-aware fuzz target for the elpspath engine

Coverage of the `?`-family was indirect: `FuzzApplyStdlib` enumerates the builtins (elpspath is default-loaded) and reaches them with **generic** value seeds built to exercise 240 unrelated callables. The path engine's defects do not live in the argument types. They live in the interaction between a nested **structure** and a step sequence that walks partway into it — a step only reaches the interesting arithmetic when it resolves against a specific container length, which a generic value generator produces by accident.

`FuzzPathEngine` generates both axes together and crosses them with all seven ops:

- **Structures**: nested sorted-maps, vectors, lists, scalars — bounded by a node budget and depth limit so an input cannot spend the fuzzer's time in the allocator.
- **Steps**: string keys, integer indices (including the negatives and magnitudes that break index arithmetic), `'*`, and `(range a b)`. Keys and indices are drawn **preferentially from what the generated structure actually contains** — a fuzzer that mostly misses spends its budget on the "key absent" branch. Misses and out-of-range values still appear; they are the other boundary.

Invariants per input: no panic escapes (applied through `env.FunCall`, **not** `env.Eval` — eval's `recover()` would turn a crash into an `internal-panic` condition and the target would report a pass on it); catchable errors only, never a nil `*LVal`; the result renders; **sealed nodes bit-identical afterwards for all seven ops including the mutating ones** (`SealedASTFingerprint`); and copy-vs-mutate — the non-mutating ops must leave the input structurally unchanged.

`TestPathGenCoverage` is the harness's own gate: a generator that silently produces one shape forever cannot fail, and nothing about that is visible from a green fuzz run.

### It found a live defect

Six seconds in, on `[]byte("C%11021100")` — `(?del '(x) 0)`. The del and range-set write-back wrote the **array** layout unconditionally:

```go
in.Cells[1].Cells = vals
dims := in.Cells[0]
dims.Cells[0].Int = len(vals)
```

An array is `[dims, data]`; a list holds its cells directly. That write was correct only because the **mutating** entry points reject lists (`errMutateList`, the substrate#378 fix). But the non-mutating `?set`/`?del`/`?nil` reach the same write-back through the unguarded `setMutate`/`deleteMutate` — and they *do* accept lists, that being the point of a copying op. They copy first, and the copy is a list whenever the input was one. On such a copy:

- a list shorter than two elements has no `Cells[1]`, so it **panicked** — an uncatchable crash where lisp code is entitled to handle a condition; or
- a longer list had the new cell slice written into its **second element** and an `Int` field of its **first** corrupted as "dims" — a mangled structure returned as the answer, **with no error at all**. That one is worse, because it is silent.

Fixed by `storeCells` dispatching on the container layout. `TestCopyOpsOnList` pins it across every non-mutating op × index/range step × list lengths 0,1,2,3,5. Red-proof: forcing the array branch unconditionally reproduces the panic on ranges at every length and the original fuzz input at length 1.

### Fuzz shard accounting

**No CI wiring needed.** `scripts/fuzz.sh` discovers targets dynamically via `go test -list` rather than from a table, so a new target is covered by the nightly sweep the day it lands — confirmed: `fuzz.sh --list` reports `FuzzPathEngine`.

The sixth shard this budget needs arrives from #389, one link below, which added four targets of its own and the shard to carry them. Re-derived on this branch after the merge, with all 23 targets present:

```
targets discovered      23
shards                  6
largest shard           4 target(s)
required                50 min
timeout-minutes         60 min
headroom                10 min
```

`make ci-gates-test` (132 passed) and `make fuzz-classify-test` (33 passed) are green.

**120 s run, clean**: 746,833 execs, 323 new interesting inputs, no failures. Seeds are synthesized byte strings written in the file; the single `testdata` entry is the minimised driver input for the defect above. **Nothing is corpus-derived.**

---

## 4. Defensive-copy audit

Every `Copy()` call site in non-test elps code, classified by the issue's question — *does this copy exist only because sharing used to be unsafe?*

| Site | Class | Rationale |
|---|---|---|
| `lisp/loader.go:59` `TextLoader` | **Replaced with sealed-share** | The loader's own analogue of item 1 — one parse, many environments. **Done, with tests + measurements.** |
| `lisp/env.go:757` `registrationFormals` | Already shares | The #374 precedent: sealed formals aliased into every env. |
| `lisp/builtins.go:424` `sealedFormalsCopy` | **Keep — semantic** | Boundary copy of the *caller's* value at registration, then sealed. It is what makes `registrationFormals`'s share safe; removing it would alias an embedder's mutable table. |
| `lisp/lisp.go:1362` `copyCells` | **Keep — mechanism** | The implementation of `Copy` itself. |
| `lisp/detach.go:207` | **Keep — mechanism** | The detach walker. |
| `lisp/lisp.go:607` `dims.Copy()` in `Array` | **Keep — semantic** | Dims is bookkeeping the array mutates in place. |
| `lisp/lisp.go:1074` `ArrayDims` | **Keep — semantic** | Hands the caller a value it may mutate; returning the live dims cell would let a caller corrupt the array. |
| `lisp/lisp.go:184` `fd.Env.Copy()`, `lisp/env.go:1618` `funEnv().Copy()` | **Keep — not LVals** | `LEnv` copies (closure environments), out of scope for the seal. |
| `Runtime.Stack.Copy()` ×9 (`builtins.go` ×4, `env.go` ×4, `macro.go` ×2, `op.go`, `libjson/json.go`) | **Keep — semantic** | Snapshots of a live, mutating call stack at error-capture time. Sharing would make every error display the stack as of whenever it was printed. |
| `elpstest/lisptest.go:439` | **Keep — fallback** | The unsealed-program path retained by item 1. |
| `lisp/builtins.go:1495–1500, 1574–1584` (sort / map element copies) | **Uncertain — reported** | Copies of user-list elements before handing them to a user comparator. Plausibly shareable when sealed, but the protection here is against a *user function* retaining or mutating an argument, which the seal does not by itself answer. Not worth guessing at for a copy that is per-comparison in a sort. |
| `lisp/builtins.go:1078` `builtinCompose` formals copy | **Uncertain — reported** | The copy becomes a new `LFun`'s formals. `registrationFormals` proves sealed formals can be *aliased*, but a composed function's formals are built at runtime and consumed while constructing the wrapper; needs the formals-sharing property test extended before sharing. |
| `lisp/op.go:177` `assert`'s `test.Copy()` | **Candidate — reported** | Evaluating a caller's unevaluated (sealed) test expression. Evaluation must not write sealed nodes, so the copy looks defensive-only — but `assert` is not hot, so the risk/benefit does not justify moving it in the same PR that changes three other sharing rules. |
| `lisp/builtins.go:1825` `range`'s `x.Copy()` | **Keep — semantic** | Copies the loop variable per element; sharing would give every element of the result the same value. |
| `cmd/elpsvet/testdata/**` | N/A | Analyzer fixtures. |

### `TextLoader`, the one hit

`TextLoader` reads once and returns a `Loader` called once per environment, deep-copying the whole parse each call. Same #365 premise, same answer. A `Reader` that does **not** seal — the interface is public and embedders implement it — keeps the per-load copy, and the decision is made once at `TextLoader` time rather than per load.

Interleaved, n=10, benchstat; 50 KB source, per-environment load:

| | before | after | delta |
|---|---|---|---|
| sec/op | 3426.0 µs ± 8% | 947.8 µs ± 6% | **−72.33%** (p=0.000) |
| B/op | 1959.6 KiB | 651.2 KiB | **−66.77%** (p=0.000) |
| allocs/op | 23.284k | 8.864k | **−61.93%** (p=0.000) |

**Framing correction — who benefits from these numbers today: nobody identified.** `TextLoader` has **zero callers** anywhere in this repository, and this PR's own downstream sweep (below) concluded that substrate reaches elps through the `Reader` path and **never had the copy `TextLoader` had**. So the −72.3% / −61.9% figures are real and reproducible, but they are measured on a function no known consumer calls. What the change does have is a cost that lands on everyone: `TextLoader` is **public API**, and its semantics move from "every load gets its own tree" to "loads share one sealed tree when the Reader seals". An embedder that today mutates what a `Loader` handed it — legal, if unwise, under the old contract — gets copy-on-write where it had ownership. That trade may well be the right one, but it should be read as *changing a public contract for a speculative future consumer*, not as *shipping a measured win to an existing one*.

`TestTextLoaderSharesSealedAST` loads one parse into four Runtimes, requires every one to agree with the first, and calls `VerifySealedASTs`. It runs under `-tags elpscheck`, where the ownership checker would panic on a cross-runtime mutable value — passing there is what separates "sealed and shared" from "shared and hoped". `TestTextLoaderCopiesUnsealed` is the red half.

### Downstream sweep (read-only, report only — no substrate commits)

- **Zero `.Copy()` sites** in substrate's non-vendor Go. Its parse cache already hands out the cached AST by identity untagged; `env.load` evaluates shared exprs without copying, so the `Reader` path never had the copy `TextLoader` had.
- **One finding worth acting on there.** Under `-tags elpscheck`, `checkedParseCacheHandoff` returns a **detached deep copy** instead of the aliased master. Its own comment gives the reason: *"an elps ownership checker would otherwise — correctly — panic on the first cache hit evaluated by a second Runtime."* Item 1's sealed exemption removes that reason. Dropping the detach would let checked builds exercise the **same sharing production uses**, instead of a detached copy that the checked configuration alone sees — the same fidelity gain the three un-skipped tests got here. Worth a substrate issue.

---

## 5. Symbol interning: profiled, skipped

Profile first, as the issue asks. Alloc profile over the lisp bench suite (`BenchmarkEnvInit`, `EnvGet`, `EnvFunCall*`, `MacroDefun`), 67.3M alloc objects / 5633 MB total:

| | alloc_objects | alloc_space |
|---|---|---|
| `lisp.Symbol` | 557,096 — **0.83%** | 59.51 MB — **1.06%** |

Below the 1–2% bar on objects, marginal on space. But the callers make the case decisive:

| caller of `lisp.Symbol` | share |
|---|---|
| `AddBuiltins` | 44.5% |
| `UsePackage` | 42.0% |
| `AddSpecialOps` | 5.9% |
| `Formals` | 3.4% |
| `AddMacros` | 1.7% |
| **`(*LEnv).eval`** | **1.7%** |

The issue proposes interning *runtime-constructed* symbols. Those are the `eval` line: 1.7% of 0.83% = **0.014% of allocations**. An intern table cannot be measured there, and would add a lookup to a path that currently does one small allocation.

The other 98.3% is registration-time — the same fixed symbol names re-allocated per environment. That is not an interning problem, it is the shape issue #379 item 5 already names (share whole builtin function objects, not just formals), and it is where the remaining env-init allocation actually is. **Recommendation: close interning; keep the profile as the argument for the builtin-object sharing item.**

---

## Chain merge: the sealed-node exemption, resolved once

#389 (one link below) reached the same `checkOwnership` exemption independently, from #372's parse-cache topology; #386 (one link above) reaches it a third time, from `LEnv.Fork`. Three conflicts, all resolved here by keeping both intents rather than picking a winner:

- **Allowlist doc comment.** #372 says *why* the exemption is needed (the checker forbade the exact topology the seal exists to make safe; found by `FuzzSharedProgramMultiEnv`; `Copy`/`detach` clear the flag so runtime storage stays checked). #379 item 2 says *what licenses it* (copy-on-write at every mutation site, `VerifySealedASTs` fingerprinting in checked builds, the `-race` watchdog). Neither is redundant. Merged into one entry carrying both, naming every sanctioned sharing topology.
- **`checkOwnership` itself.** Both sides added the identical `v.sealed` short-circuit, differing only in operand order (no operand has a side effect). Kept #389's, which arrives first in the chain, with its inline pointer to the allowlist entry.
- **Tests.** Kept all three: `TestOwnershipCheck_SealedNodesExempt` (Put path) and `TestOwnershipCheck_CopyOfSealedIsChecked` (the Copy boundary the exemption depends on) from #389, `TestOwnershipCheck_SealedASTExempt` (eval path) from here. `checkOwnership` is called from `LEnv.Put`, `LEnv.PutGlobal` **and** `env.eval`, so a red-proof through only one of those call sites would not notice the exemption going wrong at the other; a comment on the eval-path test says so, since the two otherwise read as duplicates and would invite a cleanup.

A fourth conflict git could not see, because the two changes are in different files: this branch deleted the build-tag pair defining `elpscheckActive` along with the skips it existed for, while #388 added a **new** consumer of that constant (`TestCoWCounterInactiveByDefault`, whose whole subject is the untagged build). The merge was clean and the build broke. Reinstated as `lisp/cow_counter_tag_{on,off}_test.go` — named for its one remaining consumer, with the history recorded and a note that a new use of it is usually a checked build stepping around what it exists to check.

---

## Gates

| Gate | Result |
|---|---|
| `go test ./...` | pass |
| `go test -tags elpscheck ./...` | pass |
| `go test -race ./lisp/` | pass |
| `make bench-elpscheck-smoke` (new) | pass |
| `golangci-lint run --build-tags elpscheck ./lisp/...` | 0 issues |
| `elpsvet -test=false ./...` | exit 0 |
| `elps fmt -l ./...` | clean |
| `scripts/confidentiality-guard.sh` | clean |
| `make ci-gates-test` | 132 passed, 0 failed |
| `make fuzz-classify-test` | 33 passed, 0 failed |
| `make fuzz-budget-check` | fits, 10 min headroom (23 targets / 6 shards) |
| `FuzzPathEngine -fuzztime=120s` | clean, 746,833 execs |

Every performance claim above is interleaved benchstat at n=10.

Closes #379 — items 1, 3 and 4 implemented; item 2 dispositioned with measurements and a downstream home; item 5 dispositioned with the profile that answers it. (Issue items 5 and 6 in the original list — builtin-object sharing and the libjson allocation profile — are explicitly profile-first candidates and are not touched here; item 5's profile above is the argument for picking up the first of them next.)


sam-at-luther:
**The elpspath copying-op fix moved to the release branch.**

The defect `FuzzPathEngine` found (item 3 above — the array layout written into list copies: uncatchable panic on short lists, silent structural corruption on longer ones) lives in code that #374 moves upstream, not in anything this PR optimizes. A live corruption bug should not wait on a post-release optimization PR, so `storeCells` + `TestCopyOpsOnList` were cherry-picked onto `claude/exp-seal-tooling` (`0442c52`) and #374's body now lists them as layer 6.

The red-proof was re-verified after the transplant on the release branch: forcing the array branch unconditionally reproduces both failure modes — `index out of range [1]` on 1-element lists, and at length 2 the silent case, where the test catches the input coming back as `'(10 1)` with the phantom dims `Int` scribbled in.

The updated base is merged in here (`c2a1c79`), so this PR's diff no longer carries a duplicate — `path.go` and `path_copy_list_test.go` are gone from it and the remaining 15 files are the optimizations, the fuzz target and the audit. `FuzzPathEngine` and its minimised `testdata` seed stay on this PR; the seed now passes against the fixed base. Read item 3's "fixed here" as "fixed on the base branch."

Gates re-run on the merged head: `go test ./...` and `-tags elpscheck` pass, golangci-lint 0 issues on both passes, `elpsvet -test=false ./...` exit 0, `elps fmt -l` clean.

---

sam-at-luther:
## Independent re-verification of the armed oracle (67a4435)

Re-ran the red-proof and the cost measurement from scratch on a separate checkout, rather than carrying the numbers forward. Three corrections/additions to what the body says.

### 1. The real CI cost is lower than the body's local estimate

The body reports **8.5–11.0 s** from a local sample. The step has since run in CI and can be read directly off the job:

| where | step | measured |
|---|---|---|
| #387 @ 67a4435, `Lint & Test` | `Sealed-AST oracle over the benchmarks (elpscheck)` | 21:04:27 → 21:04:32 — **5 s** |
| #385 @ f571fe1, `Lint & Test` | same step | 22:07:48 → 22:07:55 — **7 s** |

Locally, warm, 4-core amd64, n=3: **5.9 s / 7.5 s / 7.8 s**, 49 benchmarks executed. So the body's figure is a conservative over-estimate; the true marginal cost is ~5–8 s on a job that already spends ~3 min. No change needed, but the cheaper number is the honest one.

### 2. Red-proof reproduced, guard restored

`lisp/builtins.go` `stable-sort`, `if list.sealed {` → `if false && list.sealed {`:

| run | result |
|---|---|
| `go test -bench=. -benchmem -benchtime=100ms -count=1 -run='^$' -timeout=10m ./...` (benchmark.yml's exact command) | **exit 0**, 49 benchmarks, 0 FAIL — the caution, reproduced |
| `make bench-elpscheck-smoke` | **exit 2**, first at `BenchmarkSealedLiteralCoW`: `sealed AST verification failed after benchmark: 1 of 232 sealed roots were mutated in place` |

The failure then cascades across the remaining benchmarks (`BenchmarkEnvGet` at 1 of 591, `BenchmarkEnvFunCallBuiltin` at 1 of 592, … ) because the census is process-wide and the mutated root stays mutated — so the report names the first benchmark to reach the site, not only the site itself. Guard restored: smoke back to **exit 0, 49 benchmarks, 0 FAIL**.

### 3. One qualification the body should carry

`-run='^$'` is load-bearing in that first row, and it is worth saying why. Dropping it — running `go test -bench=.` *with* tests — fails loudly untagged (39 FAIL lines), because the CoW unit tests assert the `stable-sort` guard's behaviour directly and a deleted guard is exactly what they are looking at.

So the precise claim is narrower than "untagged CI cannot see this". Untagged CI **does** catch the deletion of an existing guard, via those unit tests in `Lint & Test`. What it cannot see is the case the sharing actually exposes: an in-place write reached through a benchmarked path that no unit test asserts on — a new mutation site, or an existing one reached by a new route. That is deterministic, produces no value differential and no divergence from a fresh parse, and only the seal-time fingerprint reports it. The smoke run is what covers that, and it is still the only thing that does; the benchmark workflow's own command remains green through it.

### Chain

Carried up: #387 @ `67a4435` → #386 @ `9c9890b` → #385 @ `f571fe1`. Both merges were conflict-free. Full gate set re-run on each (`go test ./...` plain and `-tags elpscheck`, `-race ./lisp/`, both golangci-lint passes, `elpsvet -test=false`, `elps fmt -l`, confidentiality guard, `fuzz-budget-check` — 10 min headroom, `ci-gates-test` 132 passed). All three PRs green in CI, and the new step is confirmed executing on all three.

---

sam-at-luther:
Superseded by #519 (collapsed with #386 into the payoffs layer of the rebuilt stack #517 → #518 → #519). One deliberate deviation: the TextLoader sealed-share is reverted there — it collides with #446's `TestTextLoaderEvaluationsGetPrivatePositions` and had no identified consumers; the reasoning is recorded in `loader.go`. Branch preserved for reference.

---

PR #388, opened 2026-08-12
lisp: copy builtin + checked-mode CoW counter (stacked on #377)

### LINKED ISSUE #378: Sealed-value semantics: decide copy-on-write vs hard error per site; expose sealed? and copy to lisp
Follow-up to the seal design (PR #374, docs/sealed-ast.md). Two related items:

## 1. CoW vs hard error (the Racket question)

Today a mutation attempt on a sealed value copy-on-writes silently (e.g. `stable-sort` on a quoted literal sorts a fresh copy and leaves the literal untouched). Racket/R7RS would raise an error instead. The CoW choice preserved compatibility, but it bifurcates semantics: the same expression behaves differently depending on whether its input is sealed. A silent copy can also mask a real bug (code that *intended* in-place mutation of what it wrongly assumed was its own list).

Decision procedure, empirical rather than aesthetic:
- Add a checked-mode counter/log on every CoW-on-sealed event (site + lisp stack).
- Run both repos' full suites and the production phylum corpus under the tag.
- Zero events in production code paths → flip the CoW sites to a catchable `cannot modify a program literal` condition (consistent with the elpspath list-reject and the never-panic policy #367); keep CoW only where an error is proven breaking.
- Nonzero → keep CoW at those sites, document each occurrence, and revisit the calling code.

The mechanism supports either semantics per site; this is a policy flip, not a redesign.

## 2. Lisp-visible contract: `sealed?` and `copy`

`IsSealed()` exists only in Go. Phylum authors cannot participate in the contract. Add:
- `(sealed? x)` — predicate over the seal bit.
- A real `copy` builtin (deep, fresh-backing, unsealed result — the lisp-level analog of `Detach`, generalizing the `(concat 'type x)` idiom that only covers one level). Documented as THE way to take ownership of data derived from program literals.

This moves the do-not-mutate contract into the language where the users live, instead of leaving it a Go-API footnote.

Related: #369, #373 (capacity-sharing decision), #367 (error-not-panic policy), docs/sealed-ast.md §4.

sam-at-luther:
Both halves of #378, minus the parts the maintainer discussion cut. Nothing is flipped: the census is evidence, the policy call stays yours.

## 1. `copy` — the only new lisp-facing surface

`(copy x)` is a **within-env ownership primitive**: a deep copy in which every data container (list, vector — including the `LArray` backing storage that `LVal.Copy` deliberately shares — sorted-map, bytes) is rebuilt with fresh backing, recursively, and **the seal is cleared** on every node of the result. Function and native leaves are **shared by reference**; strings and numbers are immutable values regardless; internal aliasing and cycles are preserved (a value reachable twice is copied once, a cycle becomes the same cycle in the copy).

It reuses the in-kernel deep-copy walker (`lisp/detach.go`) in a new `shareOpaque` mode rather than adding a second walker. It is deliberately **not** the lisp analog of `detach`/transfer: detach's rejection of `LFun`/`LNative` exists for cross-runtime transfer, which has no lisp-level use case. Within one env the environment a closure captured is the environment the copy runs in, lisp cannot mutate a function's internals (#382 unexported those fields), and natives are opaque — and rejecting them would make `(copy m)` fail on any map that happens to hold a callback. The three singletons are shared for the same reason: immutable, and unmutable from lisp.

Per the discussion there is **no `sealed?`** (its tempting use `(if (sealed? x) (copy x) x)` tests program-text provenance, not exclusive ownership — an unsealed value can still be aliased by another binding, container or closure) and **no `seal!`**. `IsSealed()` stays Go-side. Documented in `docs/func.md` and `docs/sealed-ast.md` §4.4; replaces the field workarounds (one-level `(concat 'type x)`, hand-rolled recursive copiers, json round-trips).

## 2. Checked-mode CoW census

All three copy-on-write-on-sealed sites now call `recordCoWOnSealed(env, site)`:

| site | what copies today |
|---|---|
| `stable-sort` | sorts a fresh header over a fresh backing array instead of sorting in place |
| `slice-vector` | `(slice 'vector …)` copies rather than wrap sealed backing in a mutable vector |
| `append-vector` | `(append 'vector …)` copies rather than write spare backing capacity |

Under `-tags elpscheck` the recorder counts the event, attributes it to the lisp-level frames that caused it, and prints each distinct (site, location) pair once to stderr — so a run driven by an external test binary leaves the evidence in its own log. `CoWOnSealedCounts`/`CoWOnSealedReport` expose the same data in-process. Untagged, `recordCoWOnSealed` is an empty inlined function.

### Results

Events, per site, per suite. "Corpus" is a production-scale phylum (28 lisp test files, 484 test cases), run with a `shirotester` built against this branch.

| suite | stable-sort | slice-vector | append-vector | provenance |
|---|---|---|---|---|
| `go test -tags elpscheck ./lisp/` | 24 | 4 | 9 | **all** from Go tests that construct sealed-input mutations on purpose (`cow_seal_test.go`, `sort_test.go`, the fuzz corpus, and this PR's own `copy_test.go` / `cow_counter_elpscheck_test.go`) |
| `go test -tags elpscheck ./lisp/x/debugger/` | 3 pairs | 0 | 2 pairs | debugger seed cases literally named `cow-stable-sort-literal`, `cow-append-vector-cdr`, … |
| every other elps package | 0 | 0 | 0 | |
| elps lisp-level suites (`*_test.lisp`, lisplib + examples) | 0 | 0 | 0 | |
| **production-scale phylum corpus (tagged, PASS)** | **0** | **0** | **0** | |

(The debugger row is distinct (site, location) pairs rather than totals: its seed cases are parallel subtests, so an in-process report at end-of-run sees them before they finish. Its events are named `cow-*` by hand either way.)

Anti-vacuity, using the *same* tagged binary that produced the zero: pointed at a two-line phylum whose test sorts a quoted literal, it prints

```
elps-cow: site=stable-sort loc=main_test.lisp:3:42: lisp:stable-sort <- main_test.lisp:3:19: lisp:let
```

so the corpus zero is a measurement, not a silent counter. The corpus also passes on the untagged (release-shaped) build with `copy` present.

### Untagged-cost proof

`go tool nm` over `go build .` with and without the tag:

| symbol / string | untagged | `-tags elpscheck` |
|---|---|---|
| `lisp.cowCheck`, `lisp.recordCoWOnSealed`, `lisp.cowLispLocation`, `cowEventKey` (eq/hash) | absent | present |
| site strings `slice-vector`, `append-vector`, log prefix `elps-cow:` | absent from the image | present |

`TestNoCoWCounterSymbolsUntagged` builds both binaries and asserts both directions, so the check cannot rot into a vacuous pass.

Interleaved benchmarks, `./lisp/`, n=10 alternating rounds, `-benchtime=100ms`:

**Census isolated** (copy-builtin-only commit vs this HEAD, both untagged) — this is the census's own cost:

```
sec/op   : 19/19 rows "~", geomean +0.75% (no row significant at p<=0.05)
B/op     : all rows "~"
allocs/op: all rows identical, including EnvInit (3.831k both arms)
```

**Whole branch vs base 4cab7aa** (both untagged): every deterministic metric identical except `EnvInit`, +0.38% allocs/op (3.817k → 3.832k, +15) and +0.28% B/op — that is the one extra `langBuiltin` entry and its formals for `copy`, present in both census arms above and therefore not the counter. sec/op rows are noise in both directions (`EnvFunCallRecursion` −12%, `EnvFunCallPos` −5%, both improvements at these sampling parameters, which `scripts/benchstat-gate.sh` documents as up to ±34% noise).

## 3. Flip recommendation (not applied)

**Recommend flipping all three sites to a catchable condition** — `cannot modify a program literal; take a (copy ...) first` — in a follow-up, on this evidence:

- Zero events in the production-scale phylum corpus, on a full passing run, with the counter proven live on the same binary.
- Zero events anywhere in elps outside tests written specifically to exercise the CoW paths. Nothing incidental reached a site.
- This matches the census in #378's first comment: every production sort input observed there was runtime-constructed (decoded JSON, `keys`, `make-sequence`), never a quoted literal.
- The remedy now exists and is one word. Before `copy`, an error would have been a dead end for a phylum author holding a literal-derived list; now the error text can name the fix, which is the whole reason to sequence the flip after this PR.

Scope note for whoever does it: the flip is per-site policy, not mechanism — each site already knows it is on a sealed value. Keep `stable-sort` and the two vector sites together (a partial flip re-bifurcates the semantics it is meant to unify), leave the metadata no-ops (`SetSource`, `stampMacroExpansion`) alone (they change no lisp-visible value), and land it early in a release cycle: the risk it carries is code nobody has run yet, which no counter can see.

## Test plan

- `go test ./...` — pass; `go test -tags elpscheck ./...` — pass.
- `go test -race ./lisp/` — pass.
- `golangci-lint run ./...` and `golangci-lint run --build-tags elpscheck ./lisp/...` — 0 issues.
- `elpsvet -test=false ./...` — exit 0. `elps fmt -l ./...` — clean. `elps lint` / `elps doc -m` — unchanged. `scripts/confidentiality-guard.sh` — clean.
- New: `lisp/copy_test.go` — deep-ness per container type (nested list/vector/sorted-map/bytes, and depth-4 nesting: mutate the copy at depth, assert the original pristine and the mutation landed); seal cleared against a shared cached parse using the `cow_seal_test.go` fingerprint (the copy sorts in place at depth while the cached AST stays bit-identical); function/native/map-value leaves shared **by identity**; shared substructure copied once; a self-referencing vector (cycle) copied without divergence, with the copy's cycle pointing at the copy; copy-of-a-copy independence; atoms and singletons; bytes backing disjoint.
- New: `lisp/cow_counter_elpscheck_test.go` — every site fires exactly once per event and moves only its own counter; the same expression over `(copy lit)` records nothing; the report names the lisp file and function.
- New: `lisp/cow_counter_test.go` — untagged behaviour (`CoWOnSealedCounts()` nil) and the `go tool nm` zero-cost proof with its anti-vacuity half.
- Production-scale phylum corpus: PASS tagged (evidence above) and PASS untagged.

Stacked on #377; chain: #374 → #377 → **#388** → #389 → #387 → #386 → #385. Post-release. Base branch is `claude/exp-vet-esc`, so the diff shown here is this PR's own work only.

Closes #378


sam-at-luther:
Superseded by #518 (collapsed with #377 and #389 into the verification layer of the rebuilt stack #517 → #518 → #519). Branch preserved for reference.

---

PR #389, opened 2026-08-12
lisp: two-env and container-family corruption fuzz targets (stacked on #388)

sam-at-luther:
**Draft.** Stacked on #388; chain: #374 → #377 → #388 → **#389** → #387 → #386 → #385. Base branch is `claude/exp-copy-builtin`, so the diff shown here is this PR's own work only.

Two fuzzing gaps against the corruption class tracked by #372, plus three findings the work turned up.

## Gap 1 — two-environment shared-parse differential

`lisp.FuzzSharedProgramMultiEnv`. Today's oracle (`FuzzEval`) fingerprints one sealed parse around **one** evaluation in **one** environment. The production hazard is a process-wide parse cache — substrate#375/#378 — and the property that topology needs has three separately-falsifiable clauses. The target parses **once** and evaluates the same `Program` in **2–3 independent environments, 2–3 times each**:

| # | Assertion |
|---|---|
| 1 | The shared sealed AST's fingerprint is unchanged after **every** evaluation in **every** environment — not once at the end. Asserted unconditionally. |
| 2 | Each environment's result sequence digests identically to a **freshly-parsed** copy in a virgin environment. |
| 3 | The environment's whole binding state afterwards equals the baseline's — every package, every symbol, through the sorted accessors. Strictly stronger than "E2 does not observe E1": it catches a leak that binds a symbol without changing any returned value. |
| 4 | Termination and no recovered Go panic (`IsInternalPanic`), inherited from the budgeted harness. |

**The re-evaluation dimension is load-bearing.** #369's `stable-sort` defect returns the *right* answer on the first call and the wrong one on the second, so neither a one-shot oracle nor a one-shot differential sees it.

**Determinism control.** A differential over fuzzer-written programs reports every `(time:utc-now)` as corruption unless it controls for that, and a blocklist of "nondeterministic builtins" goes stale the day one is added. So: baseline = fresh parse in a virgin env; control = a *second* fresh parse in a *second* virgin env. No differential claim is made when they disagree. A divergence that does appear is re-confirmed against another matched pair before it is reported, so the finding is attributed to the shared parse rather than to a flake. Assertion 1 needs no determinism and always runs.

## Gap 2 — structure-aware container-family targets

Three targets over an **operation sequence on a value pool** rather than independent one-shot calls, because the defects in these families are not in the argument types — they are in the relationship between two values (`slice` then `append`). Every result is fed back into the pool, so aliasing between step *i*'s output and step *j*'s input is the normal case.

| Target | Ops | What it asserts beyond the shared body |
|---|---|---|
| `FuzzSortedMapOps` | `assoc` `assoc!` `dissoc` `dissoc!` `get` `keys` `key?`, nested maps | A sorted-map lives in `Native`, so `SealedASTFingerprint` sees one opaque hole byte and every map-mutating defect is invisible to it. The new runtime-value fingerprint covers the map; the sealed oracle still covers literals stored *inside* it. |
| `FuzzSequenceOps` | `append` `append!` `append-bytes` `append-bytes!` `slice` `concat` `nth` `reverse` `stable-sort` `insert-index` `insert-sorted`, across list / vector / **bytes** | An explicit **capacity-sharing prelude**: build a slice of the base and keep *both* slice and source in the pool, so the #369/#373 three-step path is the starting point rather than a discovery. |
| `FuzzHigherOrderOps` | `map` `foldl` `foldr` `zip` `select` `reject` over **shared** structures | These builtins hand elements *out* (`elemCells[j] = seqCells(list)[i]`). Callbacks are adversarial by construction — each attempts an in-place `append!` / `stable-sort` / `assoc!` on what it receives. Sequences repeat the same `LVal` at several indices and share elements between each other. |

Shared body (`containerCall`): no panic escapes (applied through `env.FunCall`, never `env.Eval`, so eval's `recover()` cannot launder a crash), never a nil `*LVal`, the result renders, the three shared singletons are bit-identical, every sealed root in the pool is bit-identical after **every** step, and each argument the op's **docstring** does not license it to modify is bit-identical **by value**.

**Where the line is drawn — and it is #373's line, not a stricter one.** Writing through a shared backing array is the documented Go-slice semantics the maintainer decided to keep (`concat` is the documented copy idiom; copy-on-append would make vector building O(n²)). So whole-pool value stability is *not* asserted — that would report the language's own design on most inputs. What *is* asserted is the clause the seal exists for: **sealed** storage is bit-identical for every op including the mutating ones, and each op is held to its own docstring for its own arguments.

Two harness bugs the targets found in their own first minutes, both fixed: `(append-bytes! b b)` is an ordinary pool draw (an argument sharing storage with the mutated one is now exempt, computed from the object graph), and a caller-supplied write has no fixed target — a fold's callback writes the *accumulator* while a map's writes an *element* — so the higher-order family suppresses the value oracle whole-call rather than naming an argument.

`flatten` and `sort` are named in the brief but **do not exist in the language** (verified with `elps doc`); `stable-sort` is the only sort and is covered.

## Exec counts

All four ≥ 11 minutes, 2 workers each, run as two pairs on a 4-core box. **All clean, no crashers.**

| Target | Time | Execs | New interesting | Result |
|---|---|---|---|---|
| `FuzzSharedProgramMultiEnv` | 11m | 179,413 | 229 (421 total) | clean |
| `FuzzSequenceOps` | 11m | 1,094,879 | 269 (752 total) | clean |
| `FuzzSortedMapOps` | 11m | 1,095,995 | 40 (319 total) | clean |
| `FuzzHigherOrderOps` | 11m | 483,276 | 212 (512 total) | clean |

The shared-parse target is the slowest by an order of magnitude, and that is inherent: it builds four to five fully-loaded environments per input, which is what the property costs.

### 1. Uncatchable stack overflow rendering a self-referential container — reported, **fixed on `main` since**

Filed as #390 and **fixed by #391 (`050ade5`, merged to `main`)**: recursive value walks are now bounded, so a cyclic value can no longer kill the process, and `main` carries a `lisp.FuzzCyclicValueWalks` target. Everything below is the finding as this branch made it. **The fix is now ON this branch**, via the forward-merge of `main` into #374 that the merge-train pass performed.

```lisp
(set 'm (sorted-map))
(assoc! m "k" m)
m                        ; process dies
```

```lisp
(set 'v (vector))
(append! v v)
v                        ; equivalent
```

`LVal.String()` and its helpers (`exprString`, `sortedMapString`, `bodyStr`) recurse with **no cycle detection and no depth bound**. A Go stack overflow is a *fatal runtime error*, not a panic: `env.eval`'s `recover()` cannot intercept it and the process dies. In a phylum that is a chaincode crash from three lines of legal source.

Blast radius, measured: `LVal.String()`, `equal?` and `json:dump-string` all die; `length` and `reverse` survive. So this is a **class spanning several unguarded recursive walks**, and what a cyclic value should *print as* is a language decision — which is why it was reported rather than fixed here.

**A follow-on defect surfaced when that fix met the seal**, and is worth recording here because this branch's subject is what makes it legible. `main`'s `FuzzCyclicValueWalks` collects nodes with `containers()` and then writes through their `Cells` to tie knots. `containers()` walked *every* type, including `LFun` — whose cells are a function's formals and **body**, i.e. shared **sealed** parse nodes belonging to the standard library. So a knot could land in the stdlib's own AST. On `main` that silently corrupted the library mid-run and nothing could see it; on the sealing branch the teardown oracle failed the package with *"3 of 17625 sealed parse trees were mutated after parsing"* while every test still passed. Fixed on #374 by having `containers()` skip sealed nodes and their subtrees — the target keeps all its intended coverage, since everything it means to knot is unsealed runtime storage.

`containerRenderable` is the harness's response to finding 1: a step producing an unrenderable value abandons the sequence, so these targets decline to be the millionth reporter of one known defect. The bound is on *work*, not just cycles — a doubling chain builds an acyclic value whose rendered form is exponential. `TestContainerRenderableGuard` pins both directions, and the three `…SequencesRunToCompletion` gates measure that the guard is not quietly gutting coverage (97–100% of sequences run every step they asked for).

### 2. Checked mode forbade the parse-cache topology — found here, **fixed upstream on #374**

`go test -tags elpscheck ./...` with the Gap 1 target in the tree:

```
panic: ownership violation: LVal used by two Runtimes
  value: 0xc03e134fc0 type=list str="" cells=4 source=shared:1:1
  owner runtime:  0xc024f657a0 (package user)
  second runtime: 0xc024f65880 (package user)
```

Evaluating one `lisp.Program` in two `Runtime`s panics on the first shared AST node. But **that is the supported topology**: the `sealed` field's own doc comment says a sealed node "may be shared by every environment that evaluates the same parse — substrate's parse cache shares one tree process-wide". So the two halves of #372's verification tooling contradicted each other, and an embedder running the parse-cache topology could not use checked mode at all — the mode where the sealed-AST inspector lives.

**WHERE THE FIX NOW LIVES.** The exemption commit was moved to `claude/exp-seal-tooling` (#374) during the merge-train pass, and this section no longer claims it lands here. It is not this branch's defect to own: the panic fires on #374, #377 and #388 as well, *before and independently of anything this branch adds*, because sealing ships from #374 — so the ownership checker's understanding of sealing has to ship from #374 too. Left where it was, every PR between #374 and here carried a checked mode that rejected its own design. What stays with this branch is the **discovery** and `FuzzSharedProgramMultiEnv`, the target whose whole subject is that topology and which is what surfaced it; the code comment upstream says exactly that, and defers the "found by" provenance to this branch, where the target actually exists.

`ownership_check_elpscheck.go`'s own rule is explicit: *"do not add an exemption reflexively: it is either a real bug (fix it, or file it like #363) or a deliberate design (document it HERE, with the reasoning)."* This is the second, and the Allowlist section carries the full reasoning as entry 2. What licenses it: a sealed node's cross-runtime safety does not rest on ownership at all — sealed bytes never change after parse, enforced by copy-on-write at every mutation site, by the fingerprint oracle and by the -race watchdog. Ownership stays the right question for **mutable** storage, and that stays fully checked: `Copy` and `detach` clear the flag on fresh storage.

Both directions remain pinned, and the red-proofs moved with the commit — they are still the evidence for the exemption's narrowness: `TestOwnershipCheck_SealedNodesExempt` (a sealed node crosses runtimes; the same shape unsealed still panics) and `TestOwnershipCheck_CopyOfSealedIsChecked` (if `Copy` ever preserved the flag, the exemption would have silently widened to cover runtime values). All nine existing `TestOwnershipCheck_*` still pass. Checked builds only; `ownership_check_default.go` is a no-op, so release builds are untouched.

Two branches above this one in the chain reached the same exemption independently — #387 from the shared benchmark program (#379 item 2) and #386 from `LEnv.Fork` (#380). With the base copy now upstream of all three, those collisions resolve against a common ancestor instead of against each other.

### 3. `FuzzPathEngine` is not on this stack

The structure-aware elpspath target that found the #379 list write-back in six seconds lives on commit `974dea6` (branch `claude/exp-seal-opts`) and is **not** on `claude/exp-seal-tooling` or here — `lisp/lisplib/libelpspath/path_copy_list_test.go` references it in a comment, but `fuzz.sh --list` does not find it. The elpspath family is therefore currently covered only indirectly, by `FuzzApplyStdlib`'s generic value seeds.

The chain resolves this by ordering rather than by cherry-pick: `claude/exp-seal-opts` is #387, which is now stacked directly on this branch, so `FuzzPathEngine` joins the sweep one link up.

## CI capacity — **re-derived at this head**

`scripts/fuzz.sh` discovers targets dynamically via `go test -list`, so new targets need **no registration anywhere**.

**The numbers here have moved twice since this PR was opened, and both moves were the same hazard.** Originally measured on `claude/exp-seal-tooling`: 18 targets / 5 shards / largest shard 4 / headroom 10 min. Adding this branch's four took it to 22, where `ceil(22/5) = 5 × 10m + 10m = 60` against a 60-minute timeout — **zero headroom**. `fuzz-budget-check`'s condition is `needed > timeout`, so that *passes*, and a sweep sized at exactly its own backstop is the silently-truncated run the gate exists to prevent (observed shard durations already vary 4m25s–5m42s). So, at the time: a **sixth shard**.

**SUPERSEDED — it is a SEVENTH shard now.** The forward-merge of `main` into #374 (owed since #356/#357/#391 landed) brought three more targets onto the chain, and **#374 took the sixth shard for itself** to absorb them at 21 targets. This branch's four then take the sweep to **25**, where six shards measure *largest shard 5 / required 60 / timeout 60 / headroom 0* — the passing-but-truncated state, reached a second time and from the other direction: not by a merge landing three targets at once, but by a branch adding exactly the targets it exists to add.

Re-derived at this head: **25 targets / 7 shards / largest shard 4 / required 50 vs timeout 60 / headroom 10 min**, verified against the real partition — `4+4+4+4+3+3+3 = 25`, every target assigned exactly once. Seven holds through 28 targets; the 29th is the next one to watch, and `claude/fix-394-program-seal` and `claude/fix-393-395-libelpspath` are both adding test files.

Two related fixes landed on #374 rather than here, both from the same family:

- The shard denominator passed to `fuzz.sh` is now **derived** from the matrix (`strategy.job-total`) instead of being a literal. A hardcoded `/5` against a `[1..6]` matrix had shards 1–5 silently re-running a *five*-way partition while shard 6 exited 2 with "shard 6/5 is out of range" — and `fuzz-budget-check`, which reads the count from the matrix and never looked at the run line, reported "6 shards / headroom 10 min" for a sweep that was running five ways.
- `scripts/ci-gates-test.sh` now fails the PR if either the run line or the job label goes back to a literal, red-proofed in both directions.

#358 adds a sixth shard on the `main` line for the same underlying reason, so the forward-merge of `main` will meet a `shard:` matrix that differs only in length, not in shape.

## Gates

| Gate | Result |
|---|---|
| `go test ./...` | pass |
| `go test -tags elpscheck ./...` | pass |
| `make elpsvet` (both passes) | exit 0 |
| `elps fmt -l ./...` | clean |
| `scripts/confidentiality-guard.sh` | clean |
| `scripts/fuzz-budget-check.sh` | 25 targets / 7 shards, largest shard 4, headroom 10 min |
| `bash scripts/ci-gates-test.sh` | 133 passed, 0 failed |

## What remains UNCOVERED

Stated plainly, because a coverage claim without its complement is worth little.

1. **Concurrent** shared-parse evaluation. Environments here run one after another — that is the substrate topology and it keeps a crasher reproducible. Two goroutines evaluating one `Program` *simultaneously* is a different property; the -race seal watchdog is the mechanism, but no fuzz target drives it.
2. **Same-value metadata writes.** `SealedASTFingerprint` hashes `source` *contents*, so re-storing an identical location is invisible to it (documented in `sealfp.go`). Still a data race on shared storage; still the watchdog's to catch, and unfuzzed.
3. **Cyclic-value rendering** (finding 1) — deliberately excluded from these targets. The defect itself is fixed by #391, which is now on this branch; these targets keep `containerRenderable` because the work bound it also imposes (exponential acyclic renderings) is independent of the cycle fix.
4. **`LBytes` is outside the seal by construction.** `SealAST` marks only parser-producible shapes and no literal is bytes, so the sealed oracle can say nothing about a byte buffer; the value oracle is the entire guarantee for that half of the sequence family.
5. **Multi-dimensional `LArray`.** The generator builds 1-dim vectors only. `slice` / `insert-index` against `dims > 1` is reached only by `FuzzApplyStdlib`'s `fuzzval`.
6. **The `&rest key-fun` variants** of `stable-sort`, `insert-sorted` and `search-sorted`, entirely.
7. **`compose` / `unpack` / `flip`**, and `LTaggedVal` / `deftype` values through any container op.
8. **elpspath**, per finding 3 — resolved one link up the chain rather than here.
9. **Stdlib-constructed containers** reaching these ops — a nested map from `json:load-bytes` handed to `assoc!`, say — only via `FuzzApplyStdlib`'s generic seeds.
10. **The evaluator's own quote and macro aliasing paths into these builtins.** The container targets call through `FunCall` with constructed values; source-level reach is covered only by `FuzzEval` and the Gap 1 target.
11. **`Program.detach()`** — no target checks that a detached copy evaluates identically to the sealed original (it is unexported, with no consumer yet).


sam-at-luther:
Superseded by #518 (collapsed with #377 and #388 into the verification layer of the rebuilt stack #517 → #518 → #519). Branch preserved for reference.

---

PR #391, opened 2026-08-12
lisp: bound recursive value walks so cyclic values cannot kill the process (#390)

### LINKED ISSUE #390: Cyclic values kill the process: unbounded walks in String/equal?/json:dump are fatal, not catchable
Found by the new container-family fuzz targets (PR #389) in about 7 seconds. Pre-existing on `main`; not a regression from the sealing work.

## Repro (32 bytes of lisp)

```lisp
(set 'm (sorted-map))
(assoc! m "k" m)
m
```

The process dies. `LVal.String()` has no cycle detection and no depth bound, so rendering a self-referential value recurses until the Go stack overflows — and a stack overflow is a fatal runtime error that `recover()` cannot catch, so the evaluator's panic handling cannot convert it into a condition.

Same shape kills `equal?` and `json:dump-string`, and `(append! v v)` on a vector is equivalent, so this is a class across several unguarded walks rather than one bad function.

## Why it matters beyond a crash

- It violates the "an elps program must never panic the host" invariant (#367) in its strongest form: not an internal-panic surfaced as a condition, but an unrecoverable process kill.
- For an embedder running many tenants' programs in one process, any program that constructs a cycle and prints it takes down the process.
- **The existing `FuzzEval` target is exposed to it today** (verified by hand), so CI can hit it.

## What needs deciding

What should a cyclic value print as? The options are a depth/visited-set bound in the shared walk (rendering something like `#<cycle>`), or making the cycle-creating mutators reject self-insertion. The first is cheaper and covers `equal?`/`json:dump` too; the second changes documented in-place semantics. Whichever is chosen, the fix belongs in the shared walk so all consumers inherit it.

Related: #367 (never-panic sweep), #372 (verification tooling — this was found by the new targets), #389.

sam-at-luther:
## The repro

```lisp
(set 'm (sorted-map))
(assoc! m "k" m)
m
```

Thirty-two bytes, and the host process dies:

```
fatal error: stack overflow
```

`(*LVal).String` walks the value with no bound, so rendering one that contains itself recurses until the goroutine stack passes the runtime's limit. A stack overflow is a **fatal runtime error**: `recover()` cannot catch it, so the evaluator's panic handling — which turns every internal panic into a condition — is powerless, and the "an elps program must never panic the host" invariant (#367) fails in its strongest form. `equal?`, `json:dump-string` and `(append! v v)` on a vector all reach the same end, and `FuzzEval` is exposed to it today.

Pre-existing on `main`; not a regression from the sealing work.

## The guard

`lisp/cycle.go` adds a guard that a recursive walk carries down with it, in three stages, so that acyclic values pay for none of it:

| Stage | When | Cost |
|---|---|---|
| 1 | always | an `int` in a by-value struct: increment, compare. No allocation. |
| 2 | nesting past `cycleGuardDepth` (64) | a set of the nodes on the walk's **current path**, allocated on first use. A node already on the path is a cycle. |
| 3 | a cycle was found | the first frame to see it records that on state shared by the walk, every frame above returns at once, and the entry point reruns the walk in **strict** mode: the path set is live from the first frame and nothing leaves it, so each node is visited once. |

Stage 2 is path-scoped and not walk-scoped on purpose: `(list x x)` shares a substructure with itself and is a DAG, not a cycle, and it must still render as the two copies it has always been.

**Stage 3 is not an optimisation.** Stages 1 and 2 bound a walk's *depth* and not its *width*: a map holding itself under two keys is one node deep in the cycle, but unrolling it to 64 levels visits 2^64 nodes. A fix that swapped a crash for a walk that never finishes would not be a fix, and reverting stage 3 is red-proofed below.

The state object is addressed, not returned, so the guards stay pass-by-value and the walks keep their signatures; no frame stores that pointer anywhere but its own stack, so escape analysis keeps it off the heap. `TestGuardedWalksDoNotAllocate` pins that.

## Walk inventory

Every recursive walk over `*LVal` in the repository, and what happened to it.

| Walk | Reachable from lisp | Before | Disposition |
|---|---|---|---|
| `(*LVal).String` → `str` / `strNested` / `exprString` / `bodyStr` / `sortedMapString` | yes: printing a result, `format-string`, `debug-print`, and every error message that formats an operand | **fatal** | guarded; renders `#<cycle>` |
| `ErrorVal.Error` / `baseMessage` / `ErrorMessage` / `errorCellMessage` | yes: an error's cells are rendered as ordinary values | **fatal**, and it defeats a guard placed only in `str` — each cell started a *fresh* `String()` walk, resetting the bound on every lap | the guard is threaded through, so the walk continues instead of restarting |
| `(*LVal).Equal` / `equalMapKey` — `equal?` | yes | **fatal** | guarded; co-inductive (below) |
| libjson `encoder.encode` + `encodeSExpr` / `encodeSortMap` / `encodeArray` / `encodeTaggedVal` / `encodeLQuote` — `json:dump-string`, `dump-bytes`, `dump-message` | yes | **fatal** | guarded; refuses with a catchable elps error |
| `stampMacroExpansion` | yes: a macro may build its expansion with `assoc!`/`append!` — `(defmacro cyc () (let ([x (vector 1)]) (append! x x) x))` kills the process on `main` | **fatal** | guarded; stamping is idempotent, so the strict rerun finishes what the abandoned walk left |
| `lisp.GoValue` / `GoSlice` / `GoMap` / `checkGoMapInsert` | **no** — no callers inside the repository; exported for embedders | **fatal** | guarded. An embedder converting a value a tenant's program built is exactly the case #390 is about. A cyclic value has no Go representation, so `GoValue` hands back the `*LVal` (already what it does for values with no natural Go form) and `GoSlice`/`GoMap` report "not representable". |
| `(*LVal).Copy` / `copyCells` / `copyMapData` | yes: `stable-sort`, `insert-sorted`, `zip`, builtin registration | **already terminates** | **not guarded.** A cycle a program can build must pass through a sorted-map or an array, and `Copy` descends into neither: map data is rebuilt entry by entry sharing the value pointers, and an array keeps its backing storage by reference. The deep copy stops one level into the cycle by itself. `TestCopyOfCyclicValueTerminates` pins the reasoning so that a future change to either container is caught here. |
| `debugger.FormatValue` / `formatList` | yes (debugger UI) | **already terminates** | not guarded: it renders maps, arrays and tagged values as summaries (`<sorted-map len=3>`) without descending, and every cycle passes through one of those. |
| `(*LVal).Len` | yes | bounded (descends at most into an array's dimensions) | not guarded |
| `libjson.Serializer.GoValue` / `GoSlice` / `GoMap` | **no** — `Deprecated`, and dead: nothing in the repository calls it | unbounded | **not guarded**, deliberately: guarding dead deprecated code is diff for nothing. Its doc comment now says it is unbounded and points at `Dump`. |
| `analysis/`, `lint/`, `lsp/`, `formatter/`, `minifier/`, `astutil/` | — | n/a | not guarded: these walk *parse trees*. The parser only ever builds fresh nodes, so its output is a tree, and runtime values do not enter these walks. |

## The three decisions the issue left open

**Rendering: `#<cycle>` at the revisit point.** It keeps the output finite, keeps it readable — `(sorted-map "k" #<cycle>)` says exactly what the value is — and it is in the house style for values that cannot be written down (`#<builtin>`, `#<native value: T>`, `#<array dims=...>`). Recorded on `String`'s doc comment and in `lisp/cycle.go`.

**Equality: co-inductive, and documented as such.** When the comparison reaches a *pair* it is already comparing further up the path, it takes that pair to be equal and moves on. That is the greatest fixed point — equality of the two operands' infinite unfoldings, which is what R7RS requires of `equal?` on circular structure — and it is an answer rather than a guess: `false` is only ever returned for a difference actually found at a finite depth, so no equality is claimed that a longer walk could refute. `(equal? a b)` on two *distinct* maps that each hold themselves is `true`; add a key to one and it is `false`.

The pair set is never unwound, and that is not an approximation either: a pair reached a second time is either still under comparison (the co-inductive case) or has already been compared and found equal, since a pair that compared *unequal* returned `false` out of every frame up to the root instead of ever being reached again.

**JSON: a catchable error, never a partial document.** JSON has no representation for a cycle, so `json:dump-*` refuses:

```lisp
(handler-bind ([error (lambda (c e) "caught")])
  (json:dump-string m))    ; => "caught"
```

The error aborts the encode at the point of detection, so nothing exponential runs first, and no bytes are emitted.

## The rejected alternative

The issue's other option was **rejecting self-insertion in the mutators** (`assoc!`, `append!`). Not taken, for two reasons:

1. It changes documented in-place semantics. `assoc!` and `append!` are specified to mutate and return the container; making them error on one particular argument is a language change, and one that silently breaks programs that build a self-referential structure deliberately and never print it.
2. It does not cover the class. `(assoc! m "k" m)` is only the shortest cycle. `(assoc! m "k" inner)` where `inner` already holds `m` is not self-insertion, and neither is any cycle assembled by host code through the `lisp` package's exported constructors. A mutator check would have to walk the inserted value on every `assoc!` — which is a walk over `*LVal`, needing its own bound, on the *hot* path this time.

The guard belongs in the shared walk, where every consumer inherits it. The mutators are unchanged.

## Byte-identical output for acyclic values

The guard is invisible to anything that is not cyclic, at any depth, and that is tested rather than asserted:

- `TestAcyclicRenderingIsUnchangedBelowAndAboveTheGuard` renders values nested 1, 63, 64, 65 and 256 levels deep and compares against independently computed expectations, plus a DAG that shares a node *below* the guard depth — both occurrences must still render in full.
- `TestAcyclicEqualityIsUnchangedBelowAndAboveTheGuard` does the same for `Equal`, with the negative cases carrying the weight: a guard that reported equality whenever it ran out of patience would pass the positive ones.
- `TestEncodeAcyclicValueIsUnchangedBelowAndAboveTheGuard` for JSON.
- The existing suites (`lisp`, `elpstest`, `libjson`, the `.lisp` corpus) contain the broad value corpus and are unchanged, which is the strongest statement of all: no expectation anywhere in the repository moved.

## Red-proof

Each guard was reverted individually and the corresponding test re-run.

| Reverted | Result |
|---|---|
| `str` | `TestStringOfCyclicValue`: **fatal error: stack overflow**, the test binary dies |
| `str` (subprocess assertion) | `TestCyclicValueDoesNotKillTheProcess/render` FAILs with the helper's `fatal error: stack overflow` in the captured output |
| `equal` | `.../equal` FAILs, helper died: stack overflow |
| `stampMacroExpansion` | `.../stamp` FAILs, helper died: stack overflow |
| `GoValue` | `.../govalue` FAILs, helper died: stack overflow |
| libjson `encoder.enter` | `TestEncodeCyclicValueErrors`: **fatal error: stack overflow** |
| `errorCellMessage` guard threading (back to `cell.String()`, i.e. a guard in `str` only) | `.../render-error` FAILs, helper died: stack overflow — this is the case a guard placed only in `str` misses |
| **stage 3** (the escalation to strict mode) | `.../render-wide` does **not terminate**: killed at 90s. This is the 2^64 unrolling, and it is why stage 3 exists. |

`TestCyclicValueDoesNotKillTheProcess` runs each walk in a **helper subprocess** and asserts on its exit status and output. That is the honest way to assert "the process survived" when the failure mode is a process kill: no in-process assertion survives a stack overflow, as the first row of that table shows. The helper lowers the stack limit to 16MB so an unguarded walk fails in milliseconds instead of spending a second and a gigabyte first.

## Fuzzing

`FuzzCyclicValueWalks` (new, `lisp/cycle_fuzz_test.go`) closes the gap that let this sit on `main`: every existing value-shaped target (`FuzzApplyStdlib`, `FuzzDumpJSON`, `FuzzSchemaValidate`) builds its subject with `internal/fuzzval`, which constructs **trees** — so no target in the repository had ever handed a walk a value with a cycle in it. The new target generates a value, ties a fuzzer-chosen set of knots in it, and runs `String`, `Equal` (both directions, against itself and against a second knotted value), `GoValue` and `json.Dump` under the scheduled-time watchdog (`internal/fuzzwatch`), asserting termination, no panic, deterministic rendering and unmutated singletons.

```
fuzz: elapsed: 11m0s, execs: 429712 (0/sec), new interesting: 394 (total: 883)
PASS
ok  	github.com/luthersystems/elps/lisp	661.054s
```

**11 minutes, 429,712 execs, clean**, from 77 seeds to an 883-input corpus. No crasher, so `testdata/fuzz/` gains nothing — which is the point: the target is new coverage, not a regression case. Reverting any guard and rerunning it kills the worker with `fatal error: stack overflow` within seconds.

## Cost

`lisp/render_bench_test.go` adds the rendering and equality benchmarks this change needed and the repository did not have. Interleaved runs (base/new alternating, `-count=1 -benchtime=0.2s` × 10, n=10) on `origin/main` + the same benchmark file:

```
goos: linux
goarch: amd64
pkg: github.com/luthersystems/elps/lisp
cpu: Intel(R) Xeon(R) Processor @ 2.10GHz
                       │     base     │                 new                  │
                       │    sec/op    │    sec/op     vs base                │
String/int-4             12.63n ±  3%   15.53n ±  5%  +22.91% (p=0.000 n=10)
String/string-4          364.9n ±  7%   359.8n ± 12%        ~ (p=0.853 n=10)
String/list-4            323.2n ±  6%   322.1n ±  7%        ~ (p=0.592 n=10)
String/wide-4            1.824µ ±  8%   1.722µ ±  3%        ~ (p=0.089 n=10)
String/map-4             2.319µ ±  8%   2.236µ ±  7%        ~ (p=0.353 n=10)
String/vector-4          2.779µ ± 10%   2.537µ ± 14%        ~ (p=0.143 n=10)
String/nested-4          5.000µ ± 11%   4.978µ ±  7%        ~ (p=0.971 n=10)
String/deep-4            8.663µ ±  9%   8.787µ ± 17%        ~ (p=0.631 n=10)
Equal/int-4              3.649n ±  4%   5.686n ±  5%  +55.80% (p=0.000 n=10)
Equal/string-4           4.717n ±  3%   7.693n ±  4%  +63.08% (p=0.000 n=10)
Equal/list-4             27.24n ±  4%   33.33n ±  6%  +22.34% (p=0.000 n=10)
Equal/wide-4             291.4n ±  4%   325.1n ± 11%  +11.58% (p=0.000 n=10)
Equal/map-4              2.317µ ±  9%   2.360µ ± 29%        ~ (p=0.796 n=10)
Equal/vector-4           2.252µ ±  9%   2.272µ ± 25%        ~ (p=0.971 n=10)
Equal/nested-4           2.507µ ± 12%   2.483µ ± 22%        ~ (p=0.684 n=10)
Equal/deep-4             641.2n ±  4%   777.5n ±  3%  +21.25% (p=0.000 n=10)
EqualDistinct/int-4      3.385n ±  6%   5.822n ±  6%  +71.98% (p=0.000 n=10)
EqualDistinct/string-4   4.705n ±  4%   7.568n ±  3%  +60.86% (p=0.000 n=10)
EqualDistinct/list-4     27.28n ±  5%   32.47n ±  7%  +19.02% (p=0.000 n=10)
EqualDistinct/wide-4     288.4n ±  6%   323.6n ±  4%  +12.20% (p=0.000 n=10)
EqualDistinct/map-4      2.427µ ± 10%   2.314µ ± 30%        ~ (p=0.089 n=10)
EqualDistinct/vector-4   2.256µ ± 17%   2.322µ ± 19%        ~ (p=0.853 n=10)
EqualDistinct/nested-4   2.528µ ± 16%   2.588µ ± 10%        ~ (p=0.739 n=10)
EqualDistinct/deep-4     643.8n ±  5%   757.3n ±  2%  +17.63% (p=0.000 n=10)
geomean                  315.8n         357.0n        +13.04%
```

**B/op and allocs/op are identical on every single benchmark** — `~ (p=1.000)` with "all samples are equal" for all 24 — which is the claim that matters: the guard allocates nothing on the common path, at any depth up to `cycleGuardDepth`.

Everything with real content is unchanged: maps, vectors, nested values, wide lists, and rendering generally. What is left is a constant, readable straight off the table in absolute terms:

- **+2.9ns on `String` of an int**, and only an int — the walk's entry bookkeeping (a `cycleState` on the stack and a guard to carry down), measured against a 12.6ns baseline that does nothing but `strconv.Itoa`. Every other rendering benchmark is noise.
- **+2.0–3.0ns on `Equal` of two leaves**: one call frame, because `Equal` is now a shell that can rerun the comparison. `Equal/list` (3 cells) is +6ns and `Equal/wide` (64 cells) is +34ns, i.e. ~0.5ns per element — not per element per level.
- **~+4ns per nesting level** (`Equal/deep`, 32 levels, +136ns): one `descend` per container.

For scale, `equal?` reached from lisp goes through the evaluator's builtin dispatch first, which is hundreds of ns; 2–3ns of guard entry is not visible there. I did try to remove even this — the first cut put the guard at the top of both walks and cost +33% / +117% on the same benchmarks; moving it off the leaf path (commit "keep the guard off the leaf path") is what produced the table above. Removing the last few ns would mean duplicating the render and compare switches, which is a worse trade than the ns it buys.

`BenchmarkEncode` (libjson, pre-existing) was run in the same interleaving and is unchanged.

## Gates

| Gate | Result |
|---|---|
| `go test ./...` | ok |
| `go test -tags elpscheck ./...` | ok |
| `go test -race ./lisp/` | ok (51.7s) |
| `golangci-lint run ./...` | `0 issues.` |
| `./elps fmt -l ./...` | clean |
| `./elps lint --workspace=. --exclude grammar --include _examples ./...` | exit 0 (the shape CI runs) |
| `./elps doc -m` | `All functions and exports are documented.` |
| `bash scripts/confidentiality-guard.sh` | `confidentiality guard: clean` |
| `go run ./cmd/elpsvet -test=false ./...` | **not runnable here**: `cmd/elpsvet` does not exist on `main`. It arrives with the seal-tooling work; the equivalent gate on this base is `./elps lint`, above. |

The diff is 14 files, all of them the walks named in the inventory plus their tests. An earlier commit on this branch had `gofmt -w lisp/` reformatting files this change never touches; that churn is reverted, so nothing outside the fix moves.

## Merge order

Trial-merged against `claude/exp-seal-tooling` with `git merge-tree` (not merged). Two files conflict, both mechanically:

- **`lisp/error.go`** — that branch rewrites `Error()` to use the new `(*LVal).Source()` accessor and drops the `e.Source != nil` branch; this one threads a `cycleGuard` through `Error`/`baseMessage`/`ErrorMessage`/`errorCellMessage`. Resolution: keep that branch's bodies, re-apply the guard parameter.
- **`lisp/macro.go`** — both edit `stampMacroExpansion`; that branch adds the `v.sealed` early return and renames `Source`/`MacroExpansion` to unexported fields. Resolution: keep its body, wrap it in this branch's `stampGuarded` shell, with the `sealed` check ahead of the guard.

`lisp/lisp.go`, `lisp/maps.go`, `lisp/embed.go` and the libjson files auto-merge (that branch's `Quoted` → `quoted` rename lands on different lines from this change), but the merged result should be compiled rather than trusted.

Closes #390

---

### DIFF
--- a/.github/workflows/fuzz.yml
+++ b/.github/workflows/fuzz.yml
@@ -87,8 +87,10 @@ jobs:
     #     ceil(targets / shards) x FUZZTIME + overhead <= timeout-minutes
     #
     # on every PR, so this number cannot silently stop covering the sweep. At
-    # 17 targets over 5 shards that is 4 x 10m + 10m = 50; 60 leaves a shard's
-    # worth of headroom for a target being added before anyone reshards.
+    # 18 targets over 5 shards that is 4 x 10m + 10m = 50; 60 leaves a shard's
+    # worth of headroom for a target being added before anyone reshards. That
+    # ceiling holds to 20 targets; the 21st takes the largest shard to 5 and
+    # the headroom to zero, which is the moment to add a sixth shard.
     #
     # THE HEADROOM IS THE POINT, NOT THE GREEN TICK. The gate's condition is
     # `needed > timeout`, so a sweep sized at exactly timeout-minutes PASSES --
--- a/elpsutil/elpsutil_fuzz_test.go
+++ b/elpsutil/elpsutil_fuzz_test.go
@@ -1,0 +1,1589 @@
+// Copyright © 2026 The ELPS authors
+
+package elpsutil_test
+
+import (
+	"context"
+	"fmt"
+	"sort"
+	"testing"
+	"time"
+
+	"github.com/luthersystems/elps/elpsutil"
+	"github.com/luthersystems/elps/internal/fuzzseed"
+	"github.com/luthersystems/elps/internal/fuzzwatch"
+	"github.com/luthersystems/elps/lisp"
+	"github.com/luthersystems/elps/lisp/lisplib"
+	"github.com/luthersystems/elps/parser"
+)
+
+// Fuzzing the EMBEDDING SURFACE: the path a Go program takes to install its own
+// packages, builtins, special operators and macros into an interpreter, and
+// then to run lisp against them.
+//
+// # Why this package
+//
+// Issue #318 lists elpsutil/ as unfuzzed. It is 193 lines and looks trivial,
+// which is exactly why it was left last -- and it is the wrong one to leave
+// last. luthersystems/substrate imports github.com/luthersystems/elps/elpsutil
+// directly and wraps it in internal/substrate/elpsutil/elpsutil.go, so this is
+// the code that installs every chaincode-visible Go builtin. Nothing else in
+// this repository stands between a substrate package definition and the
+// interpreter's symbol table.
+//
+// The measured starting point was worse than "unfuzzed". Running the WHOLE
+// repository test suite with -coverpkg=./elpsutil reported 6.2% of statements:
+//
+//	Function            100.0%     <- incidentally, from debugger + FuzzEval
+//	Name/Formals/Eval   100.0%     <- same
+//	nopLoader             0.0%
+//	packageInit           0.0%
+//	packageBuiltins       0.0%
+//	packageSpecialOps     0.0%
+//	packageMacros         0.0%
+//	Load                  0.0%
+//	LoadAll               0.0%
+//	LibraryLoader         0.0%
+//	PackageLoader         0.0%
+//
+// The only part of elpsutil the suite touched was the one-line Function
+// constructor, used by three tests that want a throwaway builtin. The entire
+// package-loading surface -- the part substrate actually calls, as
+// `elpsutil.Load(env, elpsutil.PackageLoader(&substrate.CorePackage{}))` --
+// had never executed under test.
+//
+// This target's SEED CORPUS ALONE, with no fuzzing at all, takes that to 96.9%
+// (measured: go test -run FuzzElpsutilEmbed -coverpkg=./elpsutil). Twelve of
+// the thirteen functions reach 100%. What remains is two blocks in
+// PackageLoader:
+//
+//	e := env.DefinePackage(name);  if !e.IsNil() { return e }
+//	e = env.InPackage(name);       if !e.IsNil() { return e }
+//
+// Neither is reachable. DefinePackage errors only when its argument is neither
+// LSymbol nor LString, and the argument is lisp.Symbol(p.PackageName()), which
+// is an LSymbol by construction. InPackage errors only when the package does
+// not exist, and the line above just defined it. They are defensive branches
+// against arguments elpsutil builds itself, so no input can reach them and no
+// seed pretends to -- stated here rather than papered over with a stub that
+// would cover the lines without testing anything.
+//
+// # What is fuzzer-controlled, and why each thing is
+//
+// Two inputs, mutated independently so any package shape can be paired with
+// any program:
+//
+//	spec -- a byte program describing what to INSTALL: how many packages, what
+//	        they are named, which of the four optional interfaces each one
+//	        implements, what its PackageInit does, and the name/formals/body of
+//	        every builtin, special operator and macro it registers. Also which
+//	        of the four loader entry points combines them.
+//
+//	src  -- ordinary elps source, evaluated after the install, so the
+//	        registered functions are CALLED with arguments the fuzzer chose.
+//
+// Justification, item by item:
+//
+//   - PACKAGE NAMES are fuzzer-controlled because a name is not inert. It is
+//     the registry key (DefinePackage/InPackage), it is embedded in every FID
+//     string, it is the left half of qualified symbol resolution `pkg:sym`,
+//     and it is what GetFunName renders into error messages. An embedder picks
+//     it, so a defect keyed on a name with a ':' in it, an empty name, or a
+//     name that is not valid UTF-8 is an embedder-reachable defect. Names are
+//     drawn EITHER from a table of interesting constants OR straight from
+//     spec bytes; see specReader.name.
+//
+//   - WHICH OPTIONAL INTERFACES A PACKAGE IMPLEMENTS is fuzzer-controlled
+//     because that is the whole of packageInit, packageBuiltins,
+//     packageSpecialOps and packageMacros: each is a two-branch type
+//     assertion, and the !ok branch of all four (plus nopLoader, which only
+//     the init one can reach) is dead code unless some package declines to
+//     implement them. Four concrete Go types cover the lattice; see pkgTypes.
+//
+//   - FORMALS are fuzzer-controlled because elpsutil.Function takes an
+//     arbitrary *LVal and performs NO validation on it, and lisp.FunInPackage
+//     performs none either. Whatever an embedder passes is stored verbatim in
+//     Cells[0] and is not looked at again until call time, deep inside
+//     LEnv.bind. The formals table therefore includes shapes an embedder can
+//     plausibly produce by accident -- an LError (which is what
+//     lisp.Formals("&rest") returns), a non-QExpr scalar, a list whose cells
+//     are not symbols, duplicate names, a misplaced control symbol -- because
+//     each of those reaches argument binding as data.
+//
+//   - ONE SHARED FORMALS VALUE, reused across several registrations, is a
+//     formals shape of its own (formalsShared). elpsutil.Builtin.Formals
+//     returns fun.formals by POINTER, and env.builtin hands that same pointer
+//     to FunInPackage, so registering one *Builtin into two packages gives two
+//     LFuns aliasing one formals LVal. lisp.bind documents that it "does not
+//     modify fun or args"; this is the input that would notice if that ever
+//     stopped being true.
+//
+//   - BUILTIN BODIES are fuzzer-controlled, from a table, because the body is
+//     the part substrate writes. The interesting ones re-enter the
+//     interpreter (env.Eval, env.LoadString) or mutate interpreter state
+//     (env.InPackage) from inside a call -- which is what a real Go builtin
+//     does, and which no other fuzz target in this repository does at all.
+//     FuzzEval registers exactly one builtin, `fuzz-panic`, whose body is a
+//     bare panic.
+//
+//   - THE LOADER STRATEGY is fuzzer-controlled because Load, LoadAll,
+//     LibraryLoader and PackageLoader are four separate functions with four
+//     separate error-propagation paths, and LoadAll and LibraryLoader nest.
+//
+//   - src IS ORDINARY ELPS because installing a builtin proves nothing until
+//     something calls it. This is also the only way the fuzzer supplies the
+//     ARGUMENTS a builtin binds, which is where malformed formals bite.
+//
+// # The panic that is designed in, and why names are pre-checked
+//
+// lisp.AddBuiltins ends with
+//
+//	if exist.Type != LError { panic("symbol already defined: " + f.Name()) }
+//
+// and AddMacros/AddSpecialOps have the same shape. That is a deliberate
+// programming-error panic, it is not elpsutil's, and it is trivially reachable
+// -- `Function("true", ...)` alone does it, because Package.get answers the
+// reserved symbols `true` and `false` without consulting the symbol table, so
+// they are "already defined" in a package that is brand new and empty.
+//
+// A harness that let the fuzzer hit it would find that crasher in the first
+// second and then report the same one forever, covering nothing. So the
+// harness PRE-CHECKS every name against the package it is about to be bound in
+// and drops the ones that would collide (nameAvailable). The check is
+// conservative -- it demands an unbound symbol, which is the strictest of the
+// three Add* rules -- so it cannot leave a collision behind for any of them.
+//
+// The payoff is that the assertion needs no qualification: NO panic is
+// tolerated. There is no recover in elpsutil, none between it and the fuzz
+// function, so a panic during install is an ordinary crash that `go test
+// -fuzz` attributes to its input and writes to testdata. TestCollisionPanicIsReal
+// pins the panic the pre-check exists for, so the pre-check cannot quietly
+// become a filter that suppresses everything.
+//
+// # What is deliberately NOT fuzzer-controlled
+//
+// GO NIL IS NEVER GENERATED, in two places, and both are recorded findings
+// rather than oversights:
+//
+//   - `elpsutil.Function(name, nil, fn)` -- a plausible way to write "takes no
+//     arguments", the correct spelling being lisp.Formals() -- registers
+//     without complaint and then nil-derefs inside LEnv.bind at
+//     `fun.Cells[0].Cells` on the first call. The panic is laundered by
+//     lisp.eval's recover into an internal-panic LError, so the embedder sees
+//     a lisp error rather than a crash, at call time rather than at install
+//     time.
+//
+//   - A Loader (or PackageInit) that returns a Go nil *LVal nil-derefs
+//     immediately, in elpsutil itself, at `lerr.Type == lisp.LError` in Load /
+//     LoadAll and at `e.Type == lisp.LError` in PackageLoader. That one is an
+//     uncaught panic in the embedder's process.
+//
+// Both are contract violations by the embedder rather than defects in
+// elpsutil's own logic, and fixing either means adding validation to a public
+// constructor, which is an API decision and not this target's to take. They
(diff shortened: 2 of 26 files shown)


sam-at-luther:
## End-to-end production cost: the guard is free on a real transaction

The review measured the guard in isolation (allocations identical, +2.2ns on leaf rendering, +151%/+416% on deep acyclic `Equal` past depth 64). Nobody had measured what it costs a real transaction, so here is that number, plus the measurement that decides whether the deep-`Equal` regression is reachable at all.

**Setup.** Control `main` @ b1ab958, candidate `claude/fix-cyclic-value-walks` @ 57ecdbb. Both arms are the *same* substrate tree (`main` @ 8d10622) with byte-identical `go.sum`; the only difference between the two builds is a local `replace` pointing at a control or candidate elps worktree. Workload is a production-scale phylum driven through `router:cc-handle`, with the environment preheated outside the timed loop, so what is timed is exactly the per-request work a warm production env does.

### 1. Warm transaction, per endpoint

24 interleaved rounds per arm, `-count=1` per round, alternating which arm leads each round; benchstat pairs per-round results.

| endpoint | control | candidate | vs base | p |
|---|---|---|---|---|
| healthcheck | 633.0 µs ± 3% | 635.4 µs ± 3% | ~ | 0.943 |
| get_permission | 3.857 ms ± 2% | 3.839 ms ± 2% | ~ | 0.384 |
| get_client | 2.711 ms ± 3% | 2.707 ms ± 2% | ~ | 0.419 |
| update_client | 11.76 ms ± 1% | 11.84 ms ± 2% | ~ | 0.419 |
| **geomean** | **2.970 ms** | **2.973 ms** | **+0.11%** | |

No endpoint moves. Signs are mixed (two up, two down), which is the signature of noise and not of a tax. `allocs/op` is identical to the allocation on all four endpoints.

`B/op` rises by a deterministic but negligible amount — +32 B on healthcheck, +1071 B on update_client, i.e. 0.003–0.026% — with *zero* extra allocations. It is not the guard's map (see below, that map is never built); it is a size-class/code-layout artifact, and it is three orders of magnitude below anything a latency budget notices.

### 2. Does the real workload ever cross depth 64?

This is the question that decides whether the deep-`Equal` regression matters. I instrumented `cycleGuard.descend` and `pairGuard.descend` — the chokepoints every guarded walk (render, `Equal`, JSON encode, macro stamping, `GoValue`) passes through once per container node — to record the maximum depth reached, and ran the full corpus suite (484 tests, every endpoint) under it.

| walk | max depth reached | guard threshold | descends past threshold |
|---|---|---|---|
| value walks (render / JSON / macro / GoValue) | **24** | 64 | **0** |
| pair walks (`Equal`) | **3** | 64 | **0** |

Over ≥29.6M guarded descents, **nothing in the workload nests past depth 24** — 40 levels of headroom, and the deepest `Equal` in the entire corpus is 3 levels. The same probe over just the four benchmarked endpoints gives the same answer (24 / 2).

The consequence is categorical rather than statistical: stage 2 is never entered, the path map is never allocated, and every guarded descent in this workload is the stage-1 integer increment and compare. **The +151%/+416% deep-`Equal` figures are unreachable in practice** — the workload would have to nest 2.7× deeper before it paid a single map operation.

### 3. Full corpus suite wall time

11 paired runs. Both arms pass the same 484 tests.

| | control | candidate | paired delta |
|---|---|---|---|
| mean wall | 143.7 s | 144.4 s | +0.64 s (+0.44%), t = 0.42 |

Not significant. Worth flagging the methodology trap: the first 5 pairs ran control-first and showed a tidy-looking +1.87%, but the box drifts slower over a long session (135.5 s → 154.5 s, a 14% spread across the same binary), so a fixed order charges the second-run arm for the drift. Alternating the order flipped the sign to −0.72%. The suite is a much coarser instrument than the warm-transaction benchmark and can only bound the cost at roughly ±3%; the benchmark above bounds it at ~±1%.

### Verdict

**Free in production.** No endpoint regresses, allocations are unchanged, and the deep-`Equal` regression the review found is unreachable on this workload by a factor of 2.7 in nesting depth.

The one thing worth writing down is the condition under which that stops being true: the guard is free *because* real values are shallow, so the number that matters going forward is depth 64 versus a measured maximum of 24. If a future phylum ever nests past 64 — deeply recursive document schemas would be the way to get there — the cost lands on `Equal`, and the fix would be to raise `cycleGuardDepth` rather than to change the guard's shape, since the depth counter itself is what costs nothing. A regression test pinning the corpus maximum would catch that drift before latency does.

Measurement instrumentation is on `claude/cycle-guard-depth-probe` (probe only, not for merge — the atomics it adds make it useless for timing).


---

PR #400, opened 2026-08-13
Stop the cyclic-value fuzz target rewriting the standard library

### LINKED ISSUE #398: FuzzCyclicValueWalks silently corrupts the standard library on main, and nothing there can see it
> **Updated with stronger evidence, and one correction.** The original text said the fix could use `IsSealed()` because it "is already exported on `main`". **That is wrong — there is no seal bit on `main` at all** (`grep -rn IsSealed --include=*.go` returns nothing); the seal is chain-only. The fix is written in `main`'s own terms instead. The severity is also higher than first filed: this is not a probabilistic fuzzing hazard, it happens on **every plain `go test ./lisp/`**.

`FuzzCyclicValueWalks` (added by #391, `050ade5`) writes through cells of values it did not construct, and some of those cells are the **standard library's own formals lists**, shared process-wide.

## Why the sharing is total

`langBuiltins` / `userBuiltins` are package-level vars whose `Formals(...)` lists are built **once at package init**. `FunInPackage` (`lisp/lisp.go:517`) stores that list straight into the function value with no copy:

```go
Cells: []*LVal{formals, String("")},
```

So a single formals `*LVal` per builtin is shared by **every `LEnv` the process will ever create — including ones created after a write**.

`containers()` collects `v.Cells` for all types, including `LFun`. Confirmed directly on unfixed `main`:

```
car formals (shared process-wide): '(lis)
containers(car) returned 1 node(s)
containers() handed back the SHARED stdlib formals '(lis)  -- knot() writes through this
```

## It fires on every test run, not just under -fuzz

The triggering seed is `[]byte{kindFun, 2}`, and it is already committed in `fuzzval.Seeds()` (`internal/fuzzval/fuzzval.go:712`). So the seed-corpus run — plain `go test ./lisp/`, which is what PR CI runs — corrupts `lisp:car` today.

With a builtin-formals snapshot added to the existing `TestMain` (which already snapshots the three singletons for #274), unfixed `main` prints both of these in the same run:

```
PASS

FATAL: the process-wide builtin formals were mutated during the test run;
every LEnv built after the write saw the mutated list:
  - was: car '(lis)
    now: car '(lis #<cycle> #<cycle>)
```

`PASS` and `FATAL` together is the entire shape of this bug: the target's own assertions are satisfied, so nothing signals.

## The consequence, demonstrated behaviourally

Run the one seed, then build a **brand new** environment:

```
BEFORE (fresh env):  (car '(1 2 3))           => 1
AFTER  (fresh env, created after the run):
       (car '(1 2 3))                          => lisp:car: invalid number of arguments: 1
       (funcall car '(1 2 3))                  => lisp:car: invalid number of arguments: 1
       (map 'list car '((1 2)))                => lisp:car: invalid number of arguments: 1
```

The standard library is broken for the remainder of the process. Any later failure in an unrelated target is therefore unattributable, and a *passing* run may simply have corrupted something no assertion looks at.

Note for the nightly sweep: under `-fuzz` the corruption happens in worker subprocesses, so a coordinator-side snapshot stays clean. The detection point is the seed-corpus run.

## Fix (in `main`'s terms)

`containers()` returns immediately on `lisp.LFun`, collecting and descending nothing below a function. On `main` the shared subtrees a generated value can reach are exactly those under an `LFun` — `fuzzval.fun()` returns `env.GetGlobal` results verbatim — and `fuzzval` itself already refuses to write to them. `containers()` had not got the memo.

## Coverage survives — measured three ways

- **Per-seed container delta, node by node**, pre- and post-fix over all 142 seeds: seeds yielding ≥1 container 32 → 23. All nine differences are `rootType=function` seeds, and every dropped node is a formals list (`'(x)`, `'(lis)`, `'(bindings &rest expr)`, `'(name formals &rest expr)`). **Not one node an input actually constructed was dropped.**
- **`-fuzz` for 60s each, `go clean -fuzzcache` before both**: 511 → 513 new interesting inputs. Statistically identical.
- **A committed regression guard** so this cannot rot: asserts ≥20 seeds still yield a writable node, ≥20 still render `#<cycle>` after knotting, and that all three `knot()` branches (list, array, sorted-map) are still reached — measured 24/9/6. Verified to bite by sabotaging `containers()` to `return nil`: all five assertions fail.

Fixed on `claude/fix-398-fuzz-stdlib`.

## Related

The inverse of elps#393, from the same fork lag: #391 was written against a tree with no sealed-AST verifier and merged to `main`, while `libelpspath` lives only on the chain and so was never in the tree #391 swept. Each change is invisible to the other's checks.

sam-at-luther:
Fixes #398.

`FuzzCyclicValueWalks` (added by #391) writes through cells of values it did not construct, and some of those cells are the standard library's own formals lists. **This happens on every plain `go test ./lisp/`, not just under `-fuzz`** — the triggering seed is already committed in `fuzzval.Seeds()`.

## Why the sharing is total

The builtin tables are package-level vars whose `Formals(...)` lists are built once at package init, and `FunInPackage` stores that list into the function value with no copy:

```go
Cells: []*LVal{formals, String("")},
```

One formals `*LVal` per builtin is therefore shared by **every `LEnv` the process will ever create — including ones created after a write**. `containers()` collected `v.Cells` for all types including `LFun`, so it handed `knot()` the stdlib's own definitions to tie cycles through.

## The failure mode this PR is really about

With a builtin-formals snapshot added to the existing `TestMain`, unfixed `main` prints both of these **in the same run**:

```
PASS

FATAL: the process-wide builtin formals were mutated during the test run;
every LEnv built after the write saw the mutated list:
  - was: car '(lis)
    now: car '(lis #<cycle> #<cycle>)
```

The target's own assertions are satisfied, so nothing signals. Behaviourally, after the run a **brand new** environment gives:

```
(car '(1 2 3))  =>  lisp:car: invalid number of arguments: 1
```

The standard library is broken for the rest of the process, so any later failure in an unrelated target is unattributable — and a *passing* run may have corrupted something no assertion looks at.

## The fix

`containers()` returns immediately on `lisp.LFun`, collecting and descending nothing below a function. The shared subtrees a generated value can reach are exactly those under an `LFun` — `fuzzval.fun()` returns `env.GetGlobal` results verbatim — and `fuzzval` itself already refuses to write to them. `containers()` had not got the memo.

## Coverage survives — measured three ways

A fix that quietly turns the target into a no-op would be the same "green by not looking" failure in a new place, so this is checked explicitly:

- **Per-seed container delta, node by node**, over all 142 seeds: seeds yielding ≥1 container 32 → 23. All nine differences are `rootType=function` seeds, and every dropped node is a formals list. **Not one node an input actually constructed was dropped.**
- **`-fuzz` for 60s each**, `go clean -fuzzcache` before both: 511 → 513 new interesting inputs. Statistically identical.
- **A committed regression guard**, `TestCycleSeedsStillTieCycles`: asserts ≥20 seeds still yield a writable node, ≥20 still render `#<cycle>` after knotting, and that all three `knot()` branches (list, array, sorted-map) are still reached — measured 24/9/6. Verified to bite by sabotaging `containers()` to `return nil`: all five assertions fail.

## Note for the nightly sweep

Under `-fuzz` the corruption happens in worker subprocesses, so a coordinator-side snapshot stays clean. The detection point is the seed-corpus run — which is what PR CI runs.

## Gates

`go test ./...`, `go test -race ./lisp/...`, `golangci-lint run ./...` (0 issues), `elps fmt -l ./...`, `scripts/confidentiality-guard.sh`, `scripts/ci-gates-test.sh` (126 passed), `scripts/fuzz-budget-check.sh` — all clean.

---

### DIFF
--- a/lisp/builtin_formals_test.go
+++ b/lisp/builtin_formals_test.go
@@ -1,0 +1,71 @@
+// Copyright © 2026 The ELPS authors
+
+package lisp
+
+import (
+	"fmt"
+	"strings"
+)
+
+// builtinFormalsSnapshot is a rendering of every builtin's formal argument
+// list, and the third process-wide guard TestMain installs.
+//
+// WHY IT IS NEEDED.  The builtin tables (langBuiltins, userBuiltins) are
+// package-level vars whose Formals() lists are constructed exactly once, when
+// this package is initialized.  AddBuiltins stores f.Formals() straight into
+// LFun.Cells[0] with no copy, so ONE formals list is shared by every LEnv the
+// process will ever create -- including every environment created after a
+// write to it.  A test that walks a value, reaches a function through it and
+// appends a cell to the list it finds in Cells[0] has therefore edited the
+// standard library for the remainder of the process, and nothing in that test
+// is looking.
+//
+// That is not hypothetical: FuzzCyclicValueWalks did exactly this from its
+// own seed corpus, and every test in the package passed while it happened
+// (issue #398).  The singleton snapshot above cannot see it -- the singletons
+// are Nil and the two Bools, not the builtin table -- and main has no
+// sealed-AST verifier, so before this snapshot there was no second observer
+// at all.
+//
+// The rendering, rather than the pointers, is what is compared: an append
+// that fits in the existing slice capacity leaves the *LVal identical and the
+// backing array edited in place, which a pointer comparison would miss.
+type builtinFormalsSnapshot struct {
+	rendered string
+}
+
+func takeBuiltinFormalsSnapshot() builtinFormalsSnapshot {
+	return builtinFormalsSnapshot{rendered: renderBuiltinFormals()}
+}
+
+func renderBuiltinFormals() string {
+	var b strings.Builder
+	for _, def := range DefaultBuiltins() {
+		fmt.Fprintf(&b, "%s %s\n", def.Name(), def.Formals().String())
+	}
+	return b.String()
+}
+
+// verify reports the builtins whose formals differ from the snapshot, or ""
+// when the table is intact.
+func (s builtinFormalsSnapshot) verify() string {
+	now := renderBuiltinFormals()
+	if now == s.rendered {
+		return ""
+	}
+	was, is := strings.Split(s.rendered, "\n"), strings.Split(now, "\n")
+	var out strings.Builder
+	for i := range max(len(was), len(is)) {
+		var wl, il string
+		if i < len(was) {
+			wl = was[i]
+		}
+		if i < len(is) {
+			il = is[i]
+		}
+		if wl != il {
+			fmt.Fprintf(&out, "  - was: %s\n    now: %s\n", wl, il)
+		}
+	}
+	return out.String()
+}
--- a/lisp/cycle_fuzz_test.go
+++ b/lisp/cycle_fuzz_test.go
@@ -139,12 +139,12 @@ func walkAll(v, other *lisp.LVal) string {
 // knot ties up to n self-references into v, turning the tree fuzzval built
 // into a graph with cycles in it.
 //
-// Only sorted-maps, arrays and non-empty lists are written to: the shared Nil
-// and Bool singletons are an empty list and two symbols, and writing through
-// one of those corrupts every other holder of it (issue #274).  Every target
-// value is a node from v itself, so a knot that lands on an ancestor is a real
-// cycle and one that lands elsewhere is shared structure -- both are shapes
-// the walks have to survive.
+// Only sorted-maps, arrays and non-empty lists are written to, and only ones
+// containers() collected: the shared Nil and Bool singletons are an empty
+// list and two symbols, and writing through one of those corrupts every other
+// holder of it (issue #274).  Every target value is a node from v itself, so
+// a knot that lands on an ancestor is a real cycle and one that lands
+// elsewhere is shared structure -- both are shapes the walks have to survive.
 func knot(v *lisp.LVal, gen *fuzzval.Gen, n int) {
 	nodes := containers(v)
 	if len(nodes) == 0 {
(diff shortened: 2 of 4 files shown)


iamsamwood (commented):
Independent verification review, run in an isolated worktree at head `bbaee2a`, trial-merged against main `663c0f3`. Verdict from this pass: **the fix is correct and root-cause — this would be an approve.** Posting as a comment so the merge decision stays with the PR owner; findings below are all non-blocking.

**Fix verification.** The issue's central claim was reproduced live on unfixed main: replaying the committed seed corpus through the harness's own `knot()`/`containers()` corrupts `lisp:car`'s process-wide formals, and a brand-new env created afterwards fails `(car '(1 2 3))` with `invalid number of arguments: 1`. The fix (`containers()` returning immediately on `lisp.LFun`) removes the harness's ability to write through values it didn't construct — the correct ownership rule, not an input-skip or a defensive copy. Verified the PR's claim that the only shared subtrees reachable from `gen.Value()` are under an `LFun`.

**Red-test.** With the fix reverted (guards kept): `TestCycleContainersSkipFunctionSubtrees` fails deterministically and the new TestMain oracle FATALs showing `car '(lis)` → `'(lis #<cycle> #<cycle> #<cycle> #<cycle>)`. With the fix restored: all green, full `go test ./lisp/...` passes. The target is not neutered: `TestCycleSeedsStillTieCycles` floors hold (20/20, all three `knot()` branches reached), and a fresh 10s `-fuzz` run did 37k execs finding 228 new interesting inputs. The anti-neutering guard bites when `containers()` is sabotaged to return nil — all five assertions fail as the PR describes.

**Findings (non-blocking):**

1. `lisp/builtin_formals_test.go:41-46` — the snapshot renders `DefaultBuiltins()` only, but `Macro()` and `SpecialOp()` share formals by pointer exactly the same way (`lisp/lisp.go:554`, `:582`), and fuzzval reaches `lisp:let`, `lisp:quote`, `lisp:defmacro`, `lisp:cond` through `fun()` case 3 from committed seed `[]byte{kindFun, 3}`. Confirmed on the unfixed harness that a seed replay mutates `let '(bindings &rest expr)` and `defmacro`'s formals too, while the oracle's FATAL reports only `car`. The `containers()` fix protects them, but the process-wide observer this PR adds — the thing that makes future corruption visible — is blind to two of the three shared tables. Extending `renderBuiltinFormals()` over `DefaultSpecialOps()` and `DefaultMacros()` is ~three lines and closes the blind spot for any future writer, not just this target.

2. `lisp/cycle_fuzz_test.go:119-121` — under `-fuzz`, a formals write in a worker is detected at best by the worker-process TestMain after the fact, with no input attribution or minimization (the PR discloses this). Since the only globals reachable from a generated value are the ~16 `fuzzval.funNames` entries plus four special forms, rendering just those formals before/after each input would give per-input detection at negligible cost. Fine as a follow-up; the seed-corpus detection this PR establishes is the load-bearing part for CI.

3. Commit `03ddfb3` (observer without fix) fails `go test ./lisp/` on its own — verified at that commit. Nice demonstration that the oracle works, but it breaks any future `git bisect` predicate over the lisp package. Suggest squash-merging so no red commit enters main's history.

4. `lisp/cycle_fuzz_test.go:190-192` — "for a lambda it is a body the interpreter may hold references into" overstates: for a fuzzval-built lambda (`fun()` case 1) the body cells are input-constructed and nothing else references them. The blanket skip is still right (uniformity + demonstrably ~nil coverage cost), but one clause of honesty would stop a future reader from "fixing" the skip on the strength of a wrong premise.

Scope: clean — the entire change is test-side, which is the correct scope for a harness bug. CI 19/19 green at head; local trial merge "already up to date" against `663c0f3`.

---

Issue #401, opened 2026-08-13
libelpspath's copying ?set/?del/?nil write the ARRAY layout onto a LIST copy, corrupting a leaf the source still holds


sam-at-luther:
The non-mutating `?set` / `?del` / `?nil` accept a list, and they are supposed to: they copy first and rework the copy. The rework then wrote the **array** layout — `in.Cells[1].Cells = vals`, followed by `in.Cells[0].Cells[0].Int = len(vals)` for the dims bookkeeping — into whatever it was handed. On a list copy those two stores index the list's own **elements**, not an array's `[dims, data]` wrapper.

`errMutateList` does **not** close this. It guards the exported `SetMutate` / `DeleteMutate` / `NilMutate` entry points only. The non-mutating `Set` / `Delete` / `Nil` deliberately bypass it — they call the unexported `setMutate` / `deleteMutate` / `nilMutate` on a private copy — and that private copy is a **list** whenever the input was one. Every `?set`/`?del`/`?nil` with an index or range step into a list reaches the write-back with list layout.

## Two failure modes

**1. A wrong answer with no error, reaching a document the operation was never given.**

A correct deep copy shares its *leaves* with the source — an int or a string is an immutable `LVal`, so rebuilding one buys nothing and costs an allocation per scalar. The write-back then lands on a leaf the source is still holding:

```lisp
(use-package 'elpspath)
(set 'doc (sorted-map "l" '(1 2)))          ; "l" is a quoted list literal from the parse
(set 'cp  (elpspath:?set doc "tag" "x"))    ; an ordinary copying op that never touches "l"
(set 'res (elpspath:?del cp "l" 0))         ; a copying delete on the COPY
(debug-print (list 'source-list (elpspath:? doc "l")
                   'copy-list   (elpspath:? cp "l")
                   'result-list (elpspath:? res "l")))
```

```
'('source-list '(1 1) 'copy-list '(1 1) 'result-list (1 1))     ; before the fix
'('source-list '(1 2) 'copy-list '(1 2) 'result-list (2))       ; after
```

`'(1 2)` became `'(1 1)` in the source document — and, since it is a program literal, in the AST for every environment sharing that parse. Nothing raised.

**2. An uncatchable panic where lisp code expected a value.**

```lisp
(handler-bind ((condition (lambda (&rest c) 'caught))) (elpspath:?del '(1) 0))
```

```
error: internal-panic: ... runtime error: index out of range [1] with length 1     ; before
()                                                                                 ; after
```

[...]

Issue #395, opened 2026-08-13
libelpspath copyMap is shallow where copyList/copyVector are deep, so ?set/?del/?nil on a sorted-map return a sharing "copy"


sam-at-luther:
`copyList` and `copyVector` recurse (`path.go`), carrying the comment:

> **IMPORTANT: lists may contain containers, in which case we need to copy those containers.**

`copyMap` (`path.go:61–75`) does not. It does `m.Set(pair.Cells[0], pair.Cells[1])` — so the "copy" shares every nested container with the source. Sorted-maps are the commonest shape these functions serve, and they are the one shape where the copy contract does not hold.

## Reproduction — the package's own documentation example

```lisp
(set 'patient (sorted-map "ssn" "123" "address" (sorted-map "city" "London")))
(set 'redacted (elpspath:?nil patient "ssn"))
(elpspath:?set! redacted "address" "city" "REDACTED")
```

```
'("REDACTED" "REDACTED")
```

The write through the redacted copy reaches the original patient record. That the doc example is a redaction workflow makes this worth fixing on its own.

Structural variant — `?del!` through a `?set` copy restructures the source array in place and rewrites its dims:

```
'("(vector 2 3)" "(vector 2 3)")
```

Controls, all independent as expected: list source `'("(vector 1 2 3)" "(vector 1 2 3 777)")`, vector source the same, kernel `copy` the same. **Only maps leak.**

## Scope

**Not literal-reachable** — verified rather than assumed: `append 'vector`, `stable-sort` and `?set!` on a literal list carried through the map copy all leave `'(1 2 3)` intact, because the kernel's copy-on-write and `errMutateList` hold. So this is a local surprise and a broken contract, not cross-environment corruption.

Worth stating plainly though: that negative is **guard-dependent, not structural**. It holds *given* `builtinAppend`'s and `builtinSlice`'s `sealed` copy-on-write. Remove either and this becomes literal-reachable.

Silent in all five layers (`go test`, `-tags elpscheck`, `-race`, both elpsvet passes).

## Fix

Make `copyMap` deep, matching `copyList`/`copyVector` and the comment that already explains why. Worth adding a test that asserts the three copy helpers agree on nesting depth, so they cannot drift apart again.

Found during a corruption hunt. Same package: elps#392 (seal laundering), elps#393 (unguarded walkers → process kill).

sam-at-luther:
[...]

Issue #392, opened 2026-08-13
elpspath's read-only ? getter launders the seal off a program literal, reopening the substrate#378 corruption class


sam-at-luther:
`rangePath.Get` — the **non-mutating** query path — hands back a fresh **unsealed** `LVal` header over a **sealed** program literal's live backing array. Every kernel copy-on-write site keys off `list.sealed`, so once the flag is laundered off, none of them fire and the literal is mutated in place, permanently, for every environment sharing that parse.

This is the substrate#378 class, reached through the read path. #378 closed the mutating half (`errMutateList` on `*Mutate` entry points); nothing guards the getter, because a getter was not expected to need guarding.

## Reproduction — pure lisp, no Go, no embedder

```lisp
(defun limits () '(10 20 30 40))
(debug-print (limits))
(set 'win (elpspath:? (limits) '(range 0 2)))
(append 'vector win 999)
(debug-print (limits))
```

```
'(10 20 30 40)
'(10 20 999 40)
```

The literal in the function body is changed for the life of the process. `?` is documented and named as a query.

Variants, all confirmed:

| variant | literal after |
|---|---|
| `(stable-sort '> (elpspath:? (cfg) '(range 0 3)))` | `'(30 20 10 40)` |
| `(append! (slice 'vector (elpspath:? (cfg) '(range 0 2)) 0 2) 777)` | `'(1 2 777 4)` |
| nested under a runtime array: `(elpspath:? (vector (cfg)) 0 '(range 0 3))` | `'(30 20 10 40)` |

## Mechanism

`lisp/lisplib/libelpspath/path.go`, `rangePath.Get`:

```go
cells = cells[from:to]                 // two-index slice of LIVE backing, capacity retained
if in.Type == lisp.LArray {
    newVal = toVector(cells)           // lisp.Array(nil, cells)
} else {
    newVal = toList(cells)             // lisp.SExpr(cells)
}
```

`lisp.SExpr` / `lisp.Array` mint a fresh header with `sealed == false`. The kernel does the same aliasing correctly two files away, in `builtinSlice`:

```go
sealed := list.sealed
list = QExpr(seqCells(list)[i:j])
// "A two-index slice keeps the original backing array (and its spare
//  capacity), so a sealed input's constraint travels with the
//  intermediate value"
list.sealed = sealed
```

`builtinCdr` and `builtinRest` propagate it too. `libelpspath` is the only producer in the tree that does not.

[...]

PR #402, opened 2026-08-13
Add libelpspath, with the copy and cyclic-value fixes (elps#393, elps#395, elps#401)

### LINKED ISSUE #393: libelpspath's walkers are unguarded against cycles — a process kill #391 never had the chance to cover
> **Corrected.** The original title and framing of this issue asserted that #391's sweep *missed* `lisp/lisplib/**`. **That was wrong**, and so was the "the kernel survives and the path engine dies in the same program" contrast in the first version — the contrast only holds on a tree that has #391, and the chain does not. Both corrections are below, with the evidence. The underlying bug is real and unchanged.

`lisp/lisplib/libelpspath` walks values recursively with no cycle guard and no depth bound, so a cyclic value takes down the host through the path engine:

- `okSimpleType` ↔ `okSimpleContainerType` (`libelpspath.go:139/144/186`), mutually recursive — the type check every `?`-family builtin runs.
- `copyLVal` / `copyList` / `copyVector` (`path.go:40–110`).

```lisp
(set 'v (vector 1 2))
(append! v v)
(elpspath:? v 0)
```
```
fatal error: stack overflow
      frames: libelpspath.okSimpleType / okSimpleContainerType / BuiltinQueryGet
```

A stack overflow is **not recoverable**: `recover()` does not catch it, `handler-bind` never sees it. In an embedded chaincode interpreter this is a phylum-author-triggerable availability kill.

Measured across the whole surface, each in its own process: **all 7 elpspath builtins × 2 cyclic shapes = 14/14 fatal.** `json:dump-*`, `json:message-bytes` and the kernel walkers survive.

## Correction 1 — #391 did not skip lisplib

`050ade5` guards `lisp/lisplib/libjson/encode.go` and `json.go` and adds `libjson_cycle_test.lisp`. lisplib was in scope, and #391 made per-function judgements there — it deliberately left `libjson.(*Serializer).GoValue/GoSlice/GoMap` unguarded, documenting them as deprecated rather than fixing them.

## Correction 2 — the real cause is branch topology, not sweep scope

`libelpspath` **did not exist in the tree #391 swept**:

```
git merge-base origin/main <chain tip>          → b1ab958   (chain forked here)
git merge-base --is-ancestor 050ade5 <chain tip> → NO        (#391 landed on main after the fork)
git log --diff-filter=A -- .../libelpspath/     → b7ad5ca "lisplib: adopt elpspath v2 from substrate"  (chain-only)
```

`libelpspath` is the **only** non-test lisplib package present on the chain and absent from `main`. So `git log 050ade5 -- lisp/lisplib/libelpspath` being empty — which the first version of this issue cited as evidence of an incomplete sweep — simply reflects a path that did not exist at that commit. 136 commits on the chain, 3 on `main` since the fork, and the two changes have never met.

**Consequence the original issue got backwards:** on the chain tip the *kernel* is unguarded too, because #391 is absent entirely. `(debug-print v)` on a cyclic vector dies there by itself. Guarding `libelpspath` alone would have left the repro above fatal. The fix therefore ports #391 onto the chain first.

## The lag cuts both ways

Landing #391 on the chain **failed the chain's own `elpscheck` sealed-AST verifier** — 3 of 37,831 sealed roots mutated, bisected to #391's own `FuzzCyclicValueWalks`, whose `knot` helper writes directly into `Cells` and can land on a parsed program node. #391 was written against a tree with no such verifier, so nothing on `main` could have caught it.

## Two further unguarded walkers, found while checking the answer

1. **`libjson.(*Serializer).GoValue/GoSlice/GoMap`** — on `main` today. Not reachable from lisp (only these three call each other), deprecated, and #391 documented the hazard rather than guarding it. An embedder calling it on a tenant value still gets an uncatchable kill.
2. **`expansionFullySealed`** (`lisp/macrocache.go:703`) — chain-only, so #391 never saw it; no memo, no bound, fatal on a self-referential expansion in a Go-level repro. No lisp-level path found: the macro cache is off by default and the purity prover rejects any macro whose body mutates. Latent — and exactly the shape the merge-lag answer predicts.

Verified cycle-safe by construction: `forker.val`, `detacher.detach`, `markTree`/`treeBytes` (all memoise before descending) and `sealFP.walk` (explicit budget and depth bound). `sealAST` is safe because it sets `sealed` before recursing.

## Fix

`cycleGuard` is unexported and unreachable from `libelpspath`, and a local copy would have been the **third** implementation of the protocol (libjson's `encodeGuard` is the second). Smallest sound export: one wrapper type with `Descend`/`Ascend`/`Tracking`, keeping `cycleState`, the depth constant and strict mode private. No rerun stage is needed, because a walk that answers a cycle with an *error* stops at the first frame that finds one — only marker-producing walks (rendering) need escalation, and those stay in package `lisp`.

Guarded in both families. The copy is guarded separately from the type check, because the exported `Path` interface lets a Go embedder reach a copy without passing through the builtin gate.

Cost: one 16-byte state allocation per walk, no measurable time (`TypeCheck` and `BuiltinGet` both `~`, +1 alloc).

Fixed on `claude/fix-393-395-libelpspath`, with a subprocess-based regression test — a stack overflow cannot be asserted on in-process — red-proofed by stubbing the `Descend` calls.

Related: elps#392, elps#395 (same package), elps#397 (a second unrecoverable fatal, concurrent map write).

sam-at-luther:
Lands `libelpspath` on `main` as a standalone package, carrying three bug fixes, **without** the sealed-AST feature. This unblocks fixes that were otherwise gated behind unreleased feature work.

Fixes #393, #395. Records #401.

## The three bugs

**#395 — the copy shared nested containers.** `?set`/`?del`/`?nil` returned a "copy" that still shared its nested containers, so writing through the copy reached the source. The package's own doc example leaked:

| | `(list (? redacted "address" "city") (? patient "address" "city"))` |
|---|---|
| red | `'("REDACTED" "REDACTED")` |
| green | `'("REDACTED" "London")` |

Fixed with a **path-directed** copy: the same contract as a deep copy — the result shares no container with its source — skipping only the subtree the operation is about to overwrite. A *fully* path-aware copy (share off-path subtrees) was evaluated and rejected as unsound: for a one-step path the spine is the root alone, which is bit-for-bit the bug being fixed.

**#393 — a cyclic value killed the process.** Unrecoverable stack overflow, not a catchable error. Asserted out of process, since a stack overflow is a fatal runtime error rather than a panic:

| | exit | output |
|---|---|---|
| red | **2** | `fatal error: stack overflow`, 292 lines |
| green | **0** | `'caught-a-condition` then `'survived` |

**#401 — a copy wrote through a shared leaf.** The copying paths wrote back with *array* layout onto a private copy that is a **list**. Every correct deep copy shares immutable leaves, so the write landed on the **source's own** leaf:

```
container: src 0xc000180850, copy 0xc000180c40, shared=false   (the deep copy is correct)
leaf:      src[1] 0xc0001807e0, copy[1] 0xc0001807e0, shared=true
red   after: src='(1 1)   ← the write landed on the SOURCE's leaf
green after: src='(1 2)
```

`errMutateList` does **not** close this: it guards only the exported `SetMutate`/`DeleteMutate`/`NilMutate`. The non-mutating `Set`/`Delete`/`Nil` deliberately bypass it, calling the unexported mutators on a private copy — and that copy is a list whenever the input was one. Fixed with a layout-dispatching `storeCells`.

## A verification gap this exposed

`fpAST` walked only `Cells`, so every fingerprint assertion **went blind at the first sorted-map**. That is why the corruption survived the existing suite. Fixed to walk map entries; the fuzz target then fails on the checked-in corpus entry `025497832319941d` in the red tree.

## Battery

3 source shapes × every copy op at every path ≤3 steps × every in-place op at every path ≤3 steps × both re-query splits: **8007 combinations written through (`map:5301 list:1353 vector:1353`), none reached the source.** Per-source count floor (`requireBatteryWrote`, >500 each) so an empty sweep cannot masquerade as a pass. Red-proofed against a spine-only copy → **1044 failures**.

## Performance — a correct copy is slower than a sharing one

12 rounds, arms interleaved round-robin, one binary per arm, `-cpu=1`. Versus the pre-fix sharing copy:

| | sec/op | B/op | allocs/op |
|---|---|---|---|
| `Copy` | **+434%** | +441% | +426% |
| `BuiltinSetCopy` | **+67%** | +69% | +70% |
| `BuiltinNilCopy` | **+95%** | +98% | +104% |
| `BuiltinGet` | ~ (p=0.198) | 0% | 0% |
| `TypeCheck` | ~ (p=0.551) | 0% | 0% |

Reads pay nothing. What path-direction buys back (undirected → directed, all p=0.000):

| | undirected | directed | Δ |
|---|---|---|---|
| depth2 | 19.33µ | 14.78µ | −23.6% |
| depth4 | 56.93µ | 29.30µ | −48.5% |
| depth8 | 191.94µ | 58.98µ | **−69.3%** |
| geomean | 23.97µ | 16.45µ | **−31.4%** |

**Cycle guard cost, honestly:** exactly **+1 allocation per walk** (asserted, and asserted not to scale with width), +16 B/op, and on time it is at noise on 2 of 4 rows but resolves at **+2.9%** and **+5.7%** on the other two. Not literally free.

## What was left behind with the seal

- **`path_fuzz_test.go` invariant 4 adapted, not dropped.** Upstream compared `lisp.SealedASTFingerprint`, which does not exist on `main`. Replaced with a list-spine fingerprint. **Coverage cost:** the seal drew the mutable/immutable line for the whole document at once; the narrowed version covers only lists. The shared-leaf case that falls in the gap is picked up by invariant 5 and by a new targeted test. Documented in-file.
- **#392 deliberately not ported — and it is not a seal bug.** `LVal` has no `sealed` field on `main`, so there is nothing to launder. But its reproduction still corrupts here, so the control was run: **the identical corruption reproduces on pristine `main` with no libelpspath in the tree at all**, using only kernel `(slice 'vector …)` + `append!`. That is the pre-existing #369/#373 capacity-retention class — `rangePath.Get` is one more producer behaving exactly like the kernel's own `slice`, not a new hole. Recorded in a comment rather than silently inherited.

## Also included

`main`'s own `fuzz.yml` comment predicted that a 21st fuzz target would put `required` at exactly the 60-minute timeout — the gate would pass while describing a truncated sweep. `FuzzPathEngine` is that target, so the sixth shard it asked for is added here: `ceil(21/6)=4`, 10 minutes of headroom restored.

## Gates

`go test ./...`, `go test -race ./lisp/...`, `golangci-lint run ./...` (0 issues), `elps fmt -l`, `gofmt -l`, `confidentiality-guard.sh`, `ci-gates-test.sh` (**126 passed, 0 failed**), `fuzz-budget-check.sh` (fits, 10 min headroom), `FuzzPathEngine` 120s/838k execs clean. No `cmd/elpsvet` or elpscheck seal tag exists on `main`, so neither gate was run.

**22 files, +5708/−12.**


sam-at-luther:
**Correcting the cycle-guard cost upward, using this run's own numbers.**

The PR body reports the guard as "at noise on 2 of 4 rows, +2.9% and +5.7% on the other two." That was measured on the authoring machine. **This CI runner (arm64, 2-core, n=10) shows a cost on all four rows**, and a larger one:

| row | unguarded | guarded | Δ |
|---|---|---|---|
| `record/typecheck` | 3.724µ | 4.002µ | **+7.5%** |
| `record/copy` | 6.526µ | 7.030µ | **+7.7%** |
| `spine8/typecheck` | 14.80µ | 15.43µ | **+4.3%** |
| `spine8/copy` | 27.30µ | 29.33µ | **+7.4%** |

Run-to-run variance on these rows is ±3–9%, so any single row sits near the noise floor — but the direction is consistent across all four, which the local run's two `~` rows did not show. **The fair statement is ~3–8% on a guarded walk, runner-dependent, rather than the lower figure in the body.**

The allocation claim holds exactly, on this runner too — **+1 allocation per walk**, no more:

| row | unguarded | guarded |
|---|---|---|
| `record/typecheck` | 78 | 79 |
| `record/copy` | 141 | 142 |
| `spine8/typecheck` | 288 | 289 |
| `spine8/copy` | 568 | 569 |

+16 B/op each, the shared `cycleState`.

**Scope check — this does not reach the wider tree.** Existing benchmarks with a baseline are flat: `lisp` geomean **−0.46%** sec/op and **+0.00%** allocs across 40 benchmarks; `parser/rdparser` +0.74% sec/op, +0.00% allocs; `libjson` +0.09%. The one real movement is `libjson Package/$load` at +0.70% B/op / +0.44% allocs (p=0.000), which is `libelpspath` being registered at load — expected for a new package.

So: the guard costs a few percent on the two walks that use it, an allocation per walk, and nothing measurable anywhere else. The alternative is an unrecoverable process kill on a cyclic value (#393), which is not a trade worth optimising away — but the number should be the honest one.

---

iamsamwood (commented):
Independent verification review at head `d27b466` (base == current `origin/main` `663c0f3`; all 22 changed files read; every claimed fix red/green-verified by reverting its hunk). Posting as a comment so the branch owner drives the corrections. Verdict from this pass: **would be REQUEST_CHANGES — one new blocking defect found by the adversarial probes; everything else checks out.**

## The good news first: all three claimed fixes are real

- **#393** (cyclic value kills the process): disabling both guards' `if cyclic` rejections makes all four subtests die with `fatal error: stack overflow`; restored, green. The subprocess-based kill test with a 16MB stack cap is the right shape, and `lisp.CycleGuard` threads through both walker families and all seven builtin entry points.
- **#395** (shallow map copy): reverting the deep copy to `val := pair.Cells[1]` → `TestAliasBattery` fails with **2826 combinations reaching the source** and the redaction example leaks `London → REDACTED` into the source record; restored, green. The battery's `requireBatteryWrote` floor makes a vacuous pass impossible.
- **#401** (array layout onto list copy): forcing the array branch reproduces both documented failure modes (index-out-of-range panic and source corruption); restored, green.

## BLOCKING — new defect: `rangePath.setMutate` corrupts the result when the replacement is longer than the range

`lisp/lisplib/libelpspath/path.go:979-982`: `vals := append(cells[:from], setCells...)` writes through the same backing array that `cells[to:]` is read from two lines later. When `len(setCells) > to-from` and `to < n`, the first append overwrites cells in `[to, from+len(setCells))` before the tail append reads them — a silent wrong answer, no error. The private copy handed in by the non-mutating `Set` always has `cap == len` (`copySeqOffPath`), so the copying `?set` is **deterministically** corrupt for every such range/replacement.

Reproduced independently twice (review worktree and a second fresh build of `d27b466`):

```lisp
(use-package 'elpspath)
(?set (vector 1 2 3 4 5) '(range 0 1) (vector 97 98 99))
; got:  (vector 97 98 99 98 99 4 5)
; want: (vector 97 98 99 2 3 4 5)
(?set '(1 2 3 4 5) '(range 0 1) '(97 98 99))     ; list shape: same corruption
(?set! v '(range 0 1) (vector 97 98 99))          ; mutating variant: same corruption
(?set (vector 1 2 3 4 5) '(range 0 2) (vector 97 98)) ; equal-length control: correct
```

Equal-length and shorter replacements are unaffected, and `to == n` escapes because the tail append is skipped — exactly the corners the existing tests live in: `TestMutateArrayMapInPlace`'s only range-set case uses a shorter replacement, `query_test.go` has no range-set case, and `FuzzPathEngine` can never reach this because its value argument is always the string `"REPLACEMENT"`, which `toCells` rejects before the splice. `nilMutate` and the delete paths are safe (verified).

Suggested fix — build the result in a fresh slice instead of appending onto the input's prefix:

```go
vals := make([]*lisp.LVal, 0, from+len(setCells)+(n-to))
vals = append(vals, cells[:from]...)
vals = append(vals, setCells...)
vals = append(vals, cells[to:]...)
storeCells(in, vals)
```

Please also add a regression test for range-set with a longer-than-range replacement where `to < n` (both `?set` and `?set!`, vector and list), and consider teaching `FuzzPathEngine` to sometimes supply a small vector/list value so the splice path is fuzzed at all — invariants 4/5 can't see this because the corruption lands in the *result*, which no invariant checks.

## Non-blocking findings

1. **Out-of-range index: copying `?set` silently nils the subtree while `?set!` no-ops.** `(?set (sorted-map "k" (vector 1 2 3)) "k" 10 "x")` returns `(sorted-map "k" ())` — the vector silently replaced by nil — while `(?set! doc "k" 10 "x")` is a no-op. Cause: `indexPath.setMutate` (path.go:830-832) answers an unresolvable index with `(lisp.Nil(), nil)`, which `setChain` installs at the parent key. A silently data-destroying divergence between the `!` and non-`!` variants; worth unifying (return the untouched copy) or pinning deliberately in a test + issue.

2. **#399 is confirmed unfixed on this branch, on every route** — dot (`mmap.Set(key, newIn)`), index (`cells[index] = newIn`), range splice (shares the replacement's element LVals), and root (empty chain returns `newIn` itself); probes observed the caller's value argument mutated in all four cases. Consistent with the PR not claiming it — but the issue's minimum ask (document the by-reference contract, pin it with a characterization test) is also absent, so the contract ships silently. Suggest at least a docstring note + characterization test in this PR, and an explicit deferral note in the PR body.

3. **#392 deferral: agreed.** The range getter's capacity retention (probe: `(append! s 99)` writes into the source) is the pre-existing #369/#373 kernel class — kernel `slice` behaves identically, verified — and the `path.go:922-934` comment records it correctly. When the kernel settlement lands, `rangePath.Get` must be included in it; that follow-through currently has no owner.

4. **Nit:** `copyContainer` (path.go:96-104) silently returns nil for a multi-dimensional array — unreachable via builtins, but a Go embedder using the exported `Path` interface gets silent data loss where an error would cost nothing.

5. **Nit:** PR body says "Records #401" — the fix and full regression coverage for #401 land here, so "Fixes #401" would close it on merge (its own body says it's "not outstanding work" once this lands).

## Package bar & coordination

Docstrings/`elps doc -m` pass; wiring in `LoadLibrary` correct (and `LoadPackage` restores the caller's package — better than several existing loaders); full `go test ./...` green; CI 20/20 green; `make static-checks` shows only pre-existing `nolintlint` drift in `parser/token/token.go`, untouched here. `lisp/cycleexport.go` reviewed carefully: thin, well-documented wrapper over the kernel's existing cycle-guard protocol, no LVal writes, no seal interaction.

**Heads-up for #374**: that stack carries its own copy of this package; landing #402 first means #374's forward-merge will conflict wholesale on `lisp/lisplib/libelpspath/` and must reconcile its seal-based fuzz invariant with this branch's list-spine narrowing.

---

sam-at-luther:
## ⚠️ Do not merge — confirmed data corruption in this PR

The reviewer's `rangePath.setMutate` finding is **correct and reproducible**. I have verified it independently on a clean build of `d27b466`:

```lisp
(elpspath:?set (vector 1 2 3 4 5) '(range 0 1) (vector 97 98 99))
got:  (vector 97 98 99 98 99 4 5)
want: (vector 97 98 99 2 3 4 5)
```

The source's `2 3` are silently replaced by the replacement's own `98 99`.

### Root cause, confirmed at `path.go:979-982`

```go
vals := append(cells[:from], setCells...)
if to < n {
    vals = append(vals, cells[to:]...)
}
```

The first `append` writes `setCells` **into `cells`' own backing array** starting at `from`. The second line then reads `cells[to:]` out of that same, already-overwritten array. Any range-set whose replacement is longer than the range it replaces returns the wrong answer. It is deterministic, not racy — the private copy always has `cap == len`, so the append never reallocates and always writes in place.

### This is the same defect class the PR exists to fix

Appending into a backing array the function does not own — the shape of #373 and of what #392 turned out to be. A PR whose subject is aliasing bugs shipped one.

### Why the existing verification missed it

`FuzzPathEngine` only ever supplies a **string** value argument, so it is structurally incapable of generating a replacement longer than the range it replaces. The 8007-combination alias battery asserts the *source* is not reached — it does not assert the *result* is correct, so a wrong-but-not-aliasing answer passes it. Both gaps are being closed alongside the fix, and the widened fuzz target is being checked against the unfixed code first so a clean run means something.

Marking this not-ready. Fix, regression tests across replacement-longer/equal/shorter at several offsets over both list and vector sources, a sweep for the same `append`-into-borrowed-backing pattern elsewhere in the package, and the fuzz widening are in progress. I will re-request review when it is green with the widened target proving it bites.

Thanks for catching this — it would have shipped.

---

sam-at-luther:
## ✅ Fixed in `21d4f16` — and my diagnosis above was half wrong

Verified on a clean build of the new HEAD:

```lisp
(elpspath:?set (vector 1 2 3 4 5) '(range 0 1) (vector 97 98 99))
→ (vector 97 98 99 2 3 4 5)                       ✓

(elpspath:?set src '(range 1 2) (list 7 8 9))
→ result (vector 1 7 8 9 3 4 5)   source (vector 1 2 3 4 5)   ✓
```

The splice now builds into an explicitly allocated slice rather than appending onto the input. No three-index trick — correctness rests on owning the slice, not on capacity arithmetic.

### Correcting my previous comment

I wrote that the fuzz target missed this "because it only ever supplies a string value argument." That is **only half the reason, and the less important half.** Widening the value axis alone would not have caught it: invariants 1–5 all watch the **input**, and this corruption lands in the **result**, which nothing read. Closing it needed a new invariant 6 — a top-level range splice must begin with the source's elements before `from` and end with those from `to` on — deliberately silent about the middle, so one check covers `?set`, `?del` and `?nil` alike.

Proof the widened target actually bites: on the unfixed tree it fails at **64.75s / 373,481 execs**, minimising to a 13-byte input (a one-cell insert at `(range 0 0)` yielding `("R0" "R0")` where `("R0" ())` was due). That input is committed as corpus entry `ae76e6821af9dfd8`. On the fixed tree: **1,161,652 execs / 180s clean**.

### The exact trigger, narrower than first stated

Corruption requires **all three**: `len(setCells) > to-from`, `to < n`, and `from+len(setCells) <= cap(cells)`. That third clause is an escape hatch — a replacement large enough to overflow capacity makes the first append *reallocate*, leaving the answer accidentally correct. All three escapes are committed as named controls so the boundary is documented rather than incidental.

### Sweep for the same pattern

Every `append`-onto-a-borrowed-slice site in the package was checked. **One genuine defect**; the rest are sound with reasons, not assumptions:

- `indexPath.deleteMutate` and `rangePath.deleteMutate` — pure left-shifts (`from <= to` guaranteed), so append's memmove reads each cell before overwriting. Probed, not assumed.
- `rangePath.nilMutate` — its replacement is exactly the range's length, so it can never grow past `to`. That is *why* `?nil` was unaffected; now pinned by a test so a future change cannot re-enter the defect this way.
- The copying variants go through `copySeqOffPath`, which does `make([]*lisp.LVal, len(cells))` and owns its slice — verified.

### Why the 8007-combination battery could never have caught it

Two independent reasons, both confirmed empirically (it passes on the unfixed tree):

1. `enumeratePaths` emits only map-key strings and integer indices — **never a range step**, so `rangePath.setMutate` is unreachable from it.
2. Its only oracle is *"the source is unchanged"*, and this defect **respects** that: `Set` splices a private copy, so the source really is intact. The damage is confined to the returned value.

Both remain true after this commit. That axis now belongs to invariant 6 and the new splice battery, and the standing gap is recorded in the commit message.

### Tests and gates

`path_range_splice_test.go`, **83 subtests** — replacement longer/equal/shorter × `from=0`/`from>0` × `to==n`/`to<n` × list/vector source × list/vector replacement, through both `?set` and `?set!`. Red with the fix reverted: **25 of 83 fail**, exactly the longer-plus-`to<n` family. Green: 83/83.

All gates pass on `21d4f16`: `go test ./...`, `-race ./lisp/...`, `golangci-lint` 0 issues, `elps fmt`, confidentiality guard, `ci-gates-test.sh` 126/0, `fuzz-budget-check.sh` (10 min headroom), `FuzzPathEngine` 180s.

### Still open from the review, deliberately not folded in

The `?set`/`?set!` out-of-range divergence that nils a subtree, and #399's by-reference contract shipping undocumented. Both arguably deserve their own commit on this PR before merge — flagging rather than silently expanding scope.

Ready for review again.

---

sam-at-luther:
## govulncheck failure is on `main`, not this PR

`govulncheck` failed on `61bcabe`. It is **not caused by this branch** — every finding is in the Go standard library, reached through files this PR does not touch:

```
GO-2026-5972  encoding/asn1@go1.26.5 → fixed in go1.26.6   via lsp/server.go:184
GO-2026-5026  net/http@go1.26.5      → fixed in go1.26.6   via lisp/x/debugger/engine.go:529
"Your code is affected by 4 vulnerabilities from the Go standard library."
```

Verified by running the scan against **`main` at `95d7e18`** — the commit this PR targets, with none of its changes applied:

```
Your code is affected by 25 vulnerabilities from the Go standard library.
  lsp/server.go:184              ... x509.ParseCertificate
  lisp/x/debugger/engine.go:529  ... tls.Dialer.DialContext
  lisp/error.go:187              ... tls.Conn.Write
  minifier/minifier.go:174       ... tls.Conn.Read
exit status 3
```

(The local run used go1.25.0 against a live vulnerability database, so the count differs from CI's go1.26.5 run. The class is identical: standard-library findings in `lsp/`, `lisp/x/debugger/`, `lisp/error.go` and `minifier/` — none of which this PR modifies.)

Note this check **passed on this same branch at `21d4f16`** a few hours earlier. Nothing in the branch changed to cause it; the advisory database did. These are 2026 advisories published today, and the remedy is a toolchain bump to go1.26.6 — a repo-wide change that should not be smuggled into a bug-fix PR.

Not fixing it here. Flagging it so the failure is not read as this branch's, and so someone can decide on the toolchain bump separately. Once `main` is green again this PR should be re-merged with it so the check re-runs against a fixed base.

The other checks on `61bcabe` are unaffected and still running.

---

sam-at-luther:
Superseded by **#472**, which is this branch rebased from `f18e118` onto `8402c49`. The body here is 32 commits stale; #472's is written against today's `main`.

## The split was attempted and is not coherent

The task was to extract the three bug fixes into a `main`-based PR and leave the framework for a separate decision. **That cannot be done, and nothing was extracted.** `libelpspath` does not exist on `main`:

```
$ git ls-tree -r --name-only origin/main | grep -i elpspath
(no output)

$ git grep -n -E 'func (copyMap|copyList|copyVector|storeCells)' origin/main -- '*.go'
(no output)
```

Every file in this PR is a new file under `lisp/lisplib/libelpspath/`, plus 5 lines in `lisp/lisplib/lisplib.go` to register the package. A PR carrying "only the fixes" against `main` would be an empty diff: there is no `copyMap` to make deep, no `storeCells` to teach about list layout, and no walker to guard.

#393 already recorded this — *"`libelpspath` **did not exist in the tree #391 swept** … the **only** non-test lisplib package present on the chain and absent from `main`"* — and it settles the question. The fixes were separable from the sealed-AST feature, which is what this PR did. They are not separable from the package, because they are corrections applied to code as it is adopted. **All-or-nothing**, so #472 is the whole thing rebased.

## What #472 changes relative to here

**Dropped: `d27b466`** ("sixth fuzz shard for the 21st target"). It was written when `main` was at 20 targets over 5 shards. `main` is now at **29 over 8**. Replaying it would shrink the matrix from 8 shards to 6, and per #458 the gate would not have said so:

| shards | targets | largest shard | required | timeout | headroom | `fuzz-budget-check.sh` |
|---|---|---|---|---|---|---|
| 6 (this commit replayed) | 30 | 5 | 60 min | 60 min | **0 min** | `the sweep fits`, **exit 0** |
| 8 (#472) | 30 | 4 | 50 min | 60 min | 10 min | `the sweep fits`, exit 0 |

Both measured, not reasoned. Replaced by a commit that re-derives the budget for `FuzzPathEngine` as the 30th target — `ceil(30/8) = 4`, no shard needed — and corrects the derived summary line, which still read *"23 targets over 6 shards"* while the matrix said 8.

**Kept: `da21e81`** ("unexport the cycle guard"). It was flagged as touching `lisp/` core, but against `main` it does not: `lisp/cycleexport.go` has never existed there. It is added by `68efb34` earlier in this same branch and removed again by `da21e81`, so the two are intra-PR churn and the net change to package `lisp` is nil. Kept rather than squashed, because its message records why the guard is not exported.

**Added: a capacity clamp on `rangePath.Get`.** This PR's in-file comment declines to clamp *"because the kernel's own `(slice 'vector …)` hands back exactly the same thing."* **That premise is false on today's `main`** — #373 settled it the other way and `lisp.clampCap` now runs at every kernel producer. Measured before the clamp:

```
(set 'w (? v '(range 0 3)))     (append! w 99)   ->  src (vector 1 2 3 99 5)   <- this Get
(set 'w (slice 'vector v 0 3))  (append! w 99)   ->  src (vector 1 2 3 4 5)    <- the kernel
```

Landing this branch unclamped would have introduced the tree's only remaining producer of a class `main` had just closed.

## Corrections to the numbers in this body

- **Cycle guard cost.** This body reports *"exactly +1 allocation per walk, +16 B/op"*. Re-measured on the rebased tree, interleaved, `-count=12 -cpu=1`: **zero** allocation difference — `all samples are equal` on `B/op` and `allocs/op` for all four rows. `da21e81`'s move to caller-owned state removed the allocation. Time is not free: +1.7% to +3.3%, geomean +2.5%, all `p=0.000` and all larger than their own row's spread.
- **Fuzz shard paragraph.** The "Also included" section describing the sixth shard no longer applies; see above.
- **`#392 deliberately not ported`.** Still accurate, and the control it cites still holds.

## Status

All three defects still reproduce on the rebased tree, each red-proofed by reverting its own fix — evidence in #472. Gates all green there, including `make test-elpscheck`, `make test-examples`, `ci-gates-test.sh` (233 passed, 0 failed) and the confidentiality guard.

One new defect was found during the neighbourhood audit and filed rather than fixed: **#471** — `?del!` compacts through a backing array it does not own, so deleting through a *view* scrambles the sequence it aliases. Not covered by the alias battery here, which writes through copies rather than views.

**Recommendation: close this PR in favour of #472** once that lands. Nothing here is not there, apart from `d27b466`, which should not land.

---

sam-at-luther:
**#472 is merged** — `1c13c76` on `main`. Everything in this PR is now on `main` except `d27b466`, which was deliberately dropped and should not land: on today's tree it would take the fuzz matrix from 8 shards to 6, and `fuzz-budget-check.sh` would still report `the sweep fits` at exactly zero headroom (#458).

This PR carries nothing else that is not already on `main`. **Safe to close as superseded.** Leaving that to you rather than closing it from here.

The body above is kept as the historical record of the pre-rebase state; the corrections to it — the cycle-guard allocation number, the sixth-shard paragraph — are in the previous comment, and #472's body is the version written against today's `main`.

---

sam-at-luther:
Closing as superseded by **#472**, which merged as `1c13c76`. `libelpspath` is on `main` — 22 files under `lisp/lisplib/libelpspath/` — with #393, #395 and #401 fixed, plus the `rangePath.Get` capacity clamp that the rebase established was needed after #373 changed the kernel's behaviour.

Verified before closing rather than taken on trust. Against `main` at `1c13c76`, this branch's net diff is:

```
112 files changed, 816 insertions(+), 16386 deletions(-)
```

It is overwhelmingly a **deletion** of work that landed today — the benchstat waiver test corpus (`scripts/testdata/waivers-*.txt`, the parallel-regression fixture) and much else. Merging it now would revert roughly 16k lines.

The specific hunk that must not land, confirmed present:

```diff
-    name: Fuzz shard ${{ matrix.shard }}/8
+    name: Fuzz shard ${{ matrix.shard }}/6
-        shard: [1, 2, 3, 4, 5, 6, 7, 8]
+        shard: [1, 2, 3, 4, 5, 6]
-        run: ./scripts/fuzz.sh --shard ${{ matrix.shard }}/8
+        run: ./scripts/fuzz.sh --shard ${{ matrix.shard }}/6
```

At 30 targets that takes the sweep to `ceil(30/6) = 5 × 10m + 10m = 60 min` against a 60-minute timeout: zero headroom, and per #458 `fuzz-budget-check.sh` would still print `the sweep fits` and exit 0. A silently truncated sweep with a green gate.

The branch and its history stay for reference. The body above is the pre-rebase record; its two stale claims — the cycle-guard allocation figure and the sixth-shard paragraph — are corrected in the comments, and #472's body is the version written against today's `main`.

One item from this work is **live on `main` and not yet fixed**: #471 (`?del!` compacts through a backing array it does not own, so deleting through a view scrambles the sequence it aliases). That was merged knowingly, with the fix to follow, and an aliasing sweep of the whole package is in progress against the four defect families closed today.

---

Issue #350, opened 2026-08-10
json:load silently rounds integers above 2^53, and there is no way to load one exactly


sam-at-luther:
`json:load-string` and friends decode every JSON number through `float64`. Integers beyond 2^53 are therefore rounded on the way in, silently — no error, no warning, no condition raised. The value is simply wrong.

This is long-standing, not a recent regression: `case float64: return lisp.Float(x)` in `loadInterfaceMax` dates to the initial 2018 commit (0281c7f).

## It was already found, and recorded nowhere findable

The fuzzing work characterised this and then scoped it out. From `lisp/lisplib/libjson/fuzz_test.go`:

> NOT asserted: that Load(b) preserves the numeric text of b. encoding/json decodes every JSON number into a float64 here, so integers beyond 2^53 are rounded on the way in. **That is a known, deliberately out-of-scope defect**; asserting text preservation would turn it into permanent fuzz noise.

`FuzzDumpJSON` lists it among the reasons it cannot assert round-trip fidelity, and two seed-corpus files are committed for it (`testdata/fuzz/FuzzLoadJSON/integer-above-2pow53` and `…-string-numbers`).

Scoping it out of the harness was the right call — asserting text preservation really would have been permanent noise. Leaving it recorded only in a Go comment was not. It is not on #318 and has no ticket, so the next person to hit it rediscovers it from scratch. That is exactly the failure mode #318 exists to prevent.

## Reproductions

Verified on `main` at 5bb31c9 (v1.49.0).

| expression | result |
|---|---|
| `(json:load-string "9007199254740993")` | `9007199254740992` — 2^53+1, off by 1 |
| `(json:load-string "9223372036854775807")` | round-trips as `9223372036854776000` — off by 1,193 |
| `(json:dump-string (json:load-string "{\"id\":9223372036854775807}"))` | `{"id":9223372036854776000}` |

Secondary, and probably worth its own discussion: **every** JSON number decodes to a float, not just large ones.

| expression | result |
|---|---|
| `(type (json:load-string "42"))` | `'float` |
| `(int? (json:load-string "42"))` | `false` |
| `(json:load-string "42")` | prints as `42` |
| `(= 42 (json:load-string "42"))` | `true` |

It prints like an integer and compares equal to one, which is a large part of why this has gone unnoticed for so long.

## `:string-numbers` is a workaround, not a fix

[...]

PR #408, opened 2026-08-14
libjson: add :exact-integers so JSON integers above 2^53 survive a load (#350)

sam-at-luther:
Addresses #350.

**Position 3 of a four-PR bug-crush chain.** Stacked on #407. Review only the top commit. Diff is confined to `lisp/lisplib/libjson/` — nothing in `lisp/` core, nothing seal- or elpspath-related.

## Read this first: the P0 is not fixed by default. It is made *fixable*.

This ships an **opt-in** `:exact-integers` option. Default `json:load` still silently rounds integers above 2^53. That is a deliberate choice for a consensus-sensitive codebase, and it is the main thing to agree or disagree with in review.

**Why not just flip it.** The blast radius of `UseNumber` is *every number in every document*, not only the broken ones. `{"count":3}` stops being a float. `(type x)` goes `float` → `int` fleet-wide, and with it `int?`/`float?`, `nth` (which rejects a float index today), and any branch on type. Two nodes on different elps versions would disagree about what the same bytes *mean*.

Two things that would have made a flip catastrophic were checked, and neither holds: `(/ 7 2)` is `3.5` regardless of operand types, so there is no integer-division divergence; and `to-string`/`format-string` agree for every value below 2^53. So a default flip is *survivable* later — but it buys nothing a per-call-site opt-in doesn't, while an opt-in lets a phylum migrate one call at a time.

`TestIssue350HidingMechanism` and `default-rounds-silently` pin the broken default **on purpose**, so the eventual flip cannot happen quietly — it must rewrite those tests in the same commit.

## The bug, reproduced end-to-end

`json:dump-string` of `json:load-string` of a literal, on `main` at `f18e118` vs this branch:

| input | main **and branch default** | branch `:exact-integers` |
|---|---|---|
| `9007199254740993` | `9007199254740992` | `9007199254740993` |
| `9223372036854775807` | `9223372036854776000` | `9223372036854775807` |
| `-9223372036854775808` | `-9223372036854776000` | `-9223372036854775808` |

The int64-max drift is **193**, not 1,193 as triage recorded (`9223372036854776000 − 9223372036854775807`). Pinned with `math/big` in `TestDumpFloatIsWhereTheDigitsGo`.

**`json:dump` is not the culprit.** `encodeLInt` uses `strconv.AppendInt` and is exact for every int64 in both `:string-numbers` modes. By the time `Dump` sees the value it is already a float and the digits are gone — which is why the fix is on the load side.

## The rule is syntactic, for consensus

Integer-shaped means *no `.`, no exponent* — a property of the bytes, never of the value they denote. Every node reading the same document reaches the same answer without depending on shared floating-point or decimal-arithmetic behaviour.

## Red-then-green

`TestPackageExactIntegers` (`libjson_integer_test.lisp`) uses only the lisp surface, so it compiles and runs on `f18e118`. There it fails 6 subtests with `json:load-string: unrecognized keyword argument: exact-integers`; on this branch all 11 pass.

Being honest about what that is: for an opt-in feature, "red on main" means *feature absent*, not *behaviour wrong*. The behavioural evidence is the table above, carried by tests that pass on **both** sides (`TestLoadDefaultUnchanged`, `default-rounds-silently`) and assert the wrong output literally — which is what makes them useful as a revert check.

## What an adopter gets — release note

> ### `json:load` — new `:exact-integers` option (issue #350)
>
> **Nothing changes unless you opt in.** Default `json:load-string`/`load-bytes`/`load-message`, `:string-numbers`, all `json:dump-*`, and the Go `Load`/`LoadMax`/`Dump` signatures behave exactly as before. Verified rather than asserted: 120,118 decode results across both pre-existing modes and 450,000 encode results spanning the full float64 bit-pattern space are byte-identical to `f18e118`, and allocation counts on every pre-existing benchmark are unchanged to the byte.
>
> For a caller who turns it on:
>
> 1. A JSON number written as an integer decodes to an **int**, not a float — *every* such number, not only large ones. `(type x)` returns `'int`, `(int? x)` is true, and the value is usable where an int is required (e.g. `nth`'s index).
> 2. Integers above 2^53 keep their exact value and re-serialize with their exact digits. `to-string` changes from `9.007199254740992e+15` to `9007199254740993`.
> 3. An integer literal too large for a lisp int raises the catchable condition **`json:integer-range-error`** instead of silently becoming a rounded float. Documents that loaded before can now fail — that is the point.
> 4. Numbers with a fraction or exponent are **unchanged** and still decode as floats: `1.5`, `1.0`, `1e2`, `1E2`, `-0`. `-0` is deliberately excluded so it keeps re-serializing as `-0`.
> 5. **Known asymmetry.** The rule is syntactic and `json:dump` normalizes float text, so `100e7` decodes as a float, dumps as `1000000000`, and a *re-read* makes it an int. The value is correct at every step and stable from the second read on; what changed is the document, not the rule. All nodes reading the same bytes still agree — a phylum-authoring footgun, not a consensus hazard. Pinned by `TestLoadExactExponentFormNormalises`. Machine-generated JSON does not hit it: Go, JavaScript and Python all render `1e9` as plain digits.
> 6. **One deliberate exception to rule 3.** An oversized integer literal is accepted as a *float* when the literal is already that float's canonical text (e.g. `10000000000000000000`). Nothing is discarded, so nothing is hidden. This is required: elps renders every float in [2^63, 1e21) as plain digits, so without it a phylum holding an ordinary float of `1e19` could dump its state and be unable to read it back. `9223372036854775808` is *not* canonical text and still fails loudly.
> 7. `:string-numbers` takes precedence when both are set.
> 8. Malformed input stays catchable as `json:syntax-error`. The option needs a streaming decoder (`UseNumber` exists only there), which reports empty input and trailing content differently from `json.Unmarshal`; those are normalized so an adopter's `handler-bind` does not quietly stop firing.
>
> **Not fixed by default.** Flipping the default changes `(type x)` for every integer in every document on every node, and is a separate scheduled change.

## Round-trips and fuzzing

Byte-identical load→dump for 2^53 ± 4, −2^53 ± 4, int64 max/min and four neighbours each, plus 0, ±1, 2^31, ±2^62 — then re-loaded to prove stability. Also nested in arrays and objects. `FuzzLoadExactIntegers` asserts byte-identity for any bare integer literal: 2M execs clean.

The fuzzer earned its place: it surfaced two design defects *before* commit — exponent-form normalization, and the encoder emitting documents the option would then refuse, which would have been a liveness bug in substrate.

## Benchstat — allocation is the reliable signal here

Interleaved, n=10, `GOMAXPROCS=4`.

```
                       │     base     │                  pr                  │
                       │  allocs/op   │  allocs/op    vs base                │
Encode-4                 214.0 ± 0%     214.0 ± 0%    ~ (p=1.000 n=10)
Encode_stringNumbers-4   30.00 ± 0%     30.00 ± 0%    ~ (p=1.000 n=10)
Load/default-4           135.0 ± 0%     135.0 ± 0%    ~ (p=1.000 n=10)
Load/stringNumbers-4     141.0 ± 0%     141.0 ± 0%    ~ (p=1.000 n=10)
LoadIntegers/default-4   49.00 ± 0%     49.00 ± 0%    ~ (p=1.000 n=10)
Load/exactIntegers-4                    141.0 ± 0%     (new)
LoadIntegers/exact-4                    53.00 ± 0%     (new)
```

Every pre-existing path allocates byte-for-byte as before. The `UseNumber` cost lands only in the new arm and is the cost `:string-numbers` already pays. **Wall-clock on this host is too noisy to claim anything** (±25–55% even at 500 ms/op) and is not dressed up as a result. The decode direction had no benchmark at all before this.

## Out of scope, reported not fixed

Two unrelated defects found in the deprecated Go walkers, each deserving its own issue rather than riding a consensus-sensitive PR:

- **`GoInt`/`GoFloat64` have an inverted guard** — `if v.IsNumeric() { return 0, false }` refuses every number and *accepts* every non-number. `GoInt(Int(42))` → `(0, false)`; `GoInt(String("hello"))` → `(0, true)`. The accept-a-string case is the same class as #350 — a silent wrong number instead of an error — by a different mechanism. One-character fix, but it changes a deprecated exported API.
- **`GoSlice`/`GoMap` are unguarded entry points** into the unbounded walk `GoValue`'s doc comment warns about (#390 / #391), without carrying the warning.

`GoValue`/`GoSlice`/`GoMap` do **not** share the #350 defect — they walk LVal→Go, so there is no float64 funnel.

## Gates

`go build`, `go vet`, `go test ./...`, `golangci-lint run ./...` (0 issues), `gofmt -l` (byte-identical to base), `confidentiality-guard.sh`, `ci-gates-test.sh` (78 pass / 0 fail), `fuzz-budget-check.sh` (fits, headroom unchanged), `elps doc -m` ("All functions and exports are documented"). Full suite and `-race` re-run by me at the stack tip after rebasing.

---

### DIFF
--- a/lisp/lisplib/libjson/encode.go
+++ b/lisp/lisplib/libjson/encode.go
@@ -310,17 +310,30 @@ func (enc *encoder) encodeFloat(x float64) error {
 }
 
 // scratchFloat encodes x to enc.scratch and returns a slice of that array.
-//
-// NOTE:  scratchFloat is adapted from floatEncoder.encode in encoding/json,
-// simplified to only work with native float64 values.
-// https://cs.opensource.google/go/go/+/refs/tags/go1.16.4:src/encoding/json/encode.go;l=575
 func (enc *encoder) scratchFloat(x float64) []byte {
+	return appendJSONFloat(enc.scratch[:0], x)
+}
+
+// appendJSONFloat appends the canonical JSON text of x to b.
+//
+// "Canonical" is meant literally and is depended upon outside the encoder:
+// this is the ONLY rendering of a float this package ever emits, so the decode
+// side can ask whether a number literal it is looking at is already the
+// canonical text of the float it parses to.  That question is what separates a
+// document elps can read back unchanged from one it can only round -- see
+// loadNumber in json.go.  Any caller that formatted floats separately would
+// let the two sides drift, and the drift would show up as this package
+// refusing to load its own output.
+//
+// NOTE:  adapted from floatEncoder.encode in encoding/json, simplified to only
+// work with native float64 values.
+// https://cs.opensource.google/go/go/+/refs/tags/go1.16.4:src/encoding/json/encode.go;l=575
+func appendJSONFloat(b []byte, x float64) []byte {
 	// Convert as if by ES6 number to string conversion.
 	// This matches most other JSON generators.
 	// See golang.org/issue/6384 and golang.org/issue/14135.
 	// Like fmt %g, but the exponent cutoffs are different
 	// and exponents themselves are not padded to two digits.
-	b := enc.scratch[:0]
 	abs := math.Abs(x)
 	fmt := byte('f')
 	// NOTE:   Because ELPS only natively supports float64 values the exponent
@@ -328,13 +341,14 @@ func (enc *encoder) scratchFloat(x float64) []byte {
 	if abs != 0 && (abs < 1e-6 || abs >= 1e21) {
 		fmt = 'e'
 	}
+	n := len(b)
 	b = strconv.AppendFloat(b, x, fmt, -1, 64)
 	if fmt == 'e' {
 		// clean up e-09 to e-9
-		n := len(b)
-		if n >= 4 && b[n-4] == 'e' && b[n-3] == '-' && b[n-2] == '0' {
-			b[n-2] = b[n-1]
-			b = b[:n-1]
+		m := len(b)
+		if m-n >= 4 && b[m-4] == 'e' && b[m-3] == '-' && b[m-2] == '0' {
+			b[m-2] = b[m-1]
+			b = b[:m-1]
 		}
 	}
 	return b
--- a/lisp/lisplib/libjson/fuzz_test.go
+++ b/lisp/lisplib/libjson/fuzz_test.go
@@ -31,13 +31,17 @@ import (
 //     against the first, which is the strongest form that is actually
 //     guaranteed -- see the note on integers below.
 //
-// NOT asserted: that Load(b) preserves the numeric text of b.  encoding/json
-// decodes every JSON number into a float64 here, so integers beyond 2^53 are
-// rounded on the way in.  That is a known, deliberately out-of-scope defect;
-// asserting text preservation would turn it into permanent fuzz noise.
-// Comparing decode-of-re-encode against the FIRST decode sidesteps it: the
-// rounding has already happened before the comparison starts, and float64 ->
-// shortest-repr -> float64 is exact.
+// NOT asserted HERE: that Load(b) preserves the numeric text of b.  In the two
+// modes this target covers, encoding/json decodes every JSON number into a
+// float64, so integers beyond 2^53 are rounded on the way in (issue #350).
+// This target measures the DEFAULT, which still rounds on purpose, so
+// asserting text preservation would be permanent noise.  Comparing
+// decode-of-re-encode against the FIRST decode sidesteps it: the rounding has
+// already happened before the comparison starts, and float64 -> shortest-repr
+// -> float64 is exact.
+//
+// FuzzLoadExactIntegers, below, covers the mode where text preservation IS the
+// contract and asserts it directly.
 
 func stringNumberModes() []bool { return []bool{false, true} }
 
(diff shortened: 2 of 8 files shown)


sam-at-luther:
**Correction to the "Out of scope, reported not fixed" section of this PR body — both defects it names are not real.** Recording it here so nobody files or chases them.

**1. `GoInt`/`GoFloat64` "inverted guard".** The body claims the code reads `if v.IsNumeric() { return 0, false }`. It does not, and never has. `lisp/embed.go` on `main` reads:

```go
func GoInt(v *LVal) (int, bool) {
	if !v.IsNumeric() {
		return 0, false
	}
	...
```

`git log -S 'if v.IsNumeric() {' -- lisp/embed.go` returns **no commits**, so the inverted form is not something that was fixed in between — it never existed. Verified by running it against `main`:

- `GoInt(Int(42))` → `(42, true)`
- `GoInt(String("hello"))` → `(0, false)`
- `GoFloat64(Float(1.5))` → `(1.5, true)`
- `GoFloat64(String("hello"))` → `(0, false)`

All four are the correct results. The claimed "accepts a string, returns a silent wrong number" behaviour does not occur.

**2. `GoSlice`/`GoMap` "unguarded entry points".** Both carry cycle guards, added by #391 (`050ade5`). `GoSlice` and `GoMap` each construct a `cycleState`, pass a `cycleGuard` into the recursive walk, and return `(nil, false)` when `st.cyclic` is set. Verified by calling `GoSlice` on a self-referential `SExpr`: it returns normally rather than exhausting the stack.

Everything else in this PR body stands — the `:exact-integers` work itself was re-verified when it landed on `main` via #416, and `go test -count=1 ./...` is green there.

---

Issue #435, opened 2026-08-15
TestEvalTerminatingSeedsComplete enforces a correctness property with a wall-clock budget, and flakes under -race


sam-at-luther:
> **Note on this description.** The original text was overwritten in error by an automated edit on 2026-08-15 (an `update` call where a comment was intended). What follows is a **reconstruction** from the reporting agent's own evidence, not the original wording — the original remains in this issue's edit history via the "edited" dropdown next to the title, and restoring it from there is still worthwhile. The technical content below is faithful to the evidence gathered; only the phrasing is not the author's.

## The defect

`TestEvalTerminatingSeedsComplete` asserts a **correctness** property — that its terminating seeds terminate — but enforces it with a **2s wall-clock `context.WithTimeout`**.

The `fuzzwatch` watchdog beside it in the same file is deliberately denominated in *scheduled* time, for exactly this reason. The assertion is the outlier.

## Why it is a defect and not just a slow test

On a loaded machine the deadline fires and the failure is reported as an evaluator problem — that a budget "has been tuned into the range of correct programs." It accuses the code under test of a bug that is not there.

## Evidence — reproduced on unmodified `main`

Reproduced on pristine `ce9798d` with all changes stashed and new test files removed, and independently on `bfb6ee8`:

- **2 of 6 concurrent runs fail.**
- The same fixed seed stops at a **different place every time**. Across eight independent samples gathered by three agents: **29k / 49,792 / 56k / 61,723 / 73,618 / 89,062 / 92k / 108,562** steps.

A seed-dependent bound would stop in the same place on every run. The variance is the clock.

Further observations added since:

- It flakes under plain `make go-test` under load, not only under `-race` — so the cause is **CPU contention**, not race-detector overhead.
- It does **not** fire in CI, which is why it has survived: CI runners are comparatively quiet.
- Step count at failure varies roughly 35% between runs.

## Cost already incurred

Three separate agents working on unrelated changes hit this, and each had to reproduce against a clean tree to prove the failure was not theirs. All three concluded correctly that it was pre-existing — but only after spending the time. A gate that fires on innocent code trains people to re-run until green, which is how a real regression eventually gets waved through.

#436 was filed independently for this same defect and is closed as a duplicate; its evidence is merged above and in the comments.

## What a fix must preserve

[...]

Issue #434, opened 2026-08-15
elpstest.Runner.RunBenchmark times the benchmark file's load only when a SetupFn is configured


sam-at-luther:
## The defect

`elpstest.Runner.RunBenchmark` charges the whole benchmark file's load — parsing it and evaluating every top-level form in it, including every `benchmark` definition — to the timed region, but only when the Runner has a `SetupFn`.

`elpstest/lisptest.go`:

```go
func (r *Runner) RunBenchmark(b *testing.B, i int, path string, source io.Reader) {
	b.StopTimer()
	env, err := r.NewEnv(b)
	...
	if r.SetupFn != nil {
		b.StopTimer()
		r.SetupFn(env)
		b.StartTimer()          // <-- timer is now ON
	}

	err = lisp.GoError(env.Load(filepath.Base(path), source))   // timed iff SetupFn != nil
	...
	suite := libtesting.EnvTestSuite(env)
	ltest := suite.Benchmark(i)
	b.StartTimer()                                              // no-op if already on
	err = lisp.GoError(env.Eval(lisp.SExpr([]*lisp.LVal{ltest.Fun, lisp.Int(b.N)})))
}
```

The `SetupFn` branch ends in `StartTimer` rather than restoring the timer to the state it found it in. Without a `SetupFn` the timer stays stopped until just before the benchmark body is evaluated, which is the intent. With one, `env.Load` — an arbitrary amount of work, proportional to the size of the whole `.lisp` file rather than to the benchmark being measured — lands inside the measurement.

Two Runners that differ only in whether they set `SetupFn` therefore report numbers that are not comparable, and the difference has nothing to do with the code under benchmark. Nothing signals it.

## Scope

No Runner in this repository sets `SetupFn` — `libjson_test.go`'s is the only `RunBenchmarkFile` caller and leaves it nil — so no number produced in-tree is affected. `elpstest` is an exported package, though, and `SetupFn` is an exported field documented for exactly this use, so a downstream Runner that sets one gets the inflated measurement.

## Suggested fix

Make the `SetupFn` branch timer-neutral — stop the timer again after `SetupFn` returns, since the surrounding code had already stopped it — so `env.Load` stays untimed on both paths.

## Verification note

**Found by reading, not by running.** Whoever picks this up should write the red test first — a Runner with a no-op `SetupFn` against one without, on a benchmark file large enough for the load to dominate — and confirm the reported numbers differ. If it does not reproduce, close this rather than fix it.

## Related

[...]

PR #441, opened 2026-08-15
elpstest: give every benchmark iteration a fresh AST (#365)

### LINKED ISSUE #365: elpstest.RunBenchmark evaluates one parsed AST in a fresh Runtime per iteration

sam-at-luther:
Fixes #365

## What the defect actually costs

The issue reports that `RunBenchmark` parses once and then evaluates that one tree in every `b.N` iteration under a fresh `Runtime`. That is accurate. The question worth answering before fixing it is whether the harness is *lying* about its numbers or merely sharing something it should not, so this was measured rather than argued.

**The reported numbers are honest today, and no historical number in this repository is wrong.** Two measurements say so:

1. **Environment construction is not being measured.** `b.StopTimer()` covers the per-iteration `NewEnv` + `InitializeUserEnv`, and `StopTimer` stops allocation accounting as well as the clock. Measured on this machine, building one benchmark environment costs ~900µs, while `RunBenchmark` on `(dotimes (n 100) (+ 1 2 3))` reports 128µs/op. If setup leaked into the measurement the reported figure could not be below it.

2. **No in-tree benchmark mutates its AST.** `RunBenchmark` was instrumented to snapshot every AST node (type, `Str`, `Int`, `Float`, `Quoted`, `Spliced`, `FunType`, `Native` type, `Source`/`Meta`/`MacroExpansion` pointers, cell count and capacity, and each node's own address) before the loop and after every iteration, and the whole benchmark set was run under it — `go test -run '^$' -bench . -benchtime=4x ./...`, all packages. Zero mutations. Every benchmark in the tree evaluates a tree that comes back byte-identical.

**But the mechanism is live, not theoretical, and it is reachable from ordinary ELPS.** A quoted literal is an AST node and `stable-sort` sorts in place — pinned as current, deliberate behaviour by `lisp.TestSliceViewSharesElements` ("stable-sort mutates a quoted literal in place"). Under the instrumented harness, `(stable-sort < '(5 3 1 4 2))` mutates the parsed tree on iteration 1:

```
--- FAIL: BenchmarkProbeSortLiteralDirect
    PROBE: AST MUTATED after iteration 1
      line 4:
       before:   p=0xc0000c9110 t=int i=5 ...
       after:    p=0xc0000c91f0 t=int i=1 ...
```

So a benchmark whose source sorts a literal sorts unsorted input on iteration 1 and already-sorted input on iterations 2..N — the harness silently substitutes a different workload for the one the benchmark declares.

### Quantified: what such a benchmark would report

Two arms of the same measured region, differing only in whether the evaluated tree was shared with earlier iterations, over `(stable-sort < '(...))` on a 1500-element shuffled literal, `-count=6`:

| arm | sec/op (median of 6) | B/op | allocs/op |
|---|---|---|---|
| shared tree (main) | 5.07 ms | 1.18 MB | 15,140 |
| fresh tree (this PR) | 50.67 ms | 12.29 MB | 157,531 |
| **understatement** | **~10.0x** | **~10.4x** | **~10.4x** |

The shared arm's six samples were 2.79, 2.82, 4.81, 5.34, 8.47 and 11.95 ms — a 4.3x spread, and it tracks `b.N` rather than tracking noise: the 11.95 ms sample ran 222 iterations and the 2.82 ms sample ran 508, because the single honest iteration is amortised over more and more dishonest ones as `b.N` grows. A number that moves with `b.N` is the signature of the defect. It is not a noisy measurement of one thing; it is a weighted average of two different workloads, and the weighting is chosen by the benchmark framework.

The fresh arm, by contrast, is stable at 45–55 ms across five of its six samples — it measures the same workload every time.

Nothing in the harness signals any of this. The benchmark passes, benchstat compares it happily, and the gate cannot fire because both arms of a comparison share the flaw. That is the argument for fixing it: not that the current numbers are wrong, but that the harness cannot be trusted to keep the next ones right, and there is no failure mode that would tell anyone.

## The fix

The issue's fix, unchanged: copy the expressions per iteration, outside the timed region.

```go
iterExprs := make([]*lisp.LVal, len(exprs))
for i, expr := range exprs {
	iterExprs[i] = expr.Copy()
}
b.StartTimer()
```

This is what `lisp.TextLoader` already does for the same reason — `env.Eval(expr.Copy())` on every load, so two loads of one file cannot share nodes.

The copy is charged to nobody: it happens with the timer stopped, so it lands in neither `ns/op` nor `B/op`/`allocs/op`. Measured, it costs ~3.9µs for a small tree against the ~900µs environment build that already sat in the same untimed region — a 0.4% addition to per-iteration setup.

## Red evidence, on `ce9798d` with only the test file added

```
=== RUN   TestRunBenchmarkGivesEachIterationAFreshAST
    runbenchmark_test.go:73: RunBenchmark failed: an iteration evaluated an AST that an earlier iteration had already mutated.
        testing.Benchmark reports N=0 when the benchmark function calls Fatal and discards its output, so the
        same source evaluated twice against one parsed tree was re-run here, and the second pass reported:

          benchmark:9:7: stale-ast: iteration began with an AST a previous iteration mutated
--- FAIL: TestRunBenchmarkGivesEachIterationAFreshAST (0.04s)
=== RUN   TestQuotedLiteralIsMutatedByEval
    runbenchmark_test.go:157: premise holds: same tree, two runtimes, run 1 returned '(3 1 2) and run 2 returned '(1 2 3)
--- PASS: TestQuotedLiteralIsMutatedByEval (0.00s)
```

The catch drives `RunBenchmark` through `testing.Benchmark` on a source that checks its own literal is pristine at the top of each iteration and then sorts it. On `main` iteration 2 raises `stale-ast`, `RunBenchmark` calls `b.Fatalf`, and `testing.Benchmark` reports `N=0`. Because `testing.Benchmark` discards the failed benchmark's output, the failure message re-runs the same source against one shared tree so the log says what actually happened rather than just "N=0".

`TestQuotedLiteralIsMutatedByEval` is a **guard**, not a catch, and is labelled as such in-file: it passes on `main`. It pins the premise the catch depends on — that evaluation can leave state in the tree — without going through `RunBenchmark`. If ELPS ever copies quoted literals on evaluation, that guard skips with an explicit "this detector needs replacing" message instead of leaving the catch passing for a reason it did not intend.

The test pins `-test.benchtime` to `20x` for its own `testing.Benchmark` call and restores it afterwards. Left on the default 1s, a benchmark this small drives `b.N` into the tens of thousands chasing a time budget it can only accrue while the timer is running, and pays the untimed environment build on each one: 8.5s of wall clock to observe a property that appears on iteration 2. With the bound it takes 0.18s.

## Benchmark impact

This change alters the harness that 43 of the repository's benchmarks run through, so the reasonable expectation is that every one of their numbers moves. **None of them do**, and that is the point of putting the copy outside the timed region rather than inside it.

`scripts/bench-run-arms.sh` with `BENCH_COUNT=6`, base `ce9798d`, arms interleaved. `bench-arms-check.sh`: same goos/goarch/cpu, **105 paired benchmarks** (base 105, pr 105 — nothing unpaired), 6 samples each. Gate verdict:

```
benchstat-gate: interpreted 16 delta row(s) + 321 no-change row(s); 1 significant move(s)
in the bad direction; 0 at or above the gate (timing 15%, allocation 5%).
```

**0 rows at or above the gate.** The allocation metrics are the load-bearing evidence, because they are the ones the gate's own documentation calls effectively deterministic:

| package | `B/op` geomean | `allocs/op` geomean |
|---|---|---|
| `lisp` | +0.00% | -0.00% |
| `lisp/lisplib/libjson` | +0.04% | +0.12% |
| `parser/rdparser` | -0.00% | +0.00% |

The single significant bad-direction move in the whole comparison is `lisp` `MacroDefun-4` `B/op` **+0.02%** (p=0.041), which is 33.32Ki against 33.32Ki — a rounding artefact two orders of magnitude below the 5% gate. This flatness is itself a measurement: `StopTimer` stops allocation accounting as well as the clock, so a per-iteration deep copy of the AST — which is thousands of bytes of real allocation — shows up as nothing at all in `B/op`. If the copy had been placed after `StartTimer` these rows would all have moved.

**The timing rows are not interpretable in this sandbox and no conclusion is drawn from them.** The machine was running unrelated concurrent fuzz jobs at load ~35 on 4 cores throughout, which shows up directly in the per-row spreads: `AliasSliceVector-4` reports ±1674% on the base arm, `Package/dump-array-4` ±388%. The `sec/op` geomeans (`lisp` -8.23%, `libjson` -15.34%, `rdparser` +5.00%) carry no p-value, are informational by the gate's design, and point in both directions across packages, which is what noise looks like rather than what a harness change looks like. The three individually significant timing rows are all *improvements* of 23–44% with base-arm spreads of 86–388% — noise again, and in the direction that cannot be a regression. Interleaving the arms is what keeps that noise from pooling into one side.

**No new benchmark was added**, so the structural gap where a new benchmark has no paired baseline and the gate therefore cannot fire on it does not arise here: all 105 rows are paired, and the two new tests are `Test` functions, not `Benchmark` functions.

The one number that genuinely moves is untimed wall clock. Each iteration now does one extra deep copy of the parsed source: ~3.9µs for a small tree, against the ~900µs environment build already in the same untimed region.

The **Benchmark Comparison** check on this PR is the cleaner measurement of the same question — a quiet dedicated runner at n=10 rather than a contended sandbox at n=6 — and it adjudicates the change on its own.

## Neighbourhood audit

The brief was every other place `elpstest` builds a Runtime inside a measured loop. Sites examined and **not** changed:

- **`Runner.RunBenchmark`** (`elpstest/lisptest.go`) builds exactly one environment per benchmark, outside any loop, and passes `b.N` into the lisp-level benchmark function so the iteration happens inside ELPS. There is no Go loop to construct a runtime in. Its source is re-read and re-loaded per call through `env.Load`, which goes through `TextLoader` and therefore already copies.
- **`RunBenchmarkFile`'s `$load` sub-benchmark** does build a full runtime — `NewEnv` plus `lisplib.LoadLibrary` plus `env.Load` — inside a timed `for range b.N` with no `StopTimer`. That is correct and deliberate: `$load` measures loading, and constructing the environment is part of what loading costs. It parses from a fresh `bytes.Reader` each iteration, so it shares no AST either.
- **`BenchmarkParse`** constructs a new `lisp.Reader` per iteration inside the timed loop, which is intended — the reader's construction is part of the parse cost being measured — and it builds no runtime at all.
- **The lisp-level benchmark body inside `Runner.RunBenchmark`** reuses one tree across the ELPS-level iterations of a single call. That is in-runtime reuse under one environment, which is what a Go benchmark does with its own fixtures too, and is a different question from handing one tree to a succession of *different* runtimes. Not changed.

**Found and deliberately not fixed here**, filed as #434: `Runner.RunBenchmark` charges `env.Load` — parsing and evaluating the entire benchmark file — to the timed region if and only if a `SetupFn` is configured, because the `SetupFn` branch ends in `b.StartTimer()` and the path without it does not. No in-tree Runner sets `SetupFn`, so no number produced in this repository is affected, but `elpstest` is an exported package and `SetupFn` is an exported field, so a downstream Runner that sets one gets the inflated measurement. Found by reading, not by running; the issue says so and asks for the red test first.

## Gates

| gate | result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `go test -count=1 ./...` | clean |
| `go test -race ./...` | clean on retry — see the flake note below |
| `elps fmt -l ./...` | clean (binary deleted before commit) |
| `make go-test` | clean |
| `make race` | clean |
| `make test-elpscheck` | clean |
| `make test-examples` | clean |
| `./scripts/ci-gates-test.sh` | 210 passed, 0 failed |
| `./scripts/confidentiality-guard.sh` | clean (after `git add`) |
| `./scripts/bench-arms-check.sh` | arms comparable, 105 paired, n=6 per arm |
| `./scripts/benchstat-gate.sh` | 0 rows at or above gate |
| `./scripts/fuzz-budget-check.sh` | the sweep fits |

**Flake note, so the table is not read as cleaner than the runs were.** `lisp`'s `TestEvalTerminatingSeedsComplete/tail-recursion-bounded` failed once during the `-race` sweep with `context deadline exceeded`, and passed when `./lisp/` was re-run under `-race` immediately afterwards. It is unrelated to this change — the change is confined to `elpstest` — and it reproduces on an unmodified `ce9798d` worktree, 2 failures in 6 concurrent runs, stopping at a different step count every time (49,792 / 61,723 / 73,618 / 89,062 / 108,562 for the same fixed seed). Filed as #435.

## On fuzzing

No fuzz target was added. `RunBenchmark` is a test harness rather than a surface an ELPS program reaches — the fuzz targets exist to drive arbitrary program text through the evaluator, and the property fixed here ("two iterations must not share nodes") is an identity question about the harness, not a question about any input. The property is pinned by the catch above, and by the guard that fails loudly if the premise it relies on disappears.

---

### DIFF
--- a/elpstest/lisptest.go
+++ b/elpstest/lisptest.go
@@ -383,6 +383,22 @@ func RunTestSuite(t *testing.T, tests TestSuite) {
 
 // RunBenchmark runs a standard benchmark that executes expressions parsed from
 // source.
+//
+// Each iteration gets a fresh environment and a fresh copy of the parsed
+// expressions.  The copy matters: an ELPS expression's quoted literals are AST
+// nodes, and evaluation can write to them -- `stable-sort` sorts a literal in
+// place, which lisp.TestSliceViewSharesElements pins as current behaviour.
+// Without the copy, one parsed tree was shared by every iteration, so a
+// benchmark over such a source measured its declared workload once and its own
+// output thereafter, and the reported number moved with b.N.  This is the same
+// hazard lisp.TextLoader copies to avoid.  See issue #365.
+//
+// Both the environment and the copy are built with the timer stopped, so
+// neither is charged to ns/op, B/op or allocs/op.  Note that the untimed
+// per-iteration setup still costs wall clock, and the framework sizes b.N from
+// the timed duration alone: a source whose timed region is microseconds will
+// run for many multiples of -benchtime in real time.  Sources here are
+// expected to do enough work per iteration for that ratio to stay sane.
 func RunBenchmark(b *testing.B, source string) {
 	b.StopTimer()
 	p := parser.NewReader()
@@ -401,8 +417,12 @@ func RunBenchmark(b *testing.B, source string) {
 		if err != nil {
 			b.Fatal(err)
 		}
-		b.StartTimer()
+		iterExprs := make([]*lisp.LVal, len(exprs))
 		for i, expr := range exprs {
+			iterExprs[i] = expr.Copy()
+		}
+		b.StartTimer()
+		for i, expr := range iterExprs {
 			lerr := env.Eval(expr)
 			if lerr.Type == lisp.LError {
 				b.Fatalf("expr %d: %v", i, lerr)
--- a/elpstest/runbenchmark_test.go
+++ b/elpstest/runbenchmark_test.go
@@ -1,0 +1,158 @@
+// Copyright © 2026 The ELPS authors
+
+package elpstest
+
+import (
+	"flag"
+	"io"
+	"strings"
+	"testing"
+
+	"github.com/luthersystems/elps/lisp"
+	"github.com/luthersystems/elps/parser"
+)
+
+// This file covers issue #365: RunBenchmark parsed its source once, outside
+// the loop, and then evaluated that one tree in every b.N iteration. Each
+// iteration got a fresh Runtime but the same AST nodes, so iteration k saw
+// whatever iterations 1..k-1 left behind in them.
+//
+// That is not hypothetical in ELPS. A quoted literal in a source expression is
+// an AST node, and `stable-sort` is documented to sort in place and is pinned
+// as doing so through a literal by lisp.TestSliceViewSharesElements. So a
+// benchmark whose source sorts a literal sorts unsorted input on iteration 1
+// and already-sorted input on every iteration after it -- the harness quietly
+// substitutes a different workload for the one the benchmark declares, and the
+// reported number depends on b.N.
+//
+// The same reasoning is why lisp.TextLoader copies: it calls expr.Copy() on
+// every load precisely so two loads of one file cannot share nodes.
+const staleASTSource = `
+  ;; A quoted literal inside a function body is part of the parsed tree.
+  (defun literal () '(3 1 2))
+
+  ;; Detector: the literal must be pristine when an iteration begins. On a
+  ;; shared tree it is not, because the sort below already reordered it.
+  (if (equal? (literal) '(3 1 2))
+      ()
+      (error 'stale-ast "iteration began with an AST a previous iteration mutated"))
+
+  ;; Ordinary ELPS that mutates the tree: stable-sort sorts in place.
+  (stable-sort < (literal))
+`
+
+// staleASTIterations bounds the benchmark this test drives. Two iterations
+// would be enough to observe the sharing, but a benchmark measured by time
+// would run for -benchtime of *timed* work while paying the untimed
+// per-iteration environment build on top, which for a source this small is
+// several seconds of wall clock for a property that shows up on iteration 2.
+const staleASTIterations = "20x"
+
+// TestRunBenchmarkGivesEachIterationAFreshAST is the catch for #365. On main
+// the second iteration evaluates the literal the first one sorted, the
+// detector in staleASTSource raises stale-ast, RunBenchmark calls b.Fatalf,
+// and testing.Benchmark reports a zero result.
+func TestRunBenchmarkGivesEachIterationAFreshAST(t *testing.T) {
+	if f := flag.Lookup("test.benchtime"); f != nil {
+		prev := f.Value.String()
+		if err := f.Value.Set(staleASTIterations); err != nil {
+			t.Fatalf("set benchtime: %v", err)
+		}
+		defer func() {
+			if err := f.Value.Set(prev); err != nil {
+				t.Errorf("restore benchtime: %v", err)
+			}
+		}()
+	}
+	res := testing.Benchmark(func(b *testing.B) {
+		RunBenchmark(b, staleASTSource)
+	})
+	if res.N == 0 {
+		// testing.Benchmark discards the failed benchmark's output, so
+		// reproduce the failure here to say what actually went wrong.
+		t.Fatalf("RunBenchmark failed: an iteration evaluated an AST that an earlier iteration had already mutated.\n"+
+			"testing.Benchmark reports N=0 when the benchmark function calls Fatal and discards its output, so the\n"+
+			"same source evaluated twice against one parsed tree was re-run here, and the second pass reported:\n\n  %s",
+			sharedTreeSecondPass(t, staleASTSource))
+	}
+	if res.N < 2 {
+		t.Fatalf("benchmark ran %d iteration(s); this test cannot observe cross-iteration sharing with fewer than 2", res.N)
+	}
+}
+
+// sharedTreeSecondPass parses source once and evaluates that one tree in two
+// successive environments, exactly as the unfixed RunBenchmark did, and
+// returns what the second pass reported. It exists only to make the failure
+// message above say something.
+func sharedTreeSecondPass(tb testing.TB, source string) string {
+	tb.Helper()
+	p := parser.NewReader()
+	exprs, err := p.Read("benchmark", strings.NewReader(source))
+	if err != nil {
+		return "parse error: " + err.Error()
+	}
+	var last string
+	for pass := range 2 {
+		env := lisp.NewEnv(nil)
+		if err := lisp.GoError(lisp.InitializeUserEnv(env,
+			lisp.WithReader(p),
+			lisp.WithStderr(io.Discard),
+		)); err != nil {
+			return "environment: " + err.Error()
+		}
+		last = "no error"
+		for _, expr := range exprs {
+			if lerr := env.Eval(expr); lerr.Type == lisp.LError {
+				last = lerr.String()
+				break
+			}
+		}
+		if pass == 1 {
+			return last
+		}
+	}
+	return last
+}
+
+// TestQuotedLiteralIsMutatedByEval is a GUARD, not a catch: it passes on main.
+// It pins the premise the test above depends on -- that evaluating a tree can
+// leave state in the tree -- without going through RunBenchmark. If ELPS ever
+// copies quoted literals on evaluation, this guard fails first and says so,
+// rather than leaving the detector above passing for a reason it did not
+// intend.
+func TestQuotedLiteralIsMutatedByEval(t *testing.T) {
+	// concat copies, so `before` records the literal's state on entry rather
+	// than aliasing the node the sort is about to reorder.
+	const src = `(defun literal () '(3 1 2))
+	             (set 'before (concat 'list (literal)))
+	             (stable-sort < (literal))
+	             before`
+	p := parser.NewReader()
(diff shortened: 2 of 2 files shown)


sam-at-luther:
## CI's benchmark comparison, read

The Benchmark Comparison check has run (n=10 per arm, interleaved, one runner, arm64) and is the clean measurement the sandbox run could not be. It agrees with the conclusion in the body and is worth reading out rather than leaving as a green tick.

```
benchstat-gate: interpreted 23 delta row(s) + 314 no-change row(s); 3 significant move(s)
in the bad direction; 0 at or above the gate (timing 15%, allocation 5%).
```

Allocation geomeans, the metrics the gate treats as deterministic:

| package | `B/op` | `allocs/op` |
|---|---|---|
| `lisp` | +0.00% | +0.00% |
| `lisp/lisplib/libjson` | +0.04% | -0.00% |
| `parser/rdparser` | -0.00% | +0.00% |

Nearly every allocation row is annotated "all samples are equal" on both arms — including all 43 of the benchmarks that go through `RunBenchmark`. A per-iteration deep copy of the parsed source is real allocation, and none of it appears, which is the direct confirmation that the copy sits outside the timed region.

### The one row worth naming: `MacroDefun`

Three rows moved significantly in the bad direction, all below gate: `lisp MacroDefun-2` `sec/op` **+3.52%** (p=0.004, gate 15%), `lisp String/string-2` `sec/op` +0.42%, and `lisp MacroDefun-2` `B/op` +0.01% (gate 5%).

`String/string-2` is not a `RunBenchmark` benchmark at all — it calls `LVal.String()` directly and cannot be affected by this change — so a +0.42% "significant" move on it is a fair estimate of what this runner's noise floor produces at n=10.

`MacroDefun` is different, and I would rather flag it than let it pass: it *is* one of the 43, it was also the single significant row in the independent sandbox run (`B/op` +0.02% there), and at 35µs it is by far the smallest of the `RunBenchmark` set. The plausible mechanism is not that the copy is being timed — `B/op` moving +0.01% rather than the several-percent a charged copy would cost rules that out — but GC: one extra tree allocated and dropped per iteration raises heap pressure, and any collection that lands inside the timed region is charged to it. That effect is largest where the timed region is smallest, which is exactly `MacroDefun`.

+3.5% against a 15% gate on a benchmark whose base spread is ±3% is not something to act on, and the fix is not worth weakening to avoid it. Recording it so that if `MacroDefun` drifts again after this lands, the next person knows this change is a candidate and does not have to rediscover it.


---

PR #462, opened 2026-08-15
lsp: measure a semantic token from the source span, and give the reader's ' no token (#449)

### LINKED ISSUE #449: collectSemanticTokens mis-spans quoted symbols and escaped strings: 'a highlights only the quote, "x\ty" is one char short
Found by **running**, during the neighbourhood audit for #428 (PR #448). Two more places in `lsp/semantic_tokens.go` take a token's length from a *name* or a *decoded value* rather than from the node's span. Neither is out of bounds, so neither is what #428 is about and neither is caught by the bounds assertion #448 added to `FuzzLSPSession` — they are wrong-but-in-range highlighting.

Reproduce by opening the document and decoding `textDocument/semanticTokens/full`.

## 1. A quoted symbol's token is shifted one character left

`rdparser.applyPrefixLocation` moves a quoted node's `Col` back onto the `'`, so the symbol's location starts at the quote. The `LSymbol` case still takes `length` from `len(v.Str)`, which does not include it. The token therefore starts one character early and ends one character early.

| document | token | covers | should cover |
|---|---|---|---|
| `'foo` | `[0:0,+3)` | `'fo` | `foo` (or `'foo`) |
| `(error 'type-error 1)` | `[0:7,+10)` | `'type-erro` | `type-error` |
| `(list 'a 'b)` | `[0:6,+1)`, `[0:9,+1)` | `'`, `'` | `a`, `b` |

The last row is the visible one: for a one-character quoted symbol the token covers **only the quote**, so `a` and `b` go unhighlighted and the `'` is painted as a variable instead. Quoted one-character symbols are not exotic.

Numbers do not have this problem — `'42` goes through `tokenLength(v, content)` and comes out `[0:0,+3)`, covering `'42`.

## 2. A string's token measures the decoded value, not the source

```go
case lisp.LString:
	// String length includes quotes.
	length := len(v.Str) + 2
```

`v.Str` is the string after escape processing, so every escape sequence in the source is counted as the byte it decodes to.

| document | source width | token | shortfall |
|---|---|---|---|
| `(f "x\ty")` | 6 | `[0:3,+5)` | 1 |
| `"a\tb\tc"` | 9 | `[0:0,+7)` | 2 |
| `"é"` | 8 | `[0:0,+4)` | 4 |

Always an undershoot (a source escape is never shorter than what it decodes to), so the tail of the literal is left unhighlighted rather than the token running past the closing quote. The multi-line branch just below it already reaches for `content` and is unaffected; an unterminated `"abc` produces no token at all, so there is no overrun there either.

## Why not fixed in #448

#448 is scoped to the out-of-bounds defect and deliberately does not widen. Both of these also carry a design question that a length change alone does not answer:

- For (1), should `'` be **inside** the symbol's token (use the span, giving `'foo` one variable-coloured token), or should the symbol's token start after it and the `'` get its own `operator` token, which is what most Lisp editor modes do? Using the span is the smaller change; the second is probably the better rendering.
- For (2), the honest length is `EndCol - Col` from the node's span — the same `tokenLength(v, content)` the number case uses. Worth confirming the string node's span is reliable across the fault-tolerant parser before relying on it.

## Severity

Cosmetic, and lower than #428 was: nothing is out of range, so no client drops the response. The user sees a symbol or the tail of a string in the wrong colour, and cannot connect it to the quote or the escape.

sam-at-luther:
Fixes #449.

## The design decision for (a): the `'` is not inside the symbol's token

#449 leaves this open — the token could cover `'a` (2 characters, from the quote) or `a` (1 character, starting after it). **This PR gives the token to the atom and gives the `'` no token at all.**

That is the call PR #448 made for the other two reader prefixes, and its stated reason applies here unchanged. A semantic token is a claim about the meaning of the text it covers, and the only claim available for a quote is the one `classifySymbol` derives from *the symbol beside it*. Widen the token to `'a` and the quote gets painted with the symbol's classification: variable-coloured in `'a`, function-coloured in `'car`, keyword-coloured in `'if`. None of those says anything true about a quote. #448 rejected exactly that reasoning for `#^` and `#'` — a two-character token whose colour is "an accident of the standard library" — and suppressed them instead, leaving the punctuation to the client's syntax grammar.

It is also what makes #448's own justification true. `isSynthesizedPrefixHead` says the suppression "makes the three reader prefixes agree, since `'` already produces no token of its own". That was true only in the narrow sense that no *node* sits at the quote. In the emitted tokens the quote was very much painted: `'42` covered `'42` as a number, and `'a` covered **only** the `'`, as a variable. After this change the citation is accurate — none of `'`, `#'`, `#^` is inside any semantic token, and the atom after the prefix gets exactly its own.

Both halves of #449 still reproduced on `5ef6106`, i.e. after #448, #442 and #456 landed. Captured by running:

| document | on `5ef6106` | covers | now |
|---|---|---|---|
| `'a` | `[0:0,+1)` | `'` | `[0:1,+1)` → `a` |
| `'foo` | `[0:0,+3)` | `'fo` | `[0:1,+3)` → `foo` |
| `(error 'type-error 1)` | `[0:7,+10)` | `'type-erro` | `[0:8,+10)` → `type-error` |
| `(f "x\ty")` | `[0:3,+5)` | `"x\ty` | `[0:3,+6)` → `"x\ty"` |
| `"\U0001F600"` | `[0:0,+6)` | `"\U000` | `[0:0,+12)` → the whole literal |
| `"""raw"""` | `[0:0,+5)` | `"""ra` | `[0:0,+9)` → the whole literal |

## (b), and a third symptom the issue did not have

Length now comes from `EndPos - Pos` — the span the scanner measured — instead of `len(v.Str)+2`, which measured the string *after* escape processing.

Found by running the new property while fixing (b), and **not** in #449: the multi-line branch tested the *decoded* value for a newline, so a single-line literal containing the **escape** `\n` took it and was handed "the rest of the line".

```
(f "a\nb" c)   ->  [0:3,+9)  covering  "a\nb" c)
```

That runs past its own closing quote and over everything after it. #449 states "Always an undershoot (a source escape is never shorter than what it decodes to), so the tail of the literal is left unhighlighted rather than the token running past the closing quote" — that is not so; this one overshoots by 3. Whether a literal is multi-line is now decided by the span (`EndLine > Line`) rather than by looking for a newline in the value.

## `EndCol` is not used, deliberately — and neither is UTF-16

The obvious way to write this fix is `EndCol - Col`, the way the number case already did. That is wrong. `token.TokenEnd` derives `EndCol` by ranging over `tok.Text`, which counts **runes**, and adds them to a `Col` that `Scanner.LocStart` computes as `startPos - startLinePos + 1`, a **byte** offset. `EndCol - Col` is therefore a rune width laid on a byte base. For the 4-byte symbol `déf` it is 3. `EndPos` and `Pos` are both byte offsets and their difference is exact, so the fix uses those. `TestSemanticTokensNonASCIISymbolSpans` is the guard that says so.

On **position encoding**: LSP counts a position in UTF-16 code units by default, and this server does not comply — every column it emits, from every handler, is a byte offset, and `elpsToLSPPosition` passes `Col` through untouched. It cannot negotiate its way out either: the server is built on `glsp/protocol_3_16`, whose `ServerCapabilities` has no `positionEncoding` field at all (that arrived in LSP 3.17), so UTF-16 is not optional for it.

This PR does **not** change that, on purpose. A length is added to a start; emitting a UTF-16 length against a byte start would produce a token that is wrong in a new and less predictable way. What the fix guarantees is that **a length is in the same unit as the column it is added to** — bytes, throughout, consistent with `position.go`, the bounds assertion #448 added, and every other handler. The encoding gap is server-wide, is not fixable by a length, and is filed separately.

## Tests: red-then-green

`lsp/semantic_tokens_span_test.go` is new. Everything in it was run against `5ef6106` and the real output is recorded in the file; anything that passed there is labelled `GUARD` in place.

- **Red on `5ef6106`**: 6 test functions, 30 failing subtests. Over the committed seed corpus, **22 of 118 seeds, 433 offending tokens** — including ELPS source this repository ships (`(in-package 'foo)`, `(error 'type-error ...)`, `(set 'method-table ...)`). Now **0 of 118**.
- **Guards** (passed on `5ef6106`, must keep passing): `'#^0` and the `#^`/`#'` suppression from #448; `"x"`, `"é"`, `"😀"` (byte-measured, and `len(v.Str)+2` already got them right, so a fix in some other unit would have broken them); a genuinely multi-line raw literal; an unterminated `"abc`; `déf` and friends.
- `TestSemanticTokensPrefixesEmitNoToken` replaces what #448's `TestSemanticTokensQuotePrefixEmitsNoToken` meant to say. That test asserted only that no token was exactly `[0:0,+1)` — which `'foo` passed while covering `'fo` — so it could not have caught (a).
- `TestSemanticTokensQuotedDefinitionKeepsItsModifier` pins a coupling that is otherwise **silent**. A quoted atom now has two positions, and `classifySymbol` must keep using the node's (the quote) because that is what `buildSymbolDefsMap` keys on — for `(set 'x 1)` the analysis `Source` is column 6, the quote, not column 7. Verified by running: substitute the atom's column and the entire `lsp` package stays green, because the lookup miss falls through to `semTokenVariable`, which is also what the found symbol maps to. Only the `definition` modifier disappears, and nothing asserted modifiers.

## The fuzz assertion is now coverage, not bounds

#448 put `checkTokenBounds` into `FuzzLSPSession`. A bounds check cannot catch either defect here: both are strictly in bounds. Bounds asks that a token name *some* text; it does not ask that it name the *right* text.

`checkTokenSpans` now also asks `coverageDefect` — the same predicate the table tests use, one definition, so the fuzzer and they cannot drift. It is still checkable from the response alone and still says nothing about which answer is right: a token may be a variable or a function, but it may not start on reader punctuation, span two atoms, or stop in the middle of a string literal.

What it deliberately does **not** assert is that a token's neighbours are delimiters. That stronger rule is wrong against this fault-tolerant parser, which splits malformed input into genuinely adjacent atoms: `1.2.3` reads as the float `1.2` then the symbol `.3`, and `0x` as `0` then `x`. Both are in `fuzzseed.All()`, both tokens name exactly the text they were computed from, and flagging them would be a false positive on input committed here.

**Seeds.** The quoted-atom half needed none — the existing `uses` document is `(in-package 'user)\n(use-package 'demo)\n…` and goes red on `5ef6106` at `'use`/`'dem`. The escaped-string half is reached by nothing in the corpus (no seed here or in `fuzzseed` contains a string literal with an escape, and a literal without one has `len(v.Str)+2` exactly right), so it is seeded rather than left to a mutator that would have to produce a balanced pair of quotes *and* a backslash escape inside them.

Verified red on `5ef6106` (`seed#0`, `seed#18`, `seed#25`, …) and green here.

**Exec counts:** `-fuzztime=300s` → **159,793 execs**, 381 new interesting inputs, 0 failures. A longer run continuing from that corpus is reported in a comment below. `scripts/fuzz-budget-check.sh` unchanged at 29 targets / 8 shards / 10 min headroom (no target added).

## Neighbourhood audit

Every place in `lsp/` that derives a length from a decoded or name string rather than from the node's span:

- The `elpsToLSPRange(loc, nameLen)` callers — `call_hierarchy.go` ×4, `code_actions.go`, `definition.go` ×2, `highlight.go` ×2, `hover.go` ×2, `linked_editing.go` ×2, `references.go` ×3, `rename.go` ×3 — are **span-first**: `elpsToLSPRange` uses `EndLine`/`EndCol` whenever present and takes `nameLen` only as a fallback. Cleared by #448's audit; not redone here.
- `position.go` `locContainsCol` (`end := start + len(name)`, overridden by `loc.EndCol` when set) and `selection_range.go` `locContainsPosition` (`endCol = startCol + 1` only when `EndLine == 0`) are the same shape — span-first, name as fallback.
- `semantic_tokens.go` `isSynthesizedPrefixHead` is span-based already (#448).

So the two in `collectSemanticTokens` were the only remaining instances of the #449 class, and both are fixed.

The audit did surface a **different** defect in the span-first path itself, which is why this fix avoids `EndCol`: every one of those callers consumes `EndCol` directly, and `EndCol` is rune-counted onto a byte `Col`. Filed separately rather than widened into this PR.

## Gates

`go build`, `go vet`, `go test -count=1`, `go test -race`, `golangci-lint run` (0 issues), `elps fmt -l`, `elps doc -m`, `make go-test`, `make race`, `make test-elpscheck`, `make test-examples`, `scripts/ci-gates-test.sh` (233 passed / 0 failed), `scripts/confidentiality-guard.sh` (clean), `scripts/fuzz-budget-check.sh` (fits, 10 min headroom) — all clean.

Benchmark gate: the diff touches `lsp/` only, and no benchmark exists in `lsp/` — the benchmarked packages (`parser/rdparser`, `mcpserver`, `lisp`, `lisp/lisplib/libjson`) are untouched, so no timing or allocation row is reachable from this change.

---

### DIFF
--- a/lsp/lsp_fuzz_test.go
+++ b/lsp/lsp_fuzz_test.go
@@ -770,11 +770,11 @@ func (s *session) opSemanticTokens(a opArgs) {
 	res, err := s.srv.textDocumentSemanticTokensFull(s.ctx, &protocol.SemanticTokensParams{TextDocument: a.ident()})
 	s.record("semanticTokens", res, err)
 	if err == nil && res != nil {
-		s.checkTokenBounds(a.uri, res.Data)
+		s.checkTokenSpans(a.uri, res.Data)
 	}
 }
 
-// checkTokenBounds is the one CONTENT assertion this target makes, and the
+// checkTokenSpans is the one CONTENT assertion this target makes, and the
 // exception to "NOT asserted: that any request returns a particular result".
 //
 // It is admissible here because it does not say what the answer should be. It
@@ -797,7 +797,23 @@ func (s *session) opSemanticTokens(a opArgs) {
 // two-column span, so "#'foo" told the editor to decorate 13 characters of a
 // 5-character line. This assertion was left out of the target when that defect
 // was filed, because with it in place the target was red on its own seeds.
-func (s *session) checkTokenBounds(uri string, data []protocol.UInteger) {
+//
+// COVERAGE, ADDED FOR elps#449.  Bounds alone was too weak to be the whole
+// property.  It asks that a token name SOME text in the document; it does not
+// ask that it name the RIGHT text, and both defects in elps#449 were strictly
+// in bounds -- a quoted symbol's token slid one character left onto the ' and
+// stopped one character short of the symbol, and a string's length was measured
+// on the DECODED value so `"x\ty"` got five characters of a six-character
+// literal.  Every such token satisfies the bound and misinforms the editor.
+//
+// coverageDefect (semantic_tokens_span_test.go) is the stronger question, asked
+// of the same decoded triples: does the text a token claims look like the atom
+// it is classifying?  It is still checkable from the response alone and still
+// says nothing about which answer is right -- a token may be a variable or a
+// function, but it may not start on reader punctuation, span two atoms, or stop
+// in the middle of a string literal.  One definition of the property, shared
+// with the table tests next door, so the fuzzer and they cannot drift.
+func (s *session) checkTokenSpans(uri string, data []protocol.UInteger) {
 	if s.tokenErr != nil {
 		return
 	}
@@ -829,9 +845,21 @@ func (s *session) checkTokenBounds(uri string, data []protocol.UInteger) {
 		}
 		// A CRLF document keeps its "\r" in the line; the editor does too, so
 		// it is part of the line for this purpose.
-		if char+length > len(lines[line]) {
+		if char < 0 || length < 0 || char+length > len(lines[line]) {
 			s.tokenErr = fmt.Errorf("semantic token [%d:%d,+%d) overruns line %d, which is %d bytes: %q (%s)",
 				line, char, length, line, len(lines[line]), lines[line], uri)
+			return
+		}
+		cov := tokenCoverage{
+			tok: rawToken{
+				line: line, startChar: char, length: length,
+				tokenType: int(data[i+3]),
+			},
+			covered: lines[line][char : char+length],
+		}
+		if msg := coverageDefect(cov, lines[line]); msg != "" {
+			s.tokenErr = fmt.Errorf("semantic token [%d:%d,+%d) covers %q: %s (%s)",
+				line, char, length, cov.covered, msg, uri)
 			return
 		}
 	}
@@ -1196,6 +1224,25 @@ func FuzzLSPSession(f *testing.F) {
 	add("'#^0", "", allOpsScript())
 	add("(list #'car\n      #'cdr)", "", allOpsScript())
 	add("(lisp:function foo)", "", allOpsScript()) // longhand: not synthesized
+
+	// The shapes the COVERAGE half of checkTokenSpans exists for (elps#449).
+	//
+	// The quoted-atom half is already reached without these: `uses` above is
+	// "(in-package 'user)\n(use-package 'demo)\n...", and on 5ef6106 each of
+	// those quoted symbols covers "'use" and "'dem" respectively. The ESCAPED
+	// STRING half is reached by nothing in the corpus -- no seed here or in
+	// fuzzseed contains a string literal with an escape in it, and a literal
+	// without one has len(v.Str)+2 exactly right -- so the property would have
+	// held by the corpus never asking. Seeded rather than left to the mutator,
+	// which has to produce a balanced pair of quotes AND a backslash escape
+	// inside it before the length is ever wrong.
+	add(`(f "x\ty")`, "", allOpsScript())
+	add(`(f "a\nb" c)`, "", allOpsScript()) // escaped \n, not a newline
+	add(`(f """raw""" c)`, "", allOpsScript())
+	add("(f \"\"\"a\nb\"\"\" c)", "", allOpsScript()) // genuinely multi-line
+	add("(list 'a 'b)", "", allOpsScript())           // one-character quoted symbols
+	add("' a", "", allOpsScript())                    // whitespace after the prefix
+	add("'\na", "", allOpsScript())                   // and a newline after it
 
 	// Document B as the WORKSPACE FILE that defines a macro and document A as
 	// the file that calls it. This is the only shape that reaches
--- a/lsp/semantic_tokens.go
+++ b/lsp/semantic_tokens.go
@@ -88,8 +88,9 @@ func (s *Server) textDocumentSemanticTokensFull(_ *glsp.Context, params *protoco
 	symbolRefs := buildSymbolRefsMap(analysisResult)
 
 	var tokens []rawToken
+	src := &sourceText{content: content}
 	for _, expr := range ast {
-		collectSemanticTokens(expr, symbolDefs, symbolRefs, content, &tokens)
+		collectSemanticTokens(expr, symbolDefs, symbolRefs, src, &tokens)
 	}
 
 	// Sort by position (line, then character).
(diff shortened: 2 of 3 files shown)


sam-at-luther:
Fuzz exec counts for the coverage property added to `FuzzLSPSession`, run locally on this branch:

| run | wall clock | execs | new interesting | result |
|---|---|---|---|---|
| first | `-fuzztime=300s` | 159,793 | 381 (total 479) | PASS, 0 failures |
| second, continuing from that corpus | `-fuzztime=600s` | 316,646 | 559 (total 1,138) | PASS, 0 failures |

476,439 execs in 15 minutes with no coverage violation. The second run is the interesting one: it starts from a corpus the property has already shaped, so its 559 new interesting inputs are inputs found *under* the stronger assertion rather than under bounds alone.

Confirmed red before the fix, with the same seeds — on `5ef6106`'s `semantic_tokens.go` the target fails at `seed#0`, `seed#18` and `seed#25`:

```
semantic token [0:12,+4) covers "'dem": starts on a ' , which is reader punctuation and not part of the atom
semantic token [0:3,+5) covers "\"x\\ty": the literal is not closed and the token does not run to the end of the line
```

`seed#0` and `seed#18` come from the corpus that was already there (`uses` = `(in-package 'user)\n(use-package 'demo)\n…`); `seed#25` is one of the escaped-string seeds this PR adds. So the quoted-atom half of the property is reached by pre-existing seeds and the escaped-string half needed new ones — unlike #448, which had to seed both halves.

`scripts/fuzz-budget-check.sh` is unaffected: 29 targets, 8 shards, largest shard 4 targets, 50 min required against a 60 min timeout, **10 min headroom**. No target was added, only an assertion inside an existing one.


---

Issue #466, opened 2026-08-15
LVal.Copy shallow-copies Meta and MacroExpansion too, so a "deep copy" shares its formatting metadata and its per-node expansion ID


sam-at-luther:
Found during the neighbourhood audit for #446. The `Meta` half was observed by **running** (probe output below); the `MacroExpansion` half is the same line of code and was found by **reading**. **Reproduce-or-close** applies to the consequences, not to the sharing.

## What it is

`lisp/lisp.go`, `LVal.Copy`:

```go
cp := &LVal{}
*cp = *v // shallow copy of all fields including Map and Bytes
```

#446 separated `Source`. The same assignment carries the other two mutable metadata pointers across unchanged, so a "deep copy" and the value it copied share them at every depth:

- **`Meta *SourceMeta`** — the format-preserving metadata. One `SourceMeta` for both trees, including its `[]*token.Token` slices (`LeadingComments`, `InnerTrailingComments`) and the `*token.Token` in `TrailingComment`, each of which carries its own `*token.Location`. So for a format-preserving tree the position objects #446 separated on the node are still shared one level down, through `Meta`.
- **`MacroExpansion *MacroExpansionInfo`** — whose `ID` is documented "unique per node, monotonically increasing" and is read by `lisp/x/debugger/engine.go` (`loc.MacroID = expr.MacroExpansion.ID`) to tell expansion nodes apart when stepping. A copy duplicates the ID, so two distinct nodes carry one "unique" value. The embedded `*MacroExpansionContext` — and therefore `CallSite`, which #456 made the expansion's own object — is shared too.

## Probe

Format-preserving parse of `"; lead\n(+ 1 2) ; trail\n"`, against `5ef6106`:

```
root Meta: orig=0xc0001abb20 cp=0xc0001abb20 same=true
after cp.Meta.BlankLinesBefore=99: orig.Meta.BlankLinesBefore=99 (was 0)
LeadingComments[0] token: orig=0xc0001ac540 cp=0xc0001ac540 same=true
```

A write through the copy moves the original, same as #446's probe.

## Is anything broken today

Not established, and I am not claiming it is. The writers I found:

[...]

PR #467, opened 2026-08-15
Give LVal.Copy's result its own positions (#446)

### LINKED ISSUE #446: LVal.Copy shallow-copies Source, so a "copy" shares its *token.Location with the original — including every TextLoader evaluation

sam-at-luther:
Fixes #446

`LVal.Copy` is documented as "creates a deep copy of the receiver" and is one for `Cells` — `copyCells` allocates a fresh `*LVal` per node. `Source` was not: `*cp = *v` carried the `*token.Location` across, so a "copy" and the tree it came from held **one mutable Location object at every depth**, and a write through either moved the other.

## It still reproduces on `main`

Checked first, on `5ef6106` (not on the commit the issue was filed against). Every catch below is red there:

```
--- FAIL: TestCopyDoesNotAliasSourceLocation
    Expected and actual point to the same object: 0xc0003ca730
      &token.Location{File:"copy-source.lisp", Pos:0, Line:1, Col:1, EndPos:7, EndLine:1, EndCol:8}
    a copy shares the original's *token.Location (#446)

    expected: 1
    actual  : 99
    a write through the copy moved the original's recorded position (#446)

--- FAIL: TestCopyDoesNotAliasSourceAtDepth
    Should be zero, but was 18
    18 of 18 copied nodes share the original's *token.Location (#446)

--- FAIL: TestParseTreeCopyIsFullyPrivate
    Should be zero, but was 43
    43 of 43 copied parse-tree nodes share the original's *token.Location (#446)
```

And the caller the issue says this matters for, `lisp.TextLoader` — the entry point an embedder is pointed at for a reusable parse cache, which `macro_stamp_shared_ast_test.go` contrasts with the `Load*` entry points because it copies:

```
--- FAIL: TestTextLoaderEvaluationsGetPrivatePositions
    Expected and actual point to the same object: 0xc0003cb6d0
      &token.Location{File:"cache.lisp", Pos:11, Line:1, Col:12, EndPos:15, EndLine:1, EndCol:16}
    the first evaluation reports its position through the retained cache's object (#446)
    ... the second evaluation reports its position through the retained cache's object (#446)
    ... two evaluations of the cached tree share one *token.Location (#446)

    expected: 1
    actual  : 99
    a write through one evaluation moved the position the cache will report forever (#446)
    a write through one evaluation moved a sibling evaluation's position (#446)
```

Both "private" trees handed out by one `TextLoader` reported their positions through the **same** object, and it was the retained cache's own.

## Live or latent: latent, checked rather than assumed

I re-ran the population argument on `5ef6106` rather than taking #446's or #431's word for it. Every in-place write through a `*token.Location` in non-test code, found by grepping for field assignments through a `Location` pointer:

| Site | Verdict |
|---|---|
| `parser/rdparser/parser.go` — 5 blocks (`inheritEndPos` 504-506, `applyPrefixLocation` 517-521, the two closing-bracket blocks 735-737 / 776-778, `tokenLVal` 1023-1025) | Writes into nodes the reader has just built, and since #426/#442 into `Location`s those nodes own |
| `lisp/env.go` error constructors | Take `env.Loc.Copy()` (#421) — a re-point, not a write-through |
| `lisp/x/debugger/engine.go` 976-978 (`loc.File = expr.Source.File`) | Writes into the debugger's **own** `StepLocation`; reads `Source` |
| `lsp/inlay_hints.go`, `lsp/code_actions.go`, `lsp/workspace_symbols.go` | `Source.Line == 0` comparisons, not writes |
| everything else that touches a position | Reads a `Location` and formats it into its own `Position`/`Range`/`Span` value type |

`stampMacroExpansion` **assigns** `v.Source = callSite` rather than writing through the pointer, so it re-points a copy without touching the original.

So no writer can reach the shared object today. **The writes the tests perform are writes the tests perform** — they demonstrate the mechanism, they do not record an observed corruption. What is a real observed defect is the **sharing**, which is what the assertions above are about. Said plainly in the test file header so the next reader is not misled.

## The fix, and why one Location per node is the right count here

```go
if v.Source != defaultSourceLocation {
	cp.Source = v.Source.Copy()
}
```

`(*token.Location).Copy()` already exists and is nil-preserving; no second helper.

**#456's framing does not transfer, and I checked before concluding that.** There, N stamped nodes genuinely sat at one position, so one object per macro *call* separated the two owners and the cost fell from per-node to +2.58% B/op. Here every node has a position **of its own** — the copy's node at line 1 col 12 cannot share an object with the node at line 1 col 4 — so N nodes need N objects. There is no cheaper object count. The only way to remove the allocation is to stop these objects being mutable, which is the `LVal.Source` immutability work tracked on #362 and sequenced after the v1.50.0 API break.

**Why not fix it only inside `TextLoader`.** `TextLoader` obtains its private tree *by calling* `expr.Copy()`, so the per-node cost would be identical, only narrower — and every other `Copy()` caller would keep a "deep copy" that is not one. Rejected on those grounds, not on measurement.

**The one Location deliberately left shared** is `nativeSource`'s process-wide singleton. A copy separates two **owners**, and that object has exactly one — the process. It is read-only by contract, `SingletonSnapshot` watches its bit pattern, and the parser never leaves it in an AST (`TestParserDoesNotAliasSharedNativeLocation`), so it cannot reach a parse cache in the first place. `TestParseTreeCopyIsFullyPrivate` asserts the consequence at this end: for a tree that came from the reader, "no node shares" has no exceptions at all.

## Measured

Two arms, interleaved, on the same box, `BENCH_COUNT=10`, `scripts/bench-run-arms.sh`, base = `5ef6106`.

**Caveat stated up front: this box was shared with other agents' full test runs (load average ~20 on 4 cores) for part of the window.** That makes the *timing* rows noise-dominated — `AliasStableSortSliceView` reports ±101% spread on the base arm — and I am not claiming a timing result in either direction. The allocation rows are exact counts and are unaffected by load; every one below is ±0% with p=0.000. CI's benchmark job re-measures on its own runner.

The gate's verdict:

```
below-gate  lisp     sec/op    PackageGetVal-4               delta=+8.11% p=0.029 (gate 15%)
below-gate  lisp     B/op      AliasStableSort-4             delta=+0.02% p=0.000 (gate 5%)
below-gate  lisp     allocs/op AliasStableSort-4             delta=+0.02% p=0.000 (gate 5%)
below-gate  libjson  B/op      Package/get-nested-baseline-4 delta=+4.36% p=0.000 (gate 5%)
below-gate  libjson  allocs/op Package/get-nested-baseline-4 delta=+3.93% p=0.000 (gate 5%)

benchstat-gate: interpreted 25 delta row(s) + 318 no-change row(s); 11 significant
move(s) in the bad direction; 0 at or above the gate (timing 15%, allocation 5%).
```

**Gate verdict: 0 rows at or above the gate.** No waiver added; the two existing waivers (#412/#410, `libjson/Encode`) are reported unused and left alone.

The whole cost of this change lands on **one row**, and it is explicable to the allocation: `Package/get-nested-baseline` goes 509.0k → 529.0k allocs/op, **+20,000**, and its body is 1000 iterations of 5 `assert-equal` forms. `opAssert` does `env.Eval(test.Copy())` — it copies its test expression out of the parse tree on every evaluation — so 5,000 asserts × ~4 nodes each is exactly the 20k. Every interpreter row is flat to ±0.02%, because the values the evaluator copies at runtime were built by Go and carry the native singleton, which is skipped.

`PackageGetVal`'s +8.11% sec/op is noise: the row reports ±30% spread, and its own `allocs/op` is `0.000 ± 0%` on both arms — a benchmark that allocates nothing cannot be moved by an allocation. Under #452 the move is far inside the row's own measured spread; it is below the 15% gate regardless. **Allocations are not exempt and I did not treat them as such** — the `+3.93%` row above is adjudicated against the 5% gate on its merits.

### What the exception buys, measured

To check that skipping the native singleton is doing real work rather than being asserted, I built a third arm that copies it too and measured that against this branch — interleaved, n=5, `-benchtime=10x` so the allocation counts stay exact under load:

```
                    │  this branch  │            copy-the-singleton         │
                    │   allocs/op   │   allocs/op    vs base                │
AliasCDR-4            9.047k ± ∞ ¹   10.046k ± ∞ ¹  +11.04% (p=0.008 n=5)
AliasStableSort-4     101.7k ± ∞ ¹    127.3k ± ∞ ¹  +25.12% (p=0.008 n=5)
EnvFunCallPos-4       22.06k ± ∞ ¹    24.06k ± ∞ ¹   +9.07% (p=0.008 n=5)
EnvGet-4              128.0k ± ∞ ¹    128.0k ± ∞ ¹        ~ (p=0.444 n=5)
geomean               14.71k          15.76k         +7.17%
```

Three rows past the 5% allocation gate and a +7.17% geomean. The exception is load-bearing, not a convenience.

## Red, then green

| Test | `5ef6106` | This branch | Kind |
|---|---|---|---|
| `TestCopyDoesNotAliasSourceLocation` | FAIL (shares; write moves original) | PASS | catch |
| `TestCopyDoesNotAliasSourceAtDepth` | FAIL (18 of 18 nodes) | PASS | catch |
| `TestTextLoaderEvaluationsGetPrivatePositions` | FAIL (both evaluations, the cache's own object) | PASS | catch |
| `TestParseTreeCopyIsFullyPrivate` | FAIL (43 of 43 nodes) | PASS | catch |
| `TestCopyDoesNotAliasSourceOfReferenceTypes` | FAIL (array and sorted-map roots) | PASS | catch |
| `TestSharedTreeSeedsAgree` (folded property) | FAIL on 14 of 14 seeds | PASS | catch |
| `FuzzParsedLocationInvariants` seed corpus (folded property) | FAIL on 20+ seeds | PASS | catch |
| `TestCopyPreservesSourcePosition` | PASS | PASS | **guard** — separating the objects must not change what they say |
| `TestCopyPreservesNilSource` | PASS | PASS | **guard** — nil means "no position recorded" and must stay nil |
| `TestCopyKeepsSharingTheNativeSingleton` | PASS | PASS | **guard** — pins the one exception, so a later "improvement" that copies it fails here rather than in a benchmark |

The three guards are labelled as guards in the file, not counted as catches.

## Fuzzing

Folded into two existing targets rather than adding a twenty-ninth. `scripts/fuzz-budget-check.sh` is unchanged at 29 targets / 8 shards / 10 min headroom.

**`FuzzSharedTreeEval`** (`lisp`) — the target already about who owns a parse tree. Before the shared/private evaluation, it now copies each expression and walks the pair, asserting no node of the copy holds the original's `Location`. Costs one `Copy` and one paired walk; no extra evaluation. Iterative, so the walk cannot be the thing that overflows the stack on a deeply nested input, and capped at 200k nodes with a truncated walk saying so rather than passing silently. The native singleton is allowed rather than assumed absent, so a reader change that starts emitting it fails at *its* guard (`TestParserDoesNotAliasSharedNativeLocation`) and not confusingly at this one.

**`FuzzParsedLocationInvariants`** (`parser/rdparser`) — this one already builds an `owner` map keyed by `*token.Location` for property 1 (#426: every node owns its position). Walking the **copy** through that same map states both halves of #446's property at once: no copied node may hold an original's object, and no two copied nodes may hold one either. That uniqueness half is stronger than the paired walk above, and it comes almost free from machinery that was already there.

60s smoke on each, clean:

```
FuzzSharedTreeEval           elapsed: 1m1s, execs: 263973 (2276/sec),  new interesting: 117 (total: 860)  PASS
FuzzParsedLocationInvariants elapsed: 1m1s, execs: 479352 (16008/sec), new interesting: 69 (total: 403)   PASS
```

Both seed corpora run on every `go test`, which is where the red rows in the table above come from.

## Neighbourhood audit

Only what the prior audits did not cover. **Not re-done:** #421 cleared `LEnv.Lambda`, `LEnv.TaggedValue`, `CallFrame.Source`, `CallStack.Copy()`, `env.Loc = v.Source` as a hot-path borrow, and the read-only consumers in `lint`/`analysis/perf`/`profiler`; #442 covered `Parser.Location` ownership and the five parser write sites; #419 covered the reader's synthetic locations; #456 covered `MacroExpansionContext.CallSite`/`DefSite`, `ParseNegative`, `locateSynthesized`, the closing-bracket guards, `Parser.errorf`, and the tree-wide sweep of position consumers. This PR's neighbourhood is `LVal.Copy` itself and the other shallow-copy constructors next to it.

| Site | Verdict |
|---|---|
| **`LVal.Copy`'s `Meta *SourceMeta`** | **Found by running. Filed as #466, not fixed here.** Shared by the copy — including the `[]*token.Token` slices, so for a format-preserving tree the `Location`s this PR separated on the node are still shared one level down. Probe: `cp.Meta.BlankLinesBefore = 99` moved the original's. |
| **`LVal.Copy`'s `MacroExpansion *MacroExpansionInfo`** | **Found by reading. Filed as #466 with the above.** Shared, including an `ID` documented "unique per node" and read by `debugger/engine.go` for step-into differentiation, so a copy duplicates a value whose contract is uniqueness. |
| `Quote`, `Splice`, `shallowUnquote`, `FunRef` | Four more `*cp = *v` sites in `lisp.go`. All deliberately **shallow** — they share `Cells` too, being "the same value with one flag changed" — so sharing `Source` is consistent with their contract rather than a broken promise of privacy. #446's argument does not reach them. Left, not filed. |
| `LVal.Copy`'s `Native` | `LArray` shares its `Cells` backing and `LNative`/`LBytes` share their payload, all by documented reference semantics; `LSortMap` goes through `copyMapData`. The node `Copy` was called on now gets a private `Source` in every case — `TestCopyDoesNotAliasSourceOfReferenceTypes` pins the array and sorted-map roots, and the nodes reached *through* them keep sharing, as they must. |
| `LEnv.Copy`'s `Loc` | Shallow-copied, and documented "not quite a deep copy". `Loc` is the hot-path borrow #421 recorded and deliberately left; `eval` re-**assigns** it per step rather than writing through it. Downstream of a reported finding — not re-filed. |
| `lisp/env.go:631,676` — `Source: env.Loc` | `LEnv.Lambda` / `LEnv.TaggedValue`. Exactly what #421 recorded ("a copy here is a new allocation on a path that has none, per lambda … Left, reported"). Still present, still reachable. **Not re-filed, not widened.** |
| `analysis/expander.go:173` (`cp := *e.lastPanic`) | Copies a panic value, not a `Location`. Clean. |
| `elpstest.RunBenchmark`'s `expr.Copy()` | Now separates positions too. It runs with the timer stopped, so this adds no measured cost — which is also why the benchmark suite does not exercise the parse-tree copy path directly, and why `libjson`'s `assert`-heavy row is the only one that moved. |

## Gates

All run on this branch at `8835db4`.

| Gate | Result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `go test -race ./...` | clean |
| `golangci-lint run ./...` | **0 issues** (3 `testifylint` findings on the new test were fixed, not suppressed) |
| `elps fmt -l ./...` | clean (binary built, run, deleted before committing) |
| `make go-test` | pass |
| `make race` | clean |
| `./scripts/ci-gates-test.sh` | 233 passed, 0 failed |
| `./scripts/fuzz-budget-check.sh` | fits — 29 targets, 8 shards, largest shard 4, required 50 min vs 60 min timeout, 10 min headroom |
| `scripts/benchstat-gate.sh` | **0 rows at or above the gate** |

## Found and filed

- **#466** — `LVal.Copy` shallow-copies `Meta` and `MacroExpansion` as well, so a "deep copy" shares its formatting metadata (and the `*token.Token`s inside it, each with its own `Location`) and its per-node macro-expansion `ID`. The `Meta` half found by **running**, with probe output on the issue; the `MacroExpansion` half by **reading**. Latent by the same population argument as #446. Filed with a reproduce-or-close note on the consequences, and deliberately not fixed here — unlike `Source`, both fields are nil on the hot path, so the cost argument is different and the open question is whether the separation is wanted at all or whether the doc comments are what is wrong.

Not new, and deliberately not re-filed: `LEnv.Lambda`'s `Source: env.Loc` (recorded by #421's audit, confirmed still present), and the four intentionally-shallow constructors above.

## On the surviving draft

This branch started from uncommitted work left by an earlier agent that died mid-task; it left no commits, no push and no report, so none of its reasoning was recoverable. **Kept**, after re-establishing each claim from evidence: the `LVal.Copy` fix and its native-singleton exception (the exception's cost claim re-measured independently — the draft's `+25.2%` / `+9.1%` came back as `+25.12%` / `+9.07%`), the eight regression tests, and the fold into `FuzzSharedTreeEval`. **Added:** the fold into `FuzzParsedLocationInvariants`, the `#431`-vs-`#446` framing argument, the re-run of the writer-population audit on current `main`, the whole-tree benchmark measurement, and the `Meta`/`MacroExpansion` audit finding. **Corrected:** commit references pointing at the pre-rebase base, an `intrange` violation, and three `testifylint` findings.

### DIFF
--- a/lisp/copy_source_test.go
+++ b/lisp/copy_source_test.go
@@ -1,0 +1,299 @@
+// Copyright © 2026 The ELPS authors
+
+// Regression tests for elps#446 -- LVal.Copy sharing its *token.Location.
+//
+// LVal.Copy is documented as "creates a deep copy of the receiver", and for
+// Cells it is one: copyCells allocates a fresh *LVal per node.  Source was
+// not.  `*cp = *v` carried the *token.Location across, so the copy and the
+// original pointed at ONE mutable Location object, at every depth, and a
+// write through either moved the other.
+//
+// The caller that makes this matter is lisp.TextLoader.  It exists to hand
+// each evaluation a PRIVATE tree -- it is the entry point an embedder is
+// pointed at for a reusable parse cache, and macro_stamp_shared_ast_test.go
+// contrasts it with the Load* entry points, which do not copy.  For Cells the
+// privacy was real; for positions it was not, so every evaluation reported
+// its positions through the retained cache's objects.
+//
+// LIVE OR LATENT: latent, and checked rather than assumed -- see the
+// neighbourhood audit on the PR.  No non-test code writes THROUGH a borrowed
+// *token.Location: parser/rdparser's five write sites operate on Locations
+// the nodes they have just built own (#426/#442), lisp/env.go's three error
+// constructors take env.Loc.Copy() (#421), stampMacroExpansion ASSIGNS
+// v.Source rather than writing through it, and every position consumer
+// (lsp/, lint/, analysis/, mcpserver/, minifier/, formatter/, debugger/,
+// profiler/) reads a Location and formats it into its own value type.  So the
+// writes these tests perform are writes the TESTS perform: they demonstrate
+// the mechanism, they do not record an observed corruption.  What is a real
+// observed defect is the SHARING itself, which the tests below assert
+// directly and which failed on 5ef6106.
+//
+// Tests marked GUARD pass before the fix as well as after.  They pin
+// behaviour the fix must not break; they are not catches.
+
+package lisp_test
+
+import (
+	"io"
+	"strings"
+	"testing"
+
+	"github.com/luthersystems/elps/elpsutil"
+	"github.com/luthersystems/elps/lisp"
+	"github.com/luthersystems/elps/parser"
+	"github.com/luthersystems/elps/parser/token"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// readOne parses src and returns its single top-level expression.
+func readOne(t *testing.T, src string) *lisp.LVal {
+	t.Helper()
+	exprs, err := parser.NewReader().Read("copy-source.lisp", strings.NewReader(src))
+	require.NoError(t, err)
+	require.Len(t, exprs, 1)
+	return exprs[0]
+}
+
+// TestCopyDoesNotAliasSourceLocation is the reproduction from issue #446.
+// CATCH: both assertions failed on 5ef6106.
+func TestCopyDoesNotAliasSourceLocation(t *testing.T) {
+	orig := readOne(t, "(+ 1 2)")
+	cp := orig.Copy()
+
+	require.NotNil(t, orig.Source)
+	assert.NotSame(t, orig.Source, cp.Source,
+		"a copy shares the original's *token.Location (#446)")
+
+	cp.Source.Line = 99
+	assert.Equal(t, 1, orig.Source.Line,
+		"a write through the copy moved the original's recorded position (#446)")
+}
+
+// TestCopyDoesNotAliasSourceAtDepth pins that the separation reaches every
+// node copyCells reaches, not just the root.  The issue's probe counted the
+// root and all three cells of (+ 1 2) on one object.
+// CATCH: failed on 5ef6106 at every depth.
+func TestCopyDoesNotAliasSourceAtDepth(t *testing.T) {
+	orig := readOne(t, "(defun f (x) (let ([y (+ x 1)]) (* y y)))")
+	cp := orig.Copy()
+
+	origNodes := flatten(orig)
+	cpNodes := flatten(cp)
+	require.Len(t, cpNodes, len(origNodes), "copy has a different shape")
+	require.Greater(t, len(origNodes), 10, "test program is too small to be interesting")
+
+	shared := 0
+	for i := range origNodes {
+		if origNodes[i].Source != nil && origNodes[i].Source == cpNodes[i].Source {
+			shared++
+		}
+	}
+	assert.Zero(t, shared,
+		"%d of %d copied nodes share the original's *token.Location (#446)",
+		shared, len(origNodes))
+}
+
+// TestTextLoaderEvaluationsGetPrivatePositions is the reason #446 matters.
+// TextLoader retains one parse tree and hands each evaluation expr.Copy();
+// the copy is supposed to be private.  A builtin observes the Location the
+// evaluation actually carries.
+//
+// CATCH: on 5ef6106 both evaluations handed the builtin the RETAINED tree's
+// own *token.Location -- the same pointer, twice, and the cache's.
+func TestTextLoaderEvaluationsGetPrivatePositions(t *testing.T) {
+	env := lisp.NewEnv(nil)
+	env.Runtime.Reader = parser.NewReader()
+	require.NotEqual(t, lisp.LError, lisp.InitializeUserEnv(env).Type)
+
+	var seen []*token.Location
+	env.AddBuiltins(true, elpsutil.Function("probe-loc", lisp.Formals("x"),
+		func(_ *lisp.LEnv, args *lisp.LVal) *lisp.LVal {
+			seen = append(seen, args.Cells[0].Source)
+			return lisp.Nil()
+		}))
+
+	cr := &capturingReader{inner: parser.NewReader()}
+	load, err := lisp.TextLoader(cr, "cache.lisp", strings.NewReader("(probe-loc 'sym)"))
+	require.NoError(t, err)
+	require.Len(t, cr.exprs, 1)
+
+	retained := cr.exprs[0]
+	require.Len(t, retained.Cells, 2)
+	cached := retained.Cells[1].Source
+	require.NotNil(t, cached)
+	cachedLine := cached.Line
+
+	for i := range 2 {
+		require.NotEqual(t, lisp.LError, load(env).Type, "evaluation %d failed", i)
+	}
+	require.Len(t, seen, 2)
+
+	assert.NotSame(t, cached, seen[0],
+		"the first evaluation reports its position through the retained cache's object (#446)")
+	assert.NotSame(t, cached, seen[1],
+		"the second evaluation reports its position through the retained cache's object (#446)")
+	assert.NotSame(t, seen[0], seen[1],
+		"two evaluations of the cached tree share one *token.Location (#446)")
+
+	// The corruption the sharing enables.  The write is this test's, not
+	// production code's -- see the file header.
+	seen[0].Line = 99
+	assert.Equal(t, cachedLine, cached.Line,
+		"a write through one evaluation moved the position the cache will report forever (#446)")
+	assert.Equal(t, cachedLine, seen[1].Line,
+		"a write through one evaluation moved a sibling evaluation's position (#446)")
+}
+
+// TestCopyPreservesSourcePosition pins that separating the objects does not
+// change what they say.
+// GUARD: passes before the fix.
+func TestCopyPreservesSourcePosition(t *testing.T) {
+	orig := readOne(t, "(+ 1 2)")
+	cp := orig.Copy()
+
+	origNodes := flatten(orig)
+	cpNodes := flatten(cp)
+	require.Len(t, cpNodes, len(origNodes))
+	for i := range origNodes {
+		if origNodes[i].Source == nil {
+			assert.Nil(t, cpNodes[i].Source, "node %d: nil Source became non-nil", i)
+			continue
+		}
+		require.NotNil(t, cpNodes[i].Source, "node %d: Source was dropped", i)
+		assert.Equal(t, *origNodes[i].Source, *cpNodes[i].Source,
+			"node %d: copied position differs", i)
+	}
+}
+
+// TestCopyPreservesNilSource pins that a nil Source stays nil.  A nil Source
+// means "no position recorded" throughout the tree and is distinct from a
+// zero one; token.Location.Copy is nil-preserving for that reason (#421).
+// GUARD: passes before the fix.
+func TestCopyPreservesNilSource(t *testing.T) {
+	v := lisp.SExpr([]*lisp.LVal{lisp.Symbol("a")})
+	v.Source = nil
+	v.Cells[0].Source = nil
+	cp := v.Copy()
+	assert.Nil(t, cp.Source)
+	require.Len(t, cp.Cells, 1)
+	assert.Nil(t, cp.Cells[0].Source)
+}
+
+// TestCopyKeepsSharingTheNativeSingleton pins the ONE Location that is
+// deliberately shared and must stay shared: lisp.nativeSource's process-wide
+// singleton, stamped on every natively-constructed LVal.
+//
+// Copying it would buy nothing -- it has a single owner, the process; it is
+// read-only by contract; SingletonSnapshot watches its bit pattern; and the
+// parser never leaves it in an AST (#362/#421).  It would cost a heap
+// allocation on the interpreter's hottest path, which is the +18.6% sec/op,
+// +20.2% allocs/op that #362's own comment records and rejects.  Separating
+// two owners is what the copy is for, and here there is only one.
+//
+// GUARD: passes before the fix, and pins that the fix did not "improve" it.
+func TestCopyKeepsSharingTheNativeSingleton(t *testing.T) {
+	a, b := lisp.Int(1), lisp.Int(2)
+	require.NotNil(t, a.Source)
+	require.Same(t, a.Source, b.Source, "premise: natively-constructed values share one Location")
+
+	cp := a.Copy()
+	assert.Same(t, a.Source, cp.Source,
+		"copying a native value allocated a private Location on a hot path (#362)")
+	assert.Equal(t, token.NativeFile, cp.Source.File)
+}
+
+// TestParseTreeCopyIsFullyPrivate closes the gap the native-singleton
+// exception could otherwise leave in TextLoader's guarantee.  The exception is
+// only harmless for a parse cache if a parse tree never contains that pointer,
+// which is a property #362/#421 states and parser/rdparser's
+// TestParserDoesNotAliasSharedNativeLocation guards at the reader end.  This
+// asserts the consequence at the end that consumes it: for a tree that came
+// from the reader, "no node shares" has no exceptions at all.
+// CATCH: failed on 5ef6106 (every node shared).
+func TestParseTreeCopyIsFullyPrivate(t *testing.T) {
+	const src = `(defmacro m (x) (quasiquote (+ 1 (unquote x))))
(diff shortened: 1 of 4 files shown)


sam-at-luther:
### CI re-measured on a quiet runner, and it agrees to the digit

The body flags that my local arms ran on a box shared with other agents' test runs (load ~20 on 4 cores), so I would not stand behind the local *timing* rows. CI's benchmark job has now measured both arms on its own arm64 runner, interleaved, n=10 — and the allocation deltas come back **identical**, on a different architecture:

| Row | Metric | Local (amd64, loaded) | CI (arm64, quiet) |
|---|---|---|---|
| `Package/get-nested-baseline` | B/op | +4.36% p=0.000 | **+4.36% p=0.000** |
| `Package/get-nested-baseline` | allocs/op | +3.93% p=0.000 | **+3.93% p=0.000** |
| `AliasStableSort` | allocs/op | +0.02% p=0.000 | **+0.02% p=0.000** |
| `AliasSliceList` / `AliasCDR` | allocs/op | +0.01% p=0.000 | **+0.01% p=0.000** |

That is what an allocation-count change should look like: exact, deterministic, and architecture-independent. It also confirms the shape of the cost argument — one row moves, it is the `assert`-heavy one, and the interpreter rows are flat because the values the evaluator copies at runtime carry the native singleton that `LVal.Copy` skips.

**Gate verdict on CI: `0 at or above the gate` (timing 15%, allocation 5%)**, `32 delta rows + 311 no-change rows`, 16 significant moves in the bad direction, none reaching a threshold. No waiver used; the two existing `libjson/Encode` waivers (#412/#410) report unused, as they did locally, and were left alone.

Two things the quiet runner settles that the loaded one could not:

- **`PackageGetVal`'s local +8.11% sec/op was noise, as claimed.** CI reports it `~ (p=0.224 n=10)` at ±2–3% spread. The body's reasoning — a benchmark whose `allocs/op` is `0.000` on both arms cannot be moved by an allocation — holds.
- **The real timing cost is visible and small.** `Package/get-nested-baseline` is `+2.87% (p=0.002)` sec/op on CI, against a 15% gate. That is the 20k `Location` allocations showing up in time as well as in count, on the one benchmark that does 5,000 `assert` evaluations. Every other timing row in the tree is `~`, and the `lisp` geomean is `+0.24%`.

The full benchstat table and gate report are in the bot comment above.

---

Issue #471, opened 2026-08-16
libelpspath: the in-place ?del! compacts through a backing array it does not own, scrambling a sequence the view aliases


sam-at-luther:
Found **running**, on `claude/libelpspath-rebased` (PR #402's content rebased onto `main` at `8402c49`). Not fixed there — reported only. Reproduces with `elps run`.

`indexPath.deleteMutate` and `rangePath.deleteMutate` build their result by compacting **in place**, through the input's own backing array:

```go
// indexPath.deleteMutate
vals := append(cells[:index], cells[index+1:]...)
storeCells(in, vals)

// rangePath.deleteMutate
vals := cells[:from]
if to < n {
	vals = append(vals, cells[to:]...)
}
storeCells(in, vals)
```

Both `append`s write into `cells`' array. When `in` is a **view** over a longer sequence — which the mutating builtins accept, because a view is an ordinary array `LVal` — the shift lands in the aliased source.

## Reproduction

```lisp
(use-package 'elpspath)
(set 'v (vector 1 2 3 4 5))
(set 'w (slice 'vector v 0 3))   ; the KERNEL's own producer
(elpspath:?del! w 0)
(debug-print (list 'view w 'src v))
```

```
'('view (vector 2 3) 'src (vector 2 3 3 4 5))
```

`v` was never named by the operation. It did not shrink — it cannot — it was **scrambled**: `1` is gone and `3` is duplicated. Nothing raised.

The producer does not have to be the kernel's `slice`, and the view does not have to be bound to a variable the caller thinks of as a view:

```lisp
(set 'doc (sorted-map "a" (vector 1 2 3 4 5)))
(set 'view (elpspath:? doc "a" '(range 0 3)))
(elpspath:?del! view 0)
(elpspath:? doc "a")            ; => (vector 2 3 3 4 5)
```

Range step and index step behave identically:

```
?del! w 0                => src (vector 2 3 3 4 5)
?del! w '(range 0 1)     => src (vector 2 3 3 4 5)
```

## Not the same as the aliasing that is already handled

Three controls, all on the same tree, separate this from what is filed and from what is expected:

| | result | verdict |
|---|---|---|
| `?set!` at an index through a view | `src (vector 97 2 3 4 5)` | **expected.** `setMutate` does `cells[index] = newIn`; a view aliases its elements, so assigning one is assigning the source's. Same for `?nil!`. |
[...]

Issue #458, opened 2026-08-15
fuzz-budget-check passes at zero headroom — the state its own workflow comment calls "a truncated sweep"


sam-at-luther:
Found by **running**, during the neighbourhood audit for #443 ("every other threshold in CI that could be below its own measurement resolution"). Same family as #443: a gate that reports success in a state its own comments describe as broken.

## What it reports today, on `main` at `bfb6ee8`

```
$ ./scripts/fuzz-budget-check.sh
fuzz-budget-check:
  targets discovered      28
  shards                  6
  largest shard           5 target(s)
  scheduled FUZZTIME      10m (10 min/target)
  fixed overhead          10 min
  required                60 min
  timeout-minutes         60 min
  headroom                0 min
fuzz-budget-check: the sweep fits.          <-- exit 0
```

The gate's condition is `needed > timeout_minutes`, so `needed == timeout` passes.

`.github/workflows/fuzz.yml` already says this is wrong, in its own words:

> **THE HEADROOM IS THE POINT, NOT THE GREEN TICK.** The gate's condition is `needed > timeout`, so a sweep sized at exactly timeout-minutes PASSES — while being the silently-truncated run the gate exists to prevent, since observed shard durations already vary from 4m25s to 5m42s.

The comment goes on to describe the exact state we are now in, as a thing that happened before and was fixed by adding a shard. It has recurred: the same comment's worked example says "At 23 targets over 6 shards that is 4 x 10m + 10m = 50", but discovery now finds **28** targets, so `ceil(28/6) = 5` and `needed` is 60 against a 60-minute timeout.

Why that matters is stated in the script's own header: when the job hits its timeout, "the LAST targets in the sweep never execute at all, and the run is simply... cancelled. Nothing says 'coverage was dropped'. A truncated sweep and a finished one look alike in the check list." So the failure is silent on both sides — the gate says "fits", and the truncated run says nothing.

## Proposed fix: require a margin, and derive the margin rather than pick a round number

Change the condition from *non-exceedance* to *non-exceedance with a margin*:

```
needed + margin <= timeout_minutes
```

[...]

Issue #444, opened 2026-08-15
lint.LintFileWithAnalysis stamps Filename into the caller's *analysis.Config, racing concurrent callers


sam-at-luther:
Found by **reading** `lint/lint.go` during the #437 neighbourhood audit (every place `lint`/`analysis` mutates caller-supplied state), then **confirmed by running** a throwaway `-race` probe. The probe is reproduced below; it is not committed anywhere.

## The defect

`lint/lint.go`, `LintFileWithAnalysis`:

```go
func (l *Linter) LintFileWithAnalysis(source []byte, filename string, cfg *analysis.Config) ([]Diagnostic, error) {
	...
	if cfg == nil {
		cfg = &analysis.Config{}
	}
	cfg.Filename = filename        // writes into the caller's config
	result := analysis.Analyze(exprs, cfg)
```

The exported method takes a `*analysis.Config` the caller owns and writes a field of it. Two consequences for an embedder that holds one config and lints several files:

1. **Concurrent calls race.** Two goroutines linting different files through one config write `Filename` unsynchronised.
2. **Whoever ran last wins.** After the call the caller's config names a file the caller did not ask about, so any later analysis through that config is attributed to the wrong file.

The repo already has the fix for exactly this, one package over: `analysis.ConfigForFile(cfg, filename)` shallow-copies the config and stamps `Filename` on the copy. `AnalyzeFile` and (since #439) `mcpserver`'s lint tool both go through it. `LintFileWithAnalysis` is the one that still writes through.

## Probe output

```go
cfg := &analysis.Config{}
// two goroutines, same cfg, different filenames
l.LintFileWithAnalysis([]byte("(defun foo () 1)"), "a.lisp", cfg)
l.LintFileWithAnalysis([]byte("(defun foo () 1)"), "b.lisp", cfg)
```

`go test ./lint/ -race`:

```
WARNING: DATA RACE
Write at 0x00c0001524f0 by goroutine 11:
  github.com/luthersystems/elps/lint.(*Linter).LintFileWithAnalysis()
      lint/lint.go:665
Previous write at 0x00c0001524f0 by goroutine 10:
  github.com/luthersystems/elps/lint.(*Linter).LintFileWithAnalysis()
      lint/lint.go:665
```

## Blast radius

[...]

Issue #437, opened 2026-08-15
lint.BuildAnalysisConfig mutates the caller's LintConfig.StdlibExports map


sam-at-luther:
Found by **reading** `lint/lint.go` while working on #424, then **confirmed by running** a throwaway probe. The probe is reproduced below; it is not committed anywhere.

## The defect

`lint/lint.go` `BuildAnalysisConfig`:

```go
pkgExports := cfg.StdlibExports          // NOT copied
if pkgExports == nil {
	pkgExports = defaultStdlibExports()
}
if cfg.Registry != nil {
	for pkg, regSyms := range analysis.ExtractPackageExports(cfg.Registry) {
		pkgExports[pkg] = append(pkgExports[pkg], regSyms...)   // writes into the caller's map
	}
}
for pkg, wsSyms := range prescan.PkgExports {
	pkgExports[pkg] = append(pkgExports[pkg], wsSyms...)        // ditto
}
```

`pkgExports` aliases `cfg.StdlibExports` whenever the caller set it, so registry exports and **workspace** exports are written into the caller's map.

The field's own doc comment is what makes this a defect rather than a quirk — it invites exactly the sharing that breaks:

> Embedders that already have a configured env can pass `analysis.ExtractPackageExports(env.Runtime.Registry)` here to avoid the overhead of creating a temporary environment.

An embedder that follows that advice and keeps the map around gets two things it did not ask for:

1. Its "stdlib exports" map silently acquires symbols exported by whatever workspace was last linted.
2. Repeated calls accumulate duplicates without bound — the same symbol is appended once per call.

## Probe output

```go
shared := map[string][]analysis.ExternalSymbol{}
BuildAnalysisConfig(&LintConfig{Workspace: dir, StdlibExports: shared})
BuildAnalysisConfig(&LintConfig{Workspace: dir, StdlibExports: shared})
```

with `dir` containing `(in-package 'mypkg) (export 'my-workspace-fn) (defun my-workspace-fn () 1)`:

```
caller's StdlibExports map: 0 packages before, 1 after
shared["mypkg"] after 1 call = 1 symbols
shared["mypkg"] after 2 calls = 2 symbols
```

## Blast radius

[...]

PR #472, opened 2026-08-16
Land libelpspath on main with the copy and cyclic-value fixes (#393, #395, #401)

### LINKED ISSUE #369: Quoted program literals can be corrupted in place from pure lisp (stable-sort; append 'vector via slice-retained capacity)

### LINKED ISSUE #393: libelpspath's walkers are unguarded against cycles — a process kill #391 never had the chance to cover

### LINKED ISSUE #395: libelpspath copyMap is shallow where copyList/copyVector are deep, so ?set/?del/?nil on a sorted-map return a sharing "copy"

### LINKED ISSUE #401: libelpspath's copying ?set/?del/?nil write the ARRAY layout onto a LIST copy, corrupting a leaf the source still holds

sam-at-luther:
Supersedes #402, rebased from `f18e118` onto `1396f32`. Same package, one commit dropped, one added, one new fix that the new base made necessary. **Merged as `1c13c76`.**

`Fixes #393`, `Fixes #395`, `Fixes #401`.

## The split was investigated and is not coherent

The brief was to extract the three bug fixes into a `main`-based PR and leave the framework behind. That is not possible, and the reason is structural rather than a judgement call.

**`libelpspath` did not exist on `main`.** Every file this PR touches is new:

```
$ git ls-tree -r --name-only origin/main | grep -i elpspath
(no output)

$ git diff --name-status 1396f32..HEAD | awk '{print $1}' | sort | uniq -c
     22 A
      2 M
$ git diff --name-status 1396f32..HEAD | grep -v '^A'
M	.github/workflows/fuzz.yml
M	lisp/lisplib/lisplib.go
```

All 22 added files are under `lisp/lisplib/libelpspath/`. Only two existing files are touched at all: `lisp/lisplib/lisplib.go`, +5 lines, which registers the package, and `.github/workflows/fuzz.yml`, comment text only. **No existing Go code is modified and nothing is deleted.**

The functions the three issues name are in that new code and nowhere else:

```
$ git grep -n -E 'func (copyMap|copyList|copyVector|storeCells)' origin/main -- '*.go'
(no output)
```

So a PR "carrying only the fixes" against `main` would have been an empty diff. There was no `copyMap` on `main` to make deep, no `storeCells` to teach about list layout, and no walker to guard. Issue #393 says as much itself: *"`libelpspath` **did not exist in the tree #391 swept** … the **only** non-test lisplib package present on the chain and absent from `main`."*

The fixes are real, and they are separable from the **sealed-AST feature** — that separation is what #402 already did. They are not separable from the package, because they are corrections applied to code as it is adopted. **This is all-or-nothing.**

## What was dropped

**`d27b466` — "ci: sixth fuzz shard for the 21st target, restoring the sweep's headroom".** Dropped, not rebased.

That commit was written against a `main` at 20 targets over **5** shards. `main` is now at 29 targets over **8**. Replaying it would take the matrix from 8 shards down to 6 — and #458 is exactly why that would not have been caught: `fuzz-budget-check.sh` passes at zero headroom. Both arms measured on this tree:

| shards | targets | largest shard | required | timeout | headroom | gate |
|---|---|---|---|---|---|---|
| 6 (had `d27b466` been replayed) | 30 | 5 | 60 min | 60 min | **0 min** | `the sweep fits`, **exit 0** |
| 8 (this PR) | 30 | 4 | 50 min | 60 min | 10 min | `the sweep fits`, exit 0 |

A silent shrink of the nightly sweep behind a green tick. `fuzz.yml`'s own comment names that failure mode in capitals.

In its place, `810d456` re-derives the budget for the target this PR actually adds. `libelpspath.FuzzPathEngine` is the 30th, `ceil(30/8) = 4`, and no shard is needed. It also corrects the derived summary line, which still read *"23 targets over 6 shards"* while the matrix said 8 — the seventh and eighth shards each updated only their own paragraph.

## What was kept, deliberately: the cycle-guard commit

`da21e81` on #402, `c444c58` here — *"lisp: unexport the cycle guard, giving libelpspath its own"* — was flagged for scrutiny because it touches `lisp/` core. Against **this base it does not**. `lisp/cycleexport.go` has never existed on `main`; it is added by `9d4734b` earlier in this same branch and removed again by `c444c58`. The two are intra-PR churn, and the net effect on package `lisp` is nil:

```
$ git diff --stat 1396f32..HEAD -- lisp/ ':!lisp/lisplib/libelpspath/'
 lisp/lisplib/lisplib.go | 5 +++++
```

So the question "which PR does it belong in" dissolves: with no split to allocate it to, and no change to package `lisp` to review, it stays where it is. It is worth keeping rather than squashing away because its message records *why* the guard is not exported — nothing else took the export up, and #391 measured what a shared guard next to `String`'s hot loop costs.

It does have one real consequence, and it is a good one: the guard became caller-owned state, which took a guarded walk from **1 allocation to 0**.

## The three bugs, red-then-green on this base

Each red arm is this tree with that one fix reverted — the only coherent red, since `main` had no code to break. All three still reproduce; nothing landed that day fixed or shadowed them.

### #393 — a cyclic value kills the process

```lisp
(set 'v (vector 1 2))
(append! v v)
(handler-bind ((condition (lambda (&rest c) 'caught-a-condition))) (? v 0))
```

| | exit | output |
|---|---|---|
| red (`cyclic = false` at both `descend` sites) | **2** | `fatal error: stack overflow`, 292 lines |
| green | **0** | `'caught-a-condition` then `'survived` |

Asserted out of process: a stack overflow is a fatal runtime error, not a panic, so `recover()` never sees it and `handler-bind` cannot.

### #395 — the map copy shares its nested containers

The package's own redaction example:

```lisp
(set 'patient (sorted-map "ssn" "123" "address" (sorted-map "city" "London")))
(set 'redacted (?nil patient "ssn"))
(?set! redacted "address" "city" "REDACTED")
```

| | `(list (? redacted "address" "city") (? patient "address" "city"))` |
|---|---|
| red (`val := pair.Cells[1]`) | `'("REDACTED" "REDACTED")` |
| green | `'("REDACTED" "London")` |

The write through the redacted copy reached the patient record.

### #401 — the copy writes the array layout onto a list

```lisp
(set 'doc (sorted-map "l" '(1 2)))
(set 'cp  (?set doc "tag" "x"))
(set 'res (?del cp "l" 0))
```

| | result | one-element list through `handler-bind` |
|---|---|---|
| red (`storeCells` array-only) | `'source-list '(1 1)  'copy-list '(1 1)  'result-list (1 1)` | `internal-panic: index out of range [1] with length 1`, **exit 1** |
| green | `'source-list '(1 2)  'copy-list '(1 2)  'result-list (2)` | `()`, exit 0 |

`'(1 2)` became `'(1 1)` in the source document. A deep copy shares its immutable leaves, so the write-back landed on a leaf the source still held.

## One fix the rebase itself made necessary

`rangePath.Get` sliced `cells[from:to]`, handing back a view that keeps the source's spare capacity. The in-file comment argued — correctly, at `f18e118` — that clamping would close #369 for one producer and leave every other one open, *"because the kernel's own `(slice 'vector …)` hands back exactly the same thing."*

**That premise is false against this base.** #373 settled the class the other way, and `lisp.clampCap` now runs at every kernel producer: *"every sequence view that escapes into lisp is clamped where it is produced."* Measured on the rebased tree before the clamp went in:

```
(set 'v (vector 1 2 3 4 5))
(set 'w (? v '(range 0 3)))      (append! w 99)
  view (vector 1 2 3 99)   src (vector 1 2 3 99 5)    <- this Get, unclamped
(set 'w (slice 'vector v 0 3))   (append! w 99)
  view (vector 1 2 3 99)   src (vector 1 2 3 4 5)     <- the kernel, clamped
```

Landing it unclamped would have introduced the tree's **only** remaining producer of an aliasing class `main` had just closed. `fae8ceb` clamps it with a three-index slice, which allocates nothing and copies nothing.

Red-proofed by reverting to the two-index slice:

- `TestRangeGetClampsCapacity` — red `"range view kept the source's spare capacity: len 3 cap 5"` on every strict-prefix range, both shapes. The `to == n` arms pass in red — there is no spare capacity to keep — so they are **guards**, not catches, and are labelled as such in-file.
- `libelpspath_test.lisp` `"range view does not alias through spare capacity"` — the end-to-end arm, red on the same revert.
- Its neighbour, `"kernel slice view does not alias either"`, is green in **both** arms by design. It is a **guard**: it exists to go red if the settlement itself ever moves, so the clamp above is re-decided rather than silently left behind.

## Neighbourhood audit

Against the aliasing families fixed that day. Citing the existing audits rather than redoing them.

| family | instances in `libelpspath` |
|---|---|
| **#362 / #366 / #426 / #431 / #446** — `*token.Location` ownership | **none.** The package contains no reference to `Source`, `token.Location`, or `LVal.Copy`. Its constructors (`lisp.Array`, `lisp.SortedMap`, `lisp.QExpr`) produce position-free values, so there is no position to share. |
| **#437 / #444** — caller-supplied config retained | **none.** `LoadPackage(env)` takes no config. The only caller-supplied slices are the variadics of `Chain` and `Iter`, and `normalizePaths` rebuilds rather than retains them. |
| **#369 / #373** — sequence views | **one beyond the three filed: `rangePath.Get`, fixed above.** The `rangePath.setMutate` splice, the other member of this family, was already fixed on the branch by `442199c`. |

The audit also turned up a defect that is **not** in any of those families and is **not fixed here** — reported as **#471**. `indexPath.deleteMutate` and `rangePath.deleteMutate` compact through a backing array they do not own, so `?del!` on a *view* scrambles the sequence it aliases:

```lisp
(set 'v (vector 1 2 3 4 5))
(set 'w (slice 'vector v 0 3))   ; the KERNEL's producer
(?del! w 0)
;; view (vector 2 3)   src (vector 2 3 3 4 5)
```

It is the same mistake `442199c` fixed for `setMutate` — *"appending into a backing array the function does not own"* — left standing on the two delete paths. No capacity clamp reaches it: the write is *within* `len`. It is left for #471 rather than folded in here, because the fix costs an allocation on a path that currently does none and wants its own measurement. The alias battery does not see it because the battery writes through **copies**, and a view is not a copy — a coverage gap, not a false claim.

## Cycle guard cost

`BenchmarkCycleGuardCost`, `-count=12 -cpu=1`, both arms alternating inside a single process so machine drift moves both. `benchstat -col /arm`; the base column is **guarded**, so a negative number is the unguarded arm being faster — that is, the guard's cost.

```
                                         |   guarded   |             unguarded              |
                                         |   sec/op    |   sec/op     vs base               |
CycleGuardCost/doc=spine8/walk=typecheck   22.12u +/-1%  21.74u +/-1%  -1.72% (p=0.000 n=12)
CycleGuardCost/doc=spine8/walk=copy        41.20u +/-1%  40.29u +/-0%  -2.23% (p=0.000 n=12)
CycleGuardCost/doc=record/walk=typecheck   6.054u +/-2%  5.855u +/-2%  -3.30% (p=0.000 n=12)
CycleGuardCost/doc=record/walk=copy        10.51u +/-1%  10.25u +/-0%  -2.47% (p=0.000 n=12)
geomean                                    15.52u        15.14u        -2.43%

                                         B/op and allocs/op, all four rows:
                                           ~ (p=1.000 n=12)  -- all samples are equal
```

**Honestly: the guard is free on memory and it is not free on time.**

- **Allocations: zero difference.** Every row is `all samples are equal` on both `B/op` and `allocs/op` — 78/141/288/568 allocations, identical in both arms. This is a change from what #402 reported (+1 allocation, +16 B/op), and `c444c58` is why: the guard's state became a caller-owned local, so escape analysis keeps it on the stack. The exported `lisp.CycleGuard` had to allocate the shared state itself across a package boundary. `TestCycleGuardAllocationCost` pins the 0, and `TestCycleGuardAllocationDoesNotScale` pins that it does not grow with width — including an arm placing *n* siblings at exactly `cycleGuardDepth`, the one depth at which a by-value guard would build a path set per node. Both pass.
- **Time: +1.7% to +3.3%, geomean +2.5%.** All four rows resolve at `p=0.000`, and each move is larger than that row's own spread (+/-0% to +/-2%), so per #452 these are real signals rather than noise — not rows to wave through. They are also an order of magnitude under the 15% timing threshold, and #391 measured this shape before: a guard on a walk costs a little, which is why the kernel's rendering and equality walks were kept out of it.

**Benchmark gate.** This PR is purely additive, so no benchmark that exists on `1396f32` changes — there is no cross-base row to compare, because every benchmark in the diff is new. The measurement above is the internal guarded-vs-unguarded comparison the package carries for exactly this purpose, and it is reported rather than gated. CI's `Required: benchmark` agrees: green.

## Gates

All run locally against `1396f32`.

| gate | result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `go test -count=1 ./...` | ok, exit 0 |
| `go test -race -count=1 ./...` | ok, exit 0, no `DATA RACE` |
| `elps fmt -l ./...` | no output, exit 0 |
| `elps doc -m` | `All functions and exports are documented.` |
| `make go-test` | exit 0 |
| `make race` | exit 0 |
| `make test-elpscheck` | exit 0 |
| `make test-examples` | exit 0 |
| `./scripts/ci-gates-test.sh` | **233 passed, 0 failed** |
| `./scripts/confidentiality-guard.sh` | `confidentiality guard: clean` |
| `./scripts/fuzz-budget-check.sh` | 30 targets / 8 shards, required 50 min, **headroom 10 min** |

`gofmt -l` reports 17 files on this tree, all of them pre-existing on `1396f32`; every file in this diff is `gofmt`-clean.

CI: 22 of 22 checks green on `fae8ceb`, including all five `Required:` aggregates and all 8 fuzz shards.

## What is left in #402

#402 carries nothing this PR does not, apart from `d27b466` — the sixth-shard commit, which is stale and should not land. Safe to close as superseded.


sam-at-luther:
Merged as `1c13c76` on `main`. All 22 checks green on the final head `fae8ceb` against base `1396f32`: `Required: build & test`, `Required: fuzz` (8/8 shards), `Required: benchmark`, `Required: govulncheck`, `Required: tree-sitter`, plus `Lint & Test`, `Build (Windows)`, `CI Gate Self-Test` and `Benchmark Comparison`.

`CI Gate Self-Test` completed in 17s rather than wedging, so #468 did not bite on either run.

---

Issue #397, opened 2026-08-13
Package.Get writes funNames unsynchronised, so concurrent MCP doc requests can kill the server with a fatal map write


sam-at-luther:
`Package.Get` is named as a read and behaves as a write:

```go
func (pkg *Package) Get(k *LVal) *LVal {
	v := pkg.get(k)
	if v.Type == LFun {
		if fid := v.FID(); pkg.funNames[fid] != k.Str {
			pkg.funNames[fid] = k.Str      // unsynchronised map write
		}
	}
	return v
}
```

`Package` carries no mutex. `lisp/package.go:114-115` and `:241` write the same map.

## The reachable path

`mcpserver/service.go` builds a **fresh doc environment per request** but shares the **same `*Package` values** into every one:

```go
env, err := lisplib.NewDocEnv()
...
for _, name := range s.registry.PackageNames() {
	env.Runtime.Registry.AddPackage(s.registry.Package(name))   // same *Package, every request
}
```

So two concurrent doc requests that resolve a function in the same embedder package race on `funNames`. Go's runtime throws on concurrent map write **and** on a read racing a write:

```
fatal error: concurrent map writes
```

That is **not recoverable** — no `recover()`, no `handler-bind`. Same unrecoverable class as elps#393, reached through a different door.

The window is not only first-call: the guard `pkg.funNames[fid] != k.Str` is itself an unsynchronised read of a map another goroutine may be writing, which the runtime also treats as fatal.

`cmd/doc.go:55` shares packages the same way; it is single-threaded today, so the MCP server is the live exposure.

## Not the same issue as the sharing itself

Sharing one `*Package` across registries is a deliberate Go-API ownership contract — `AddPackage`'s two in-repo callers exist to do exactly that, a `Package`'s symbols are builtin closures and native payloads that `Copy` cannot deep-copy and `SealAST` declines to descend into, so there is no copy-and-seal analogue. That contract is fine; it was merely undocumented, and `Package`'s own doc comment ("belongs to the LEnv that creates it") contradicted it.

**This issue is narrower: a supported sharing pattern is not concurrency-safe, because a method documented as a getter mutates.** Either the mutation needs synchronising, or `Get` must stop writing and the name-tracking move to a write path (`Put`/`Export`, which `:241` already does), or the contract must say single-goroutine and the MCP server must stop sharing.

[...]

PR #478, opened 2026-08-16
libelpspath: stop ?del! scrambling the source a view aliases (#471), and sweep the view axis

### LINKED ISSUE #471: libelpspath: the in-place ?del! compacts through a backing array it does not own, scrambling a sequence the view aliases

sam-at-luther:
Fixes #471.

Based on `main`, not on #472's branch: #472 merged while this was in progress, so `libelpspath` is on `main` at `1c13c76` and #471 is a live defect on the default branch rather than one about to land. The branch was reset onto `main` and the fix reapplied; the diff is five files, all under `lisp/lisplib/libelpspath/`.

## Verdict

**`libelpspath` is clean apart from #471.** The four families were swept with probes that were run, not read. #471 reproduced, was fixed, and its reach turned out wider than filed. Nothing else in the package writes through storage it does not own, retains a caller's container, shares a `*token.Location`, or walks a cyclic value unguarded. What was checked and found clean is listed below, because a clean verdict is only worth what the probes behind it are worth.

One correction to the issue as filed: #471 is **not** limited to strict sub-windows.

## #471

`indexPath.deleteMutate` and `rangePath.deleteMutate` compacted in place, through the input's own backing array:

```go
vals := append(cells[:index], cells[index+1:]...)
vals := cells[:from]; vals = append(vals, cells[to:]...)
```

Both appends shift the tail left **through** `cells`' array. That array is not the function's to write whenever `in` is a **view** — an ordinary array `LVal` whose cells are a window onto a longer sequence, which the kernel's `(slice 'vector …)` and this package's own `rangePath.Get` both hand out, and which nothing in an `LVal` marks.

Reproduced **running**, on the rebased tree:

```lisp
(set 'v (vector 1 2 3 4 5))
(set 'w (slice 'vector v 0 3))
(elpspath:?del! w 0)
;; view (vector 2 3)   src (vector 2 3 3 4 5)
```

`v` was never named by the operation. It did not shrink — it cannot — it was scrambled: `1` gone, `3` duplicated, nothing raised.

**The answer was right, which is why nothing caught it.** A left shift copies before it overwrites, so the view came out correct and only the aliased source was damaged. That is the whole difference from `rangePath.setMutate`, which had the identical defect and was found immediately because its overlap corrupted its own answer.

No capacity clamp reaches it: #369/#373 close appends **past** `len`, and this writes **within** `len`.

### Two shapes the issue did not list

Found **running**, while sweeping around the reported ones:

| shape | before | after |
|---|---|---|
| **full-length view** — `(slice 'vector v 0 5)` on a 5-element `v` | `src (vector 2 3 4 5 5)` | `src (vector 1 2 3 4 5)` |
| **view of a view** — `(slice 'vector (slice 'vector v 0 4) 0 2)` | `mid (vector 2 2 3 4)`, `src (vector 2 2 3 4 5)` | both intact |

The full-length case matters for the framing: a window covering the whole sequence aliases just as completely as a partial one. "Being the whole sequence" is not the same as owning it, so a fix conditioned on the window being shorter than its backing array would have been wrong.

Cases that were **accidentally** correct before the fix, and are correct on purpose now: deleting a suffix (including the whole window) shifts nothing, because there is no tail to move. Only a delete with a non-empty tail wrote through. That narrowed the shapes that could expose it without making any of them safe.

### The fix

Both paths build the compaction in a slice they allocate, with an exact capacity, and hand it to `storeCells` as before. A delete through a view now leaves the source completely alone. The change is ~10 lines of code in two functions.

The issue also floats making `?del!` on a view an **error**. Not done, and deliberately: there is no reliable way to ask an `LVal` whether it is a view — `cap` is not it, since the kernel clamps its own producers — so the test would be both unsound and a behaviour change. Allocating is what the issue proposes and it is sufficient.

## Everything else: what was checked, and how

Every row was **run**. Where a family is clean by construction rather than by probe, that is said explicitly.

| family | probe | result |
|---|---|---|
| **#369/#373 sequence views** | audited every slice expression in the package's four non-test files, then checked `len == cap` for every `(from,to)` on lengths 0..8, both shapes | **clean.** `rangePath.Get` is the *only* function returning a sub-sequence of a caller's array, and its clamp holds in all 162 cases |
| **#362/#366/#426/#431/#446 `*token.Location`** | grep for `token.`, `.Source`, `.Copy()` across non-test sources | **clean, by construction.** Zero references. The package never reads, writes or retains a position; its constructors take `nativeSource()`'s shared singleton, which the kernel's own snapshot guard already watches |
| **#437/#444 caller-supplied containers** | passed each of the 7 builtins an argument array with spare capacity holding a sentinel, across 4 step shapes; separately mutated a caller's `[]Path` after `Chain` had consumed it | **clean.** No builtin replaced an element of the caller's array or appended into its spare capacity; `Chain` neither retains nor writes the caller's slice (`expandPaths` always rebuilds) |
| **#390/#391/#393 unguarded walks** | 21 cyclic shapes through all 7 builtins out of process: direct self-containing vector, cycle nested inside a map, cycle as the *new value* of a set, self-containing sorted-map, and a mutual two-vector cycle | **clean.** All 21 raised a catchable condition; the process survived. The Go-embedder route that bypasses `okSimpleType` is guarded too (`Set`/`Delete`/`Nil` via the exported `Path` all refuse) |
| **guard does not over-reject** | deep acyclic values at depths 10/63/64/65/200/5000, and a DAG with a repeated sibling buried past `cycleGuardDepth` | **clean.** All accepted and copied in full — the guard costs no correctness on the shapes it must not refuse |
| **#395 shallow-copy-claiming-deep** | the package's own `TestCopyHelpersAgreeOnNestingDepth` plus the copy battery | **clean.** All copy entry points route through `copyGuarded` |
| **#397/#364/#420 unsynchronised state** | enumerated package-level state; 16 goroutines each building an env, loading the package and running 150 operations, under `-race` | **clean.** The only package-level state is the read-only `builtins` templates and an immutable error. The cycle guard's map is a caller-owned local (`var st cycleState`) per walk, never shared. Formals templates verified byte-identical and pointer-identical after 20 loads |

### The `rangePath.Get` clamp on `main`

Checked rather than assumed, since it was written against a premise (`slice` retains capacity) that #373 had already invalidated. It is present as `cells[from:to:to]`, which is `len == cap`, i.e. exactly what `lisp.clampCap` produces. Verified by running: `len == cap` for every `(from,to)` pair on lengths 0..8 for both vectors and lists, and `(append! (? v '(range 0 3)) 99)` leaves `v` alone. **Correct, and not re-touched here.**

## Closing the two blind spots

The fix alone would leave in place the reason it was missed. Two things could have seen this and neither did, for the same reason.

**The alias battery writes through copies.** A view is not a copy, and it had no arm that mutates one — the issue names this as a coverage gap. `path_view_alias_test.go` adds the axis.

The property is: **a mutating operation on a view may change the backing array only at the positions its path names.** Stated that way rather than "must not touch the source", because element assignment through a view is legitimate and documented — `(?set! w 0 97)` does set the source's element 0. #471 is not that: it moved every element after the deleted one, positions no path named. That formulation catches #471 everywhere it is reachable while permitting the aliasing that is meant to work.

- 255 cases per operation across window sizes, pads and every index and range step, for all three mutating operations. All clean.
- **Red-proof**: the same sweep against a replica of the pre-#471 bodies fails **57 of 255**.
- `TestInPlaceReplicaMatchesTheShippedAnswer` pins the replica as a replica — identical answers in all 85 comparable cases, differing only in what it writes through. That identity *is* the defect's silence, and it makes the red-proof evidence about the shipped code rather than about a straw man that is merely broken.

**The fuzzer's documents are all harness-built**, so their cells belong to nobody else and a write past what the path named lands in the harness's own array. `FuzzPathEngine` gains **invariant 7**, folded into the existing target rather than added as a new one.

Two things were needed and either alone is insufficient:

- **Order.** Built *after* the main call, the witness copies a document a `!` operation has already shrunk, so the window is too small to show an unnamed shift — the property passes on a tree that has #471 in it. It is now taken before the call, like the splice oracle. This was caught by testing the red arm, not by reasoning.
- **Seeds.** The general drivers reach the shape **zero** times: it needs a vector document *and* exactly one step *and* a step naming a strict non-empty subset of the positions, and nothing in the existing corpus lines all three up. 14 dedicated seeds — one per operation per step form — were found by search over short inputs.

`TestViewWitnessSeedsAreReached` pins both facts, including the zero, so the seeds do not later look like superstition.

**Red before green**, verified: with the fix reverted the seed corpus fails on the `?del!` index and `?del!` range seeds; restored, green.

```
--- FAIL: FuzzPathEngine/seed#62
    ?del! through a view wrote into the backing array it does not own:
    position 1, inside the view's window changed from () (0xc0000efd50) to () (0xc0000efdc0),
    and the path does not name it (named: [0])
```

Fuzz budget is unchanged — still 30 targets over 8 shards, required 50 min, **headroom 10 min**, and `.github/` is untouched (`shard: [1..8]`).

The lisp-visible arms in `libelpspath_test.lisp` cover both producers, both step forms, offset and full-length windows, and a view reached through a document — plus two **guard** arms that pass in both directions (element assignment through a view still reaching the source; the range splice still not), so the semantics around the fix are re-decided rather than drifted.

## Benchmarks

The fix trades an allocation for correctness, so it is measured rather than asserted to be free. Both arms alternate inside one process; `benchstat -col /arm`, base column is the allocating (fixed) arm, so a negative number is the old in-place arm being cheaper.

**Allocations: exactly +1 per delete, every row, no exceptions.**

```
                                  │ allocating  │              inplace               │
                                  │  allocs/op  │ allocs/op   vs base                │
DeleteCompaction/n=4/step=index     10.000 ± 0%   9.000 ± 0%  -10.00% (p=0.000 n=12)
DeleteCompaction/n=16/step=index    10.000 ± 0%   9.000 ± 0%  -10.00% (p=0.000 n=12)
DeleteCompaction/n=64/step=index    10.000 ± 0%   9.000 ± 0%  -10.00% (p=0.000 n=12)
  (all nine rows identical: 10 vs 9)
```

`B/op` grows with the sequence, as a copied slice must: +24 B at n=4, +128 B at n=16, +512 B at n=64 — 8 bytes per element pointer.

**Time: geomean +7.1%, worst row +12.5%, all under the 15% threshold.**

```
                                  │  allocating  │               inplace                │
DeleteCompaction/n=4/step=index     592.7n ± 15%   585.8n ± 11%        ~ (p=0.219 n=12)
DeleteCompaction/n=16/step=index    809.3n ±  8%   707.9n ±  3%  -12.52% (p=0.000 n=12)
DeleteCompaction/n=64/step=index    1.278µ ±  2%   1.198µ ±  1%   -6.22% (p=0.000 n=12)
geomean                             844.6n         785.0n         -7.06%
```

Worth recording per #452: a first pass at `-benchtime=200x` reported one row at **-34.5% (p=0.005)** with ±21%/±19% spreads — a move that exceeds its own row's spread, and so would have been reported as a real regression. At `-benchtime=50000x` that same row is `~ (p=0.219)`. It was measurement noise, and the rule that a move must beat its row's spread does not by itself protect against an under-iterated benchmark.

This is a micro-benchmark of the delete step alone, so it is the worst case. Cross-base rows below.

### Cross-base rows: zero, and why that is not reassurance

Every benchmark that exists on both `1c13c76` and this branch is **unchanged on both allocation metrics** — `~ (p=1.000 n=6)`, "all samples are equal", on `B/op` and `allocs/op` alike, including all four `BuiltinDelCopyDeep` depths.

That zero needs its explanation stated rather than banked, because on its own it would be a false reassurance:

```
$ grep -oE 'lisp\.Int|rangeStep' path_bench_test.go path_bench_depth_test.go
(no output)
```

**The existing benchmark suite never passes an integer or range step.** Every step in it is a map key, so no benchmark on `main` reaches `indexPath.deleteMutate` or `rangePath.deleteMutate` at all. The rows are unchanged because they do not execute the changed code — not because the change is free. `BenchmarkDeleteCompaction` is added for exactly that reason, and it is where the real +1 allocation shows.

Timing on the cross-base rows is all `~` except `Copy` at -1.6%, i.e. an improvement, and the spreads run to ±77% — this sandbox sat at load average ~24 on 4 cores throughout, from other tenants. **CI's benchmark job is the authoritative gate here**: it pins `GOMAXPROCS`, measures both arms in-job on the bench runner, and applies #443's resolution check. The allocation numbers above are load-independent and exact, and they are the ones this change is actually about.

## Gates

All run on this branch at `1c13c76` + these commits.

| gate | result |
|---|---|
| `go build ./...` | clean |
| `go vet ./...` | clean |
| `go test -count=1 ./...` | ok, exit 0, 37 packages |
| `go test -race -count=1 ./...` | ok, exit 0, no `DATA RACE` |
| `elps fmt -l ./...` | no output, exit 0 (binary deleted before commit) |
| `elps doc -m` | `All functions and exports are documented.` |
| `make go-test` | exit 0 |
| `make race` | exit 0 |
| `make test-elpscheck` | exit 0 — the singleton integrity checks are also the strongest evidence for the `*token.Location` row above |
| `make test-examples` | exit 0 |
| `./scripts/ci-gates-test.sh` | **233 passed, 0 failed** (see note) |
| `./scripts/confidentiality-guard.sh` | `confidentiality guard: clean` |
| `./scripts/fuzz-budget-check.sh` | 30 targets / 8 shards, required 50 min, **headroom 10 min** — unchanged |

`FuzzPathEngine`, 5 minutes on this branch: **1,508,405 execs**, 546 new interesting inputs (corpus 72 → 618), `PASS`, no new crashers in `testdata/`.

### A note on `ci-gates-test.sh`, and a defect it exposed

The **first** run of it on this branch reported `226 passed, 7 failed` — all in fuzz target *discovery*, naming targets in packages this PR does not touch (`FuzzEval`, `FuzzLexer`, `FuzzParseProgram`, `FuzzLoadJSON`, `FuzzMinifySource`, `FuzzFormat`), with `FuzzPathEngine` unaffected. A re-run on the identical tree gave `233 passed, 0 failed`, matching `main`. The first run was on a cold build cache; the failures are transient and environmental, not caused by this diff.

What is **not** transient is that they were silent. `scripts/fuzz.sh`:

```sh
list_targets() {
	go test ${tagflags+"${tagflags[@]}"} -list '^Fuzz' "$1" 2>/dev/null | grep -E '^Fuzz[A-Za-z0-9_]*$'
}
```

stderr goes to `/dev/null` and the exit status is discarded, so a package that fails to build is indistinguishable from a package with no fuzz targets. Demonstrated **running**, in a throwaway module:

```
--- ./good     targets discovered: [FuzzGoodOne FuzzGoodTwo]
--- ./broken   targets discovered: [<none>]
broken pkg exit status: 1        # discarded by list_targets
```

`fuzz.sh` guards the *total-zero* case ("a fuzz gate that runs nothing cannot fail — treating as an error") but nothing guards *partial* loss, which is the same failure mode one notch quieter: I observed 8 of 30 targets vanish behind a green-looking discovery. Filed separately with this reproducer rather than fixed here — it is in `scripts/`, outside this package's scope.

### DIFF
--- a/lisp/lisplib/libelpspath/libelpspath_test.lisp
+++ b/lisp/lisplib/libelpspath/libelpspath_test.lisp
@@ -145,3 +145,88 @@
   (append! view 99)
   (assert-equal (vector 1 2 3 99) view)
   (assert-equal (vector 1 2 3 4 5) src))
+
+;;; ---- issue #471: a delete through a view must not touch the source ----
+;;;
+;;; A view is an ordinary array LVal whose cells are a window onto a longer
+;;; sequence, so the mutating builtins accept one and cannot tell.  The two
+;;; deleteMutate paths used to compact IN PLACE, shifting the tail left
+;;; through the aliased source's own backing array.  The view's answer came
+;;; out right -- a left shift copies before it overwrites -- and only the
+;;; source was wrecked, which is why nothing caught it.
+;;;
+;;; The source cannot shrink, so there is no "correct" amount for it to
+;;; change: the requirement is that it does not change at all.
+
+(test-let* "?del! index through a kernel slice view leaves the source alone"
+  ((src (vector 1 2 3 4 5))
+   (view (slice 'vector src 0 3)))
+  (elpspath:?del! view 0)
+  (assert-equal (vector 2 3) view)
+  (assert-equal (vector 1 2 3 4 5) src))
+
+(test-let* "?del! range through a kernel slice view leaves the source alone"
+  ((src (vector 1 2 3 4 5))
+   (view (slice 'vector src 0 3)))
+  (elpspath:?del! view '(range 0 1))
+  (assert-equal (vector 2 3) view)
+  (assert-equal (vector 1 2 3 4 5) src))
+
+; The other producer of a view in the tree is this package's own range Get,
+; and it reaches the same defect by a route that never mentions `slice`.
+(test-let* "?del! through elpspath's own range view leaves the source alone"
+  ((src (vector 1 2 3 4 5))
+   (view (elpspath:? src '(range 0 3))))
+  (elpspath:?del! view 0)
+  (assert-equal (vector 2 3) view)
+  (assert-equal (vector 1 2 3 4 5) src))
+
+; A view does not have to be anchored at 0, and the shift landed wherever the
+; window did: this arm reported (vector 1 2 4 5 5) before the fix.
+(test-let* "?del! through an offset view leaves the source alone"
+  ((src (vector 1 2 3 4 5))
+   (view (slice 'vector src 2 5)))
+  (elpspath:?del! view 0)
+  (assert-equal (vector 4 5) view)
+  (assert-equal (vector 1 2 3 4 5) src))
+
+; A full-length view aliases just as completely as a partial one; the window
+; being the whole sequence is not the same as owning it.  Reported
+; (vector 2 3 4 5 5) before the fix.
+(test-let* "?del! through a full-length view leaves the source alone"
+  ((src (vector 1 2 3 4 5))
+   (view (slice 'vector src 0 5)))
+  (elpspath:?del! view 0)
+  (assert-equal (vector 2 3 4 5) view)
+  (assert-equal (vector 1 2 3 4 5) src))
+
+; A view reached through a document, rather than bound to a variable the
+; caller thinks of as a view.
+(test-let* "?del! through a view stored in a document leaves the source alone"
+  ((src (vector 1 2 3 4 5))
+   (doc (sorted-map "v" (slice 'vector src 0 3))))
+  (elpspath:?del! doc "v" 0)
+  (assert-equal (vector 2 3) (elpspath:? doc "v"))
+  (assert-equal (vector 1 2 3 4 5) src))
+
+; The controls that separate #471 from what is expected and from what was
+; already fixed.  These are GUARDS: they pass both before and after the fix,
+; and exist so the semantics around it are re-decided rather than drifted.
+;
+; Assigning an element through a view is ordinary aliasing and is documented:
+; setMutate does cells[index] = newIn, and a view shares its elements.
+(test-let* "?set! at an index through a view does reach the source"
+  ((src (vector 1 2 3 4 5))
+   (view (slice 'vector src 0 3)))
+  (elpspath:?set! view 0 97)
+  (assert-equal (vector 97 2 3) view)
+  (assert-equal (vector 97 2 3 4 5) src))
+
+; The range splice was the same defect and was fixed earlier; it builds its
+; result in a slice it allocates.
+(test-let* "?set! range splice through a view does not reach the source"
+  ((src (vector 1 2 3 4 5))
+   (view (slice 'vector src 0 3)))
+  (elpspath:?set! view '(range 0 1) (vector 90 91))
+  (assert-equal (vector 90 91 2 3) view)
+  (assert-equal (vector 1 2 3 4 5) src))
--- a/lisp/lisplib/libelpspath/path.go
+++ b/lisp/lisplib/libelpspath/path.go
@@ -893,11 +893,39 @@ func (s *indexPath) deleteMutate(in *lisp.LVal) (*lisp.LVal, error) {
 	if err != nil {
 		return nil, err
 	}
-	index, ok := resolveIndex(len(cells), s.index)
+	n := len(cells)
+	index, ok := resolveIndex(n, s.index)
 	if !ok {
 		return lisp.Nil(), nil
 	}
-	vals := append(cells[:index], cells[index+1:]...)
+	// IMPORTANT: the compaction is built in a slice this function allocates,
+	// not by shifting cells down inside their own backing array.
+	//
+	// `append(cells[:index], cells[index+1:]...)` writes the tail one position
+	// to the left THROUGH cells' array. That array does not belong to this
+	// function whenever in is a VIEW over a longer sequence -- and the
+	// mutating builtins accept one, because a view is an ordinary array LVal
+	// (the kernel's (slice 'vector ...), or this package's own rangePath.Get).
+	// The shift then lands in the aliased source, which cannot shrink, so it
+	// is left scrambled rather than shortened:
+	//
+	//	(set 'v (vector 1 2 3 4 5))
+	//	(set 'w (slice 'vector v 0 3))
+	//	(?del! w 0)
+	//	  view (vector 2 3)   src (vector 2 3 3 4 5)   <- 1 gone, 3 duplicated
+	//
+	// Nothing raised, and the ANSWER was right: a left shift copies before it
+	// overwrites, so the view came out correct while the source was wrecked.
+	// That is why neither the suite nor the fuzzer caught it -- see issue
+	// #471, and rangePath.setMutate, which carries this same comment because
+	// its overlap corrupted its own answer and so was found immediately.
+	//
+	// This is the same class as #369/#373 but no capacity clamp reaches it:
+	// the write is WITHIN len, not past it. The capacity below is exact, so
+	// the cost is one allocation on a path that previously did none.
+	vals := make([]*lisp.LVal, 0, n-1)
+	vals = append(vals, cells[:index]...)
+	vals = append(vals, cells[index+1:]...)
 	storeCells(in, vals)
 	return in, nil
 }
@@ -1070,10 +1098,23 @@ func (s *rangePath) deleteMutate(in *lisp.LVal) (*lisp.LVal, error) {
 	if err != nil {
 		return nil, err
 	}
-	vals := cells[:from]
-	if to < n {
-		vals = append(vals, cells[to:]...)
-	}
+	// IMPORTANT: allocated rather than compacted in place, for exactly the
+	// reason indexPath.deleteMutate above spells out -- `append(cells[:from],
+	// cells[to:]...)` shifts the tail left through a backing array this
+	// function does not own when in is a view, scrambling the source it
+	// aliases (issue #471). Both delete paths had it; the range one reaches it
+	// through '(range a b) rather than an integer step:
+	//
+	//	(?del! w '(range 0 1))   ->   src (vector 2 3 3 4 5)
+	//
+	// The old code skipped the append entirely when to == n, which is why a
+	// delete of a suffix -- including the whole view -- was accidentally
+	// correct: there was no tail to shift. Only a delete with a non-empty tail
+	// wrote through, which narrowed the shapes that could expose it without
+	// making any of them safe.
+	vals := make([]*lisp.LVal, 0, n-(to-from))
+	vals = append(vals, cells[:from]...)
+	vals = append(vals, cells[to:]...)
 	storeCells(in, vals)
 	return in, nil
 }
--- a/lisp/lisplib/libelpspath/path_bench_delete_test.go
+++ b/lisp/lisplib/libelpspath/path_bench_delete_test.go
@@ -1,0 +1,91 @@
+// Copyright © 2026 The ELPS authors
+
+package libelpspath
+
+import (
+	"testing"
+
+	"github.com/luthersystems/elps/lisp"
+)
+
+// BenchmarkDeleteCompaction measures what the #471 fix costs.
+//
+// The fix replaced an in-place compaction — which allocated nothing, because
+// it shifted the tail down inside the caller's own array — with one that
+// builds the result in a slice of its own.  That is one allocation on a path
+// that previously did none, so it is measured rather than asserted to be
+// free.  Issue #452's rule applies: allocations are not exempt from the gate,
+// and a clamp or a copy that forces one has to show the number.
+//
+// The two arms alternate inside a single process, as BenchmarkCycleGuardCost
+// does, so machine drift moves both together and benchstat -col /arm reads
+// the difference rather than the noise.  The "inplace" arm is the pre-fix
+// body, kept alive as inPlaceViewDelete in path_view_alias_test.go and pinned
+// to the shipped answer by TestInPlaceReplicaMatchesTheShippedAnswer — so the
+// comparison is against what actually used to run, not a sketch of it.
+//
+// Both arms are given a fresh sequence per iteration, since both consume the
+// one they are handed.  The allocation that setup costs is charged to both
+// equally and is what the /arm comparison cancels.
+func BenchmarkDeleteCompaction(b *testing.B) {
+	steps := []struct {
+		name string
+		step *lisp.LVal
+	}{
+		{"index", lisp.Int(0)},
+		{"range1", rangeStep(0, 1)},
+		{"range3", rangeStep(0, 3)},
+	}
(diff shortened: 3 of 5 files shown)


sam-at-luther:
## CI's benchmark numbers, which are worse than the ones in the body

The body quotes my sandbox measurements. CI measured both arms on the pinned bench runner, interleaved, n=10 each, and its `DeleteCompaction` numbers are **materially worse than mine**. Correcting the record rather than leaving the flattering number standing.

### The gate verdict

```
benchstat-gate: interpreted 20 delta row(s) + 404 no-change row(s);
1 significant move(s) in the bad direction; 0 at or above the gate
(timing 15%, allocation 5%).
```

`Required: benchmark` — **success**, 0 rows at or above the gate. The single below-gate move is `libelpspath Copy-2 +4.79% (p=0.043, gate 15%)`, which is `copyLVal` and does not touch either changed function.

Cross-base allocation, which is what this change is actually about:

```
info  libelpspath B/op      geomean  delta=+0.00%
info  libelpspath allocs/op geomean  delta=+0.00%
```

That matches my local run, and for the reason given in the body: no benchmark on `main` passes an integer or range step, so none of them reaches the changed code.

### What the isolated step actually costs

`BenchmarkDeleteCompaction` is new, so it has no base column and is reported rather than gated. On the bench runner (arm64):

| n | step | inplace | allocating | delta |
|---|---|---|---|---|
| 4 | index | 317.8n ± 2% | 356.6n ± 6% | **+12.2%** |
| 4 | range1 | 333.4n ± 1% | 358.4n ± 4% | +7.5% |
| 4 | range3 | 329.8n ± 4% | 352.9n ± 3% | +7.0% |
| 16 | index | 318.3n ± 3% | 363.2n ± 3% | +14.1% |
| 16 | range1 | 335.9n ± 4% | 382.4n ± 6% | +13.8% |
| 16 | range3 | 331.6n ± 6% | 387.0n ± 4% | +16.7% |
| 64 | index | 335.4n ± 3% | 459.2n ± 3% | **+36.9%** |
| 64 | range1 | 349.1n ± 9% | 473.8n ± 3% | +35.7% |
| 64 | range3 | 352.8n ± 7% | 468.9n ± 4% | +32.9% |

My sandbox reported geomean +7.1% and a worst row of +12.5%. **The real figure is +7% to +37%, and it grows with sequence length** — my box was too loaded to resolve the n=64 rows, and I under-reported. The scaling is expected once stated plainly: the old code shifted in place with no allocation, the new one allocates an n-element slice and copies into it, so the gap widens as n does.

### Why this is still the right trade, stated explicitly rather than assumed

- It is not a gate row and does not regress anything that exists on `main`.
- The absolute cost is ~124ns on a 64-element delete.
- The **copying** `?del` — the non-mutating one, and the common one on a transaction path — is unaffected: it already allocates a whole private copy in `copySeqOffPath`, which is why the cross-base `BuiltinDelCopyDeep` rows are `+0.00%` on both allocation metrics at every depth. The cost lands only on in-place `?del!` over a long sequence.
- The alternative is a silent write into a sequence the caller never passed.

If the in-place delete on long sequences turns out to be hot for someone, the way to buy the allocation back is to detect that the operation owns its backing array outright and shift in place only then — but that needs an ownership signal an `LVal` does not currently carry, which is a larger change than #471 should make.


---

PR #483, opened 2026-08-16
fuzzwatch: write down the stall-vs-share boundary, and guard the floor

### LINKED ISSUE #453: fuzzwatch measures scheduler stall, not CPU share — a scheduled-time budget does not detect a starved-of-CPU process
Found by **running**, while fixing #435.

## What I measured

`internal/fuzzwatch` exists so a descheduled process is not blamed for a hang, and its package doc describes what it measures precisely and correctly: "how much wall clock passed during which it did NOT run on schedule". A heartbeat goroutine ticks every 100ms and charges any gap over `4 × tick` (400ms) as lost time.

That is a measure of **scheduling latency for a mostly-idle goroutine**. It is not a measure of **CPU share**, and under Linux CFS the two come apart badly. A long run queue gives a waking, near-idle task good latency even when its share of the CPU is tiny.

Directly measured on the 4-core sandbox, with 200 competing spinners (load average ~250), a `Budget` created and then slept against for 3s:

```
wall=3.003386989s  scheduled=3.003386989s  lost=0s  longest=0s
verdict=hung ... 0s lost to scheduler stall (longest single stall 0s)
```

**Zero lost time.** In the same conditions, evaluation of a fixed seed slowed by roughly 50x (35ms of CPU work failing to complete inside a 2s wall-clock budget). fuzzwatch would have reported that process as running normally throughout.

## Why it matters, and why it mostly does not

It does **not** invalidate the nine watchdogs using it today. They are sized at 20,000x–90,000x the mean work (30s against 0.33ms, 15s against 0.73ms, 20s against 0.91ms), so a 50x slowdown still leaves three orders of magnitude of headroom. In its actual role — "is this call hung?" — fuzzwatch is fine, and against the failure mode it was written for (a process genuinely frozen: VM steal, cgroup throttling with long periods, an SMR pause) it works as designed.

It matters because the package doc reads as a general remedy for "this machine is not running us", and the first suggested fix in #435 was to give a **2s** budget the same treatment. At 57x headroom rather than 90,000x that does not work, and I would have shipped it if I had not measured. PR #447 documents this at the call site so the next person does not repeat the experiment.

## Suggested directions, in rough order

1. **Document the limit in the package doc.** One paragraph distinguishing scheduler *stall* from CPU *share*, and stating the headroom regime the instrument is honest in. Cheapest, and probably sufficient.
2. **Record the calibration.** The existing doc lists measured mean work against each watchdog; adding "and this instrument resolves stalls above 400ms, not slowdowns" makes the sizing argument checkable.
3. **Measure share as well as stall,** if a tighter budget is ever wanted: `/proc/self/stat` utime+stime deltas against wall clock. This over-counts when several test goroutines are busy (parallel subtests), so it is not a drop-in, and I would not build it without a concrete need.

## For whoever picks this up

The probe is three lines against `fuzzwatch.New` plus a sleep; the load was `for i in $(seq 1 200); do ( while :; do :; done ) & done` on 4 cores. Reproducing it should take five minutes. If the conclusion is "documented headroom is sufficient, no code change", closing with that reasoning recorded is a good outcome — the point is that the boundary is written down somewhere before someone sizes a budget against it.

sam-at-luther:
Fixes #453

## Verdict: the instrument is right, the documentation was not

#453 reproduces, and more sharply than it was filed. **No behaviour change to `fuzzwatch`.** It is correct for what it claims to detect — a process genuinely frozen, where gaps land well above the 400ms resolution floor. It is simply not, and cannot be, a load-aware budget. What was missing is a written boundary and something that stops the next person crossing it.

## Reproduced

On the 4-core sandbox with 200 spinners (`for i in $(seq 1 200); do ( while :; do :; done ) & done`, load average > 50):

**Probe A — the issue's probe, a `Budget` and a 3s sleep:**

```
wall=3.003s scheduled=3.003s lost=0s longest=0s
verdict=hung starved=false
```

Zero lost time, under load heavy enough to make the machine unusable. Confirms the filing exactly.

**Probe B — the same budget over CPU-bound work**, which is what a fuzz target actually does:

| run | work (169ms unloaded) | CPU share | fuzzwatch reported |
|---|---|---|---|
| 1 | 17.375s (103x slower) | 0.9% | `scheduled=17.357s lost=300ms` |
| 2 | 12.279s (73x slower) | 1.2% | `scheduled=12.66s lost=0s` |

The second run is the sharper one: a process that received 0.15s of CPU across 12.66s of wall clock was reported as having run normally throughout. Not under-counted — **not counted at all**.

## Why, structurally

Two independent reasons, neither a bug:

1. A stall is only visible above `tolerance * tick` = **400ms**. Starvation on a busy machine is not one long freeze but a few microseconds of CPU handed out every millisecond, forever. Every one of those gaps is below the floor.
2. The heartbeat goroutine is, by construction, almost entirely idle — and CFS gives a waking, near-idle task excellent latency no matter how long the run queue is. The probe is precisely the kind of task starvation does not touch.

So "scheduled time" is a truthful name only against the failure mode the package was built for (VM steal, cgroup throttling, an SMR pause). Against contention it degenerates to the wall clock, silently.

## Why the eleven live watchdogs are fine

They are sized 20,000x–90,000x above the mean work they bound — FuzzEval 0.33ms against 30s, FuzzApplyStdlib 0.73ms against 15s, FuzzSchemaValidate 0.91ms against 20s. A 100x slowdown still leaves two to three orders of magnitude spare, so whether the instrument separates stall from share never arises.

That headroom, **not the accounting**, is what makes them sound. That is the thing worth writing down, because it is exactly the assumption a tighter budget would quietly drop.

## What this PR adds

- **Package doc**: a "What this does NOT measure: CPU share" section carrying the measurements above, the mechanism, and the regime in which the instrument is honest.
- **`MinHonestBudget = 10s`**: names the floor. Just under the smallest live budget (15s, `lisp/cycle_fuzz_test.go` and `lisp/lisplib/fuzz_test.go`) and far above the 2s that #435 proposed. A backstop against a new call site, not a target to design against.
- **`TestEveryBudgetIsAboveTheHonestFloor`**: AST-scans every `fuzzwatch.New` in the repository — 11 sites, all evaluated, none skipped — and fails below the floor. A source scan rather than a check inside `New`, because `New` runs inside `f.Fuzz` bodies where a panic would be recorded as a crasher and blamed on the input. A budget expression the evaluator cannot read fails asking to be taught, rather than being silently skipped.
- **`TestTheFloorGuardCanFail`**: its negative control.
- **`TestCPUShareProbe`**: the reproduction harness, skipped unless `FUZZWATCH_SHARE_PROBE=1`, since it needs external load to mean anything. It asserts nothing and only prints, so the boundary can be re-derived rather than believed.

## Explicitly rejected

**Direction 3 in the issue — measuring share via `/proc/self/stat` utime+stime — is not implemented.** PR #447 (#435) already measured that a fuzzwatch-based scheduled-time budget does not work and rejected it; that measurement is the origin of this issue and is not re-proposed. A share-based instrument additionally over-counts whenever several goroutines of the process are legitimately busy (parallel subtests), so it is not a drop-in, and there is no call site that needs it: every live budget clears the floor by 50x or more. The doc records the option and its caveat for whoever arrives with a concrete need.

**No threshold was loosened and no watchdog was retuned.** Nothing here changes when any target fires.

## Red, then green

`TestEveryBudgetIsAboveTheHonestFloor` is a **guard** — it passes on `main` and is labelled as such in-file. Its ability to fire is demonstrated two ways.

Against a real call site, with `cycleWatchdog` temporarily set to the 2s #435 proposed:

```
--- FAIL: TestEveryBudgetIsAboveTheHonestFloor (0.04s)
    lisp/cycle_fuzz_test.go:86: fuzzwatch budget is 2s, below MinHonestBudget (10s).
        fuzzwatch resolves scheduler STALL above 400ms, not CPU SHARE: measured on 4 cores
        under 200 spinners it reported lost=0s while real work ran 103x slower at 0.9% share.
        A budget this tight cannot tell 'the machine was busy' from 'the code hung', so it will
        fire on innocent inputs under load. PR #447 measured exactly this for a 2s budget and
        rejected it; see #453 and the fuzzwatch package doc.
```

Restored, it reports `checked 11 fuzzwatch budget(s) across 11 file(s) against a 10s floor` and passes. `TestTheFloorGuardCanFail` runs the same evaluator over a synthetic file on every CI run, so the control does not depend on anyone remembering to repeat the experiment.

## Gates

`go build`, `go vet`, `go test -count=1`, `go test -race`, `golangci-lint run` (0 issues), `elps fmt -l`, `elps doc -m`, `make go-test`, `make race`, `make test-elpscheck`, `make test-examples`, `./scripts/ci-gates-test.sh` (205/0), `./scripts/confidentiality-guard.sh`, `./scripts/fuzz-budget-check.sh` (unchanged, 10 min headroom). No workflow or `scripts/` files touched.

---

### DIFF
--- a/internal/fuzzwatch/callsite_guard_test.go
+++ b/internal/fuzzwatch/callsite_guard_test.go
@@ -1,0 +1,383 @@
+// Copyright © 2026 The ELPS authors
+
+package fuzzwatch
+
+import (
+	"fmt"
+	"go/ast"
+	"go/parser"
+	"go/token"
+	"io/fs"
+	"os"
+	"path/filepath"
+	"slices"
+	"strconv"
+	"strings"
+	"testing"
+	"time"
+)
+
+// GUARD, not a catch. Every call site in the tree already clears the floor by
+// 50x or more, so this test passes on main and always has. It is here to stop
+// a budget arriving BELOW the floor later, which is the specific mistake #453
+// exists to prevent and the one PR #447 had to measure its way out of.
+//
+// It is a source scan rather than a check inside New because New is called
+// from inside f.Fuzz bodies: a panic there would be recorded as a crasher and
+// attributed to the input, which is a worse failure than the one being
+// prevented. A test failure names the file and the number and says why.
+//
+// The evaluator below is deliberately small. It understands the shapes the
+// repository actually uses -- a duration constant, a sum of them, a literal
+// multiple of time.Second -- and REFUSES anything else rather than guessing.
+// An unrecognised shape fails the test asking for the evaluator to be taught,
+// because silently skipping call sites it cannot read is exactly how a guard
+// like this rots into one that cannot fail.
+
+// budgetSite is one evaluated fuzzwatch.New call.
+type budgetSite struct {
+	Pos    token.Position
+	Budget time.Duration
+}
+
+// budgetsIn returns every fuzzwatch.New budget in f that could be evaluated,
+// plus a description of each one that could not. Shared by the repository scan
+// and by the negative control below, so the control exercises the same code
+// path the real guard runs on.
+func budgetsIn(fset *token.FileSet, f *ast.File) (sites []budgetSite, unreadable []string) {
+	consts := durationConsts(f)
+	ast.Inspect(f, func(n ast.Node) bool {
+		call, ok := n.(*ast.CallExpr)
+		if !ok {
+			return true
+		}
+		sel, ok := call.Fun.(*ast.SelectorExpr)
+		if !ok || sel.Sel.Name != "New" {
+			return true
+		}
+		if pkg, ok := sel.X.(*ast.Ident); !ok || pkg.Name != "fuzzwatch" {
+			return true
+		}
+		if len(call.Args) != 1 {
+			return true
+		}
+		pos := fset.Position(call.Pos())
+		d, ok := evalDuration(call.Args[0], consts)
+		if !ok {
+			unreadable = append(unreadable,
+				fmt.Sprintf("%s:%d: %s", pos.Filename, pos.Line, exprString(call.Args[0])))
+			return true
+		}
+		sites = append(sites, budgetSite{Pos: pos, Budget: d})
+		return true
+	})
+	return sites, unreadable
+}
+
+const floorDiagnosis = "fuzzwatch resolves scheduler STALL above %s, not CPU SHARE: measured on 4 cores\n" +
+	"    under 200 spinners it reported lost=0s while real work ran 103x slower at 0.9%% share.\n" +
+	"    A budget this tight cannot tell 'the machine was busy' from 'the code hung', so it will\n" +
+	"    fire on innocent inputs under load. PR #447 measured exactly this for a 2s budget and\n" +
+	"    rejected it; see #453 and the fuzzwatch package doc."
+
+func TestEveryBudgetIsAboveTheHonestFloor(t *testing.T) {
+	root := repoRoot(t)
+
+	var files []string
+	err := filepath.WalkDir(root, func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			return err
+		}
+		if d.IsDir() {
+			switch d.Name() {
+			case ".git", "testdata", "node_modules", "_examples":
+				return fs.SkipDir
+			}
+			return nil
+		}
+		if !strings.HasSuffix(path, ".go") {
+			return nil
+		}
+		// The probe under internal/fuzzwatch/probe is a manual reproduction
+		// harness for the #453 measurement; it deliberately uses a short
+		// budget because measuring the instrument is its whole purpose.
+		if strings.Contains(filepath.ToSlash(path), "/internal/fuzzwatch/") {
+			return nil
+		}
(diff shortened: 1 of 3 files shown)


sam-at-luther:
Follow-up found while auditing this neighbourhood, filed separately rather than folded in here: **#488** — `Budget.Check` returns `Hung` for a window it simultaneously classifies as `Starved()`, because the `Scheduled() >= total` arm is tested first and never consults `Starved()`. `Inconclusive` is therefore unreachable whenever the process gets more than ~25% of the CPU.

It is complementary to #453 rather than a duplicate. #453 (this PR) is about stalls fuzzwatch **cannot see** — CPU share, and anything under the 400ms floor. #488 is about a stall it **did** see, reported in full (`54s lost, longest single stall 10.7s, 36% of wall scheduled`), and then discarded when reaching a verdict. It cost a live failure in `lisp/x/debugger` under 400 burners: 4/4 loaded, 0/4 idle.

I have kept this PR to documentation plus the call-site floor, as the issue and its brief asked, because #488 is a behaviour change with a real semantic trade-off (some genuine hangs on a busy machine would become `Inconclusive`) and deserves its own decision and its own red-then-green.


---

PR #487, opened 2026-08-16
parser: keep the lexer watchdog from outranking the real check

### LINKED ISSUE #454: Two parser termination watchdogs still use a bare wall-clock time.After instead of fuzzwatch
Found by **reading**, during the neighbourhood audit for #435. **Not reproduced** — please reproduce-or-close.

## What

Nine termination watchdogs in this repository are denominated in scheduled time via `internal/fuzzwatch`, which exists precisely so that wall clock during which the process was not run by the OS is not charged to the code under test:

`lisp/eval_fuzz_test.go`, `lisp/cycle_fuzz_test.go`, `lisp/lisplib/fuzz_test.go`, `lisp/lisplib/libschema/fuzz_test.go`, `lisp/x/debugger/debug_fuzz_test.go`, `lsp/lsp_fuzz_test.go`, `lint/lint_fuzz_test.go`, `elpsutil/elpsutil_fuzz_test.go`, `analysis/analysis_fuzz_test.go`, `analysis/workspace_fuzz_test.go`, `mcpserver/tool_agreement_fuzz_test.go`.

Two are not:

- `parser/lexer/fuzz_test.go:115` — `case <-time.After(30 * time.Second): t.Fatal("lexer did not terminate")`
- `parser/rdparser/fuzz_test.go:217` — `case <-time.After(30 * time.Second): t.Fatalf("%s did not terminate on %q", mode, name)`

Both assert that a call terminates, both fail the build when the deadline fires, and both measure it with the wall clock — the exact shape `internal/fuzzwatch`'s package doc opens by describing as the problem it was written to solve.

## Why it is probably fine today, and why it is still worth a line

The headroom is enormous. These targets lex and parse small fixed inputs; the work is microseconds against a 30s bound. Reaching it needs starvation far beyond anything I produced. So the honest claim is **inconsistency with the house pattern**, not a live flake — I have no reproduction, and I did not try to manufacture one.

It is worth recording because the value of a house pattern is that the next person does not have to re-derive it. Someone reading `parser/rdparser/fuzz_test.go` for a template will copy `time.After`, and the note in `fuzzwatch`'s doc that "three fuzz targets in this repository assert that a call terminates, using the same shape" is now describing thirteen, two of which were never converted.

Note also #453: fuzzwatch resolves scheduler *stall*, not CPU *share*, so converting these buys less than it appears to. At 30s against microseconds of work either instrument has ample margin; the argument for converting is uniformity and the removal of a copyable anti-pattern, not a measured defect.

## Suggested resolution

Either convert both to the `fuzzwatch.New` / `Budget.Check` loop the other nine use (mechanical, ~15 lines each, and the `Inconclusive` → `t.Skipf` arm comes with it), or add a one-line comment at each site saying the wall clock is deliberate here and why. Both are fine; silently having two of each is not.

I deliberately did not widen PR #447 to cover them — that PR is about a *correctness* assertion enforced by a clock, and these are liveness watchdogs with four orders of magnitude of margin.

sam-at-luther:
Fixes #454

#454 was filed from reading and asked for reproduce-or-close. **The filed concern does not hold — the two `time.After` uses are correct in context and are not converted.** Reproducing it did surface a real defect at one of the two sites, which this PR fixes.

## Why the two are not converted

The issue frames them as two of thirteen watchdogs of one shape, two never converted. They are not the same shape:

| | the eleven `fuzzwatch` sites | the two `time.After` sites |
|---|---|---|
| context | inside an `f.Fuzz` body | plain `Test` functions |
| input | mutator-generated, open set | fixed checked-in corpora |
| hang | a live possibility | already known to terminate |

All eleven `fuzzwatch.New` call sites bound mutator-generated input — four via helpers (`runSessionBudgeted`, `debugEvalBudgeted` and friends) called from fuzz bodies. The two cited sites are `TestLexerTerminatesOnUndecodableInput` (eight literal byte strings) and `TestParserSurvivesPathologicalInput` (`fuzzseed.Pathological()`), replaying closed regression corpora. Slowest subtest measured: **0.12s against a 30s bound**.

Converting also buys nothing measurable, which #453 now quantifies: fuzzwatch resolves scheduler *stall* above 400ms, not CPU *share*. Under 200 spinners on 4 cores it reported `lost=0s` while real work ran 73–103x slower at ~1% CPU share. Under the only load that could threaten a 30s bound it reports `scheduled == wall`, so a converted watchdog fires at exactly the moment these do — the same clock, wearing a hat.

Both sites now carry that reasoning in-file, which is the second remedy #454 itself offered ("add a one-line comment at each site saying the wall clock is deliberate here and why").

## What reproducing it found

`TestLexerTerminatesOnUndecodableInput` ran `lexAll(t, ...)` **on a worker goroutine**, and `lexAll` called `t.Fatal` in four places:

```go
done := make(chan []*token.Token, 1)
go func() {
    done <- lexAll(t, []byte(src))   // lexAll calls t.Fatal internally
}()
```

`testing.T.FailNow` — which `t.Fatal` calls — is documented as usable only from the goroutine running the test. From anywhere else it runs `runtime.Goexit` on the caller and the test itself keeps going. So the worker died without ever sending on its channel, the 30s watchdog ran to completion, and a genuine `TokenStream` contract violation was reported as **"lexer did not terminate"** — for a lexer that had terminated perfectly well.

The watchdog outranked the check it was meant to backstop. `lexAll`'s token-per-byte limit is the *real* termination check here: deterministic, load-independent, and it names the defect. The clock is only a backstop for a lexer that blocks rather than loops.

### Red, then green

With a contract violation injected into `lexAll` (`return`/`t.Fatal` on non-empty input), running all eight subtests:

**Before — on `main`:**
```
8 subtests reporting "lexer did not terminate"
real    4m1.439s
```

Each subtest waits out its own 30s watchdog and prints the wrong diagnosis.

**After — this branch:**
```
--- FAIL: TestLexerTerminatesOnUndecodableInput (0.00s)
    --- FAIL: TestLexerTerminatesOnUndecodableInput/abc\xff (0.00s)
        fuzz_test.go:140: INJECTED: pretend the TokenStream contract was violated
    --- FAIL: TestLexerTerminatesOnUndecodableInput/\xff (0.00s)
        fuzz_test.go:140: INJECTED: pretend the TokenStream contract was violated
    ...
real    0m0.708s
```

**340x faster, and the diagnosis is the actual violation.**

## The change

- `lexAll` returns `([]*token.Token, error)` instead of calling `t.Fatal`; the verdict is taken on the test goroutine.
- `FuzzLexer` already called it correctly (on the test goroutine) and is updated for the new signature.
- Both watchdog sites gain the deliberate-wall-clock rationale.
- The rdparser site has no equivalent problem — `parse()` takes no `*testing.T` and its goroutine only closes a channel — and is otherwise unchanged.

## Neighbourhood audit

I scanned every `*_test.go` in the repository for `t.Fatal`-family calls reachable from a `go func()` body, directly or through a `*testing.T`-taking helper. The scanner finds **exactly this one site** on `main` and none after the fix (its one remaining report is a false positive: a same-named `lexAll` in `package lexer`'s `literal_test.go`, which is not called from a goroutine).

Two further things turned up under load and are **filed separately rather than folded in here** — see the issues linked below. Neither is touched by this PR.

## Gates

`go build`, `go vet`, `go test -count=1`, `go test -race`, `golangci-lint run` (0 issues), `elps fmt -l`, `elps doc -m`, `make go-test`, `make race`, `make test-elpscheck`, `make test-examples`, `./scripts/ci-gates-test.sh` (205/0), `./scripts/confidentiality-guard.sh`, `./scripts/fuzz-budget-check.sh` (unchanged). No workflow or `scripts/` files touched.

---

### DIFF
--- a/parser/lexer/fuzz_test.go
+++ b/parser/lexer/fuzz_test.go
@@ -4,6 +4,8 @@ package lexer_test
 
 import (
 	"bytes"
+	"errors"
+	"fmt"
 	"testing"
 	"time"
 
@@ -33,30 +35,46 @@ import (
 // multi-token emission (emitMacroChar) can never trip it.
 const maxTokenSlack = 8
 
-// lexAll drains the lexer and returns the tokens read.  It fails the test if
+// lexAll drains the lexer and returns the tokens read.  It reports an error if
 // the lexer emits more tokens than the input can justify, which is how a
 // non-consuming loop shows up.
-func lexAll(t *testing.T, src []byte) []*token.Token {
-	t.Helper()
+//
+// It returns an error rather than calling t.Fatal, because
+// TestLexerTerminatesOnUndecodableInput runs it on a WORKER GOROUTINE.
+// testing.T.FailNow -- which t.Fatal calls -- is documented as usable only
+// from the goroutine running the test: from anywhere else it runs
+// runtime.Goexit on the caller and the test itself keeps going.  Here that had
+// teeth.  The worker died without ever sending on its channel, so the 30s
+// watchdog below ran to completion and reported "lexer did not terminate" for
+// a lexer that had terminated perfectly well -- 30 seconds to print the wrong
+// diagnosis, for each of the eight inputs.  Returning the error puts the
+// verdict back on the test goroutine, where a genuine contract violation
+// fails in microseconds and says what it actually was.
+//
+// Note that this bound -- tokens emitted against bytes consumed -- is the real
+// termination check for the lexer, not the clock.  It is deterministic, it
+// cannot be perturbed by machine load, and it names the defect.  The watchdog
+// downstream is only a backstop for a wedge this cannot see.
+func lexAll(src []byte) ([]*token.Token, error) {
 	lx := lexer.New(token.NewScanner("fuzz", bytes.NewReader(src)))
 	limit := len(src) + maxTokenSlack
 	var out []*token.Token
 	for {
 		toks := lx.ReadToken()
 		if len(toks) == 0 {
-			t.Fatal("ReadToken returned an empty slice, violating the TokenStream contract")
+			return out, errors.New("ReadToken returned an empty slice, violating the TokenStream contract")
 		}
 		for _, tok := range toks {
 			if tok == nil {
-				t.Fatal("ReadToken returned a nil token")
+				return out, errors.New("ReadToken returned a nil token")
 			}
 			out = append(out, tok)
 			if tok.Type == token.EOF || tok.Type == token.ERROR || tok.Type == token.INVALID {
-				return out
+				return out, nil
 			}
 		}
 		if len(out) > limit {
-			t.Fatalf("lexer emitted %d tokens for %d bytes of input without terminating "+
+			return out, fmt.Errorf("lexer emitted %d tokens for %d bytes of input without terminating "+
 				"(last token %v %q): it is not consuming input",
 				len(out), len(src), out[len(out)-1].Type, out[len(out)-1].Text)
 		}
@@ -70,7 +88,10 @@ func FuzzLexer(f *testing.F) {
 		f.Add(src)
 	}
 	f.Fuzz(func(t *testing.T, src []byte) {
-		toks := lexAll(t, src)
+		toks, err := lexAll(src)
+		if err != nil {
+			t.Fatal(err)
+		}
 		for i, tok := range toks {
 			if tok.Source == nil {
 				t.Fatalf("token %d (%v %q) has no source location", i, tok.Type, tok.Text)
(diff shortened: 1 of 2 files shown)


sam-at-luther:
The two things the neighbourhood audit turned up under load, filed separately rather than folded in here:

- **#489** — the debugger's `require.Eventually(..., 2*time.Second, ...)` breakpoint waits fail on a merely contended machine: 5/5 deterministic failures in `dapserver` at load ~24 on 4 cores (other tenants, no synthetic load), and 4/4 in `engine_test.go` under 400 burners, against 0/4 idle. ~112 wall-clock budgets in that family (84 at 5s, 28 at 2s). Same shape as the lexer fix here — the deterministic signal the engine already emits should decide, with the clock demoted to a backstop — but it is a family-wide decision, not a one-line change.

- **#488** — `fuzzwatch.Check` returns `Hung` for a window it simultaneously classifies as `Starved()`. Surfaced by the same load run, in the same package.

Neither is touched by this PR.


---

PR #496, opened 2026-08-16
lisp: give LVal.Copy's result its own Meta and MacroExpansion

### LINKED ISSUE #466: LVal.Copy shallow-copies Meta and MacroExpansion too, so a "deep copy" shares its formatting metadata and its per-node expansion ID

sam-at-luther:
Fixes #466

## Reproduced

The issue's probe, run against `95e2e1a`:

```
root Meta: orig=0xc000131ab0 cp=0xc000131ab0 same=true
after cp.Meta.BlankLinesBefore=99: orig.Meta.BlankLinesBefore=99 (was 0)
LeadingComments[0] token: orig=0xc00011a1b0 cp=0xc00011a1b0 same=true
LeadingComments[0].Source: orig=0xc0001220a0 cp=0xc0001220a0 same=true
TrailingComment token: orig=0xc00011a300 cp=0xc00011a300 same=true
MacroExpansion: orig=0xc00011c330 cp=0xc00011c330 same=true; ID orig=7 cp=7
MacroExpansionContext: same=true
```

Same as the issue's, plus the two rows it did not print: the comment token's own `*token.Location` is shared, and so is `MacroExpansionContext`.

## The open question: code or doc comments

#466 leaves this open, and #467's author deliberately left these fields alone. The answer splits by pointer, and the evidence is different for each.

**`Meta` — the code.** Three things say so:

1. `SourceMeta` is not a description of a node, it is per-node **mutable** state. `parser/rdparser` writes every field of it in place, and `hoistOperandComments` *moves* `LeadingComments` from one node to another — `append` onto the outer node, `= nil` on the inner one.

2. The parser already treats two LVals sharing one `*SourceMeta` as an anomaly it has to defend against:

   ```go
   if outer.Meta == inner.Meta {
       // Shared Meta: the comments are already on the node the printer
       // writes.  Moving them would move them onto themselves.
       return
   }
   ```

   An in-tree guard against a state means the state is anomalous. `LVal.Copy` manufactured it on purpose, at every node, on every call.

3. Leaving it shared leaves the sentence #467 just added to `Copy` false one level down. "The copy owns its positions. Every node Copy reaches gets its own `*token.Location`" — except the ones on `Meta.LeadingComments[i].Source` and `Meta.TrailingComment.Source`, which both trees reached. Taking the doc option here would mean *weakening* #467's guarantee one release after landing it.

The cost argument the issue raises is confirmed: `Meta` is nil outside format-preserving parsing, so this is a nil check on the hot path. It is not the #421 trade that made `Source` a judgement call.

**`MacroExpansionInfo` — the code; its embedded `*MacroExpansionContext` — legitimately shared.** The context describes the macro **call**, is already documented "shared across all nodes in one expansion", and #456 made its `CallSite` an object the expansion owns rather than one borrowed from a live parse tree. There is no third party to separate it from, so copying it would separate nothing and would make that documented sharing false for copied nodes. The per-node wrapper around it does get separated.

**`MacroExpansionInfo.ID` — the doc comment.** This is the one place the issue's alternative reading is right. `ID` was documented "unique per node, monotonically increasing". `LVal.Copy` cannot honour that under *any* implementation: it takes no `*Runtime`, so it has no counter to draw a fresh value from, and drawing one from anywhere else would defeat the point — the value exists to come from the runtime that did the expanding. So the behaviour stands and the comment now says what is true, and names the consumer that has to know:

```go
// lisp/x/debugger/stepper.go
if loc.MacroID != s.start.MacroID {
```

The stepper steps on `MacroID` *changing*, so two distinct nodes carrying one ID read to it as one node and it will not pause between them. That is what "unique per node" was buying, and a copy does not deliver it.

## Red then green

`lisp/copy_meta_test.go`, 13 tests. Against `95e2e1a` (the fix's two assignment lines reverted, the new `Copy` methods left in so the file compiles):

```
--- FAIL: TestCopyDoesNotAliasSourceMeta
--- FAIL: TestCopyDoesNotAliasMetaCommentTokens
--- FAIL: TestCopyDoesNotAliasMetaAtDepth          18 of 18 copied nodes with metadata share the original's *SourceMeta
--- FAIL: TestCopyMetaSurvivesACommentHoist
--- FAIL: TestCopyMetaCommentSlicesArePrivate
--- FAIL: TestCopyDoesNotAliasMacroExpansionInfo
--- PASS: TestCopyPreservesMetaContent               (GUARD)
--- PASS: TestCopyPreservesNilMeta                   (GUARD)
--- PASS: TestCopyPreservesNilComments               (GUARD)
--- PASS: TestCopyKeepsSharingTheMacroExpansionContext (GUARD)
--- PASS: TestCopyDuplicatesTheMacroExpansionID      (GUARD)
--- PASS: TestCopyPreservesNilMacroExpansion         (GUARD)
--- PASS: TestTokenCopy                              (new API, neither)
```

Six catches, six guards labelled in-file, one unit test for API that did not exist on `main` and is labelled as such rather than left to read as a catch.

The most useful of the six states the consequence as **output** rather than as a pointer comparison. `hoistOperandComments`' move, performed through a copy:

```
Error: Not equal:
  expected: "; lead\n(+ 1 2) ; trail\n"
  actual  : "(+ 1 2) ; trail\n"
Messages: moving comments off the COPY changed how the ORIGINAL formats (#466)
```

A source comment silently dropped from a file `elps fmt` rewrites — not a debugging curiosity.

## Live or latent

Latent, on the same population argument #467 records and by the same method. The parser's `Meta` writes all happen on nodes it has just built, before anything can have copied them; no non-test caller of `LVal.Copy` writes through `Meta` or `MacroExpansion` afterwards; the formatter and minifier read `Meta` and never call `Copy`. So the writes the tests perform are the tests' writes — they demonstrate the mechanism. What is **observed** is the sharing, which the tests assert directly.

## Neighbourhood audit

- `parser/token.Token` gained a `Copy`, nil-preserving and separating `Source`, on the reasoning `Location.Copy` already documents (#362/#366). It is the only new exported API.
- `lisp.Quote` shallow-copies an unquoted operand including its `Meta` pointer. That is **not** this defect and is deliberate — `hoistOperandComments`' `outer.Meta == inner.Meta` guard exists precisely to handle it, and its comment says so. Left alone.
- `SingletonSnapshot.singletonLValEqual` compares `Meta` and `MacroExpansion` by pointer. Singletons carry nil for both and `Copy` preserves nil, so the snapshot is unaffected; `TestCopyPreservesNilMeta` and `TestCopyPreservesNilMacroExpansion` pin that.
- `LArray` shares its `Cells` backing and `LSortMap` its value pointers, both deliberately. Those are about the values reached *through* a node; this change is about the node's own metadata and does not touch them.
- `#382` plans to unexport both fields in the same release break, and `#362` tracks making `LVal.Source` immutable, which would settle the `Location`s reachable through `Meta` along with the rest. Neither is anticipated here.

## Gates

`go build`, `go vet`, `go test -count=1 ./...`, `go test -race ./...`, `golangci-lint run ./...`, `elps fmt -l ./...`, `elps doc -m`, `make go-test`, `make race`, `make test-elpscheck`, `make test-examples`, `CI_GATES_REQUIRE_SHELLCHECK=1 ./scripts/ci-gates-test.sh`, `./scripts/confidentiality-guard.sh`, `./scripts/fuzz-budget-check.sh` — all clean.

Benchmarks: `Meta` and `MacroExpansion` are nil on every hot path, so the added work is two nil-pointer comparisons per node and no allocation. No row moves outside its own spread.

---

### DIFF
--- a/lisp/copy_meta_test.go
+++ b/lisp/copy_meta_test.go
@@ -1,0 +1,414 @@
+// Copyright © 2026 The ELPS authors
+
+// Regression tests for elps#466 -- LVal.Copy sharing Meta and MacroExpansion.
+//
+// #446 (PR #467) separated LVal.Source, and the comment it left on Copy now
+// says "the copy owns its positions".  The same `*cp = *v` carried the other
+// two metadata pointers across untouched, so that sentence was true of the
+// node and false one level down: SourceMeta holds []*token.Token, every token
+// holds a *token.Location, and both trees reached the same objects.
+//
+// WHY THIS IS THE CODE AND NOT THE DOC COMMENT.  #466 leaves that open --
+// "these are metadata ABOUT a node which a copy legitimately shares" is a
+// coherent reading, and if it were the right one then Copy's doc comment and
+// MacroExpansionInfo.ID's would be what needed fixing.  The evidence says it
+// is the right reading for exactly one of the three pointers:
+//
+//   - SourceMeta is not a description of a node, it is per-node MUTABLE state
+//     that the parser writes in place at a dozen sites, and that
+//     rdparser.hoistOperandComments MOVES between nodes (append onto outer,
+//     `= nil` on inner).  The parser already special-cases two LVals holding
+//     one *SourceMeta -- the `outer.Meta == inner.Meta` guard, whose comment
+//     explains that moving the comments would move them onto themselves.  An
+//     in-tree guard against a state means the state is an anomaly, and
+//     LVal.Copy manufactured it deliberately.  The comment-hoist test below
+//     performs that exact move through a copy and watches the ORIGINAL's
+//     rendered output change.
+//
+//   - MacroExpansionInfo is per node too (the ID is what distinguishes one
+//     node from another), so the struct is separated.
+//
+//   - Its embedded *MacroExpansionContext genuinely is shared metadata: it
+//     describes the macro CALL, it is documented "shared across all nodes in
+//     one expansion", and #456 already made its CallSite an object the
+//     expansion owns.  It stays shared, and the context test below pins that.
+//
+// The one place the doc comment WAS the thing that was wrong is
+// MacroExpansionInfo.ID, documented "unique per node".  LVal.Copy cannot
+// honour that under any implementation -- it takes no *Runtime and so has no
+// counter to draw a fresh ID from -- so the comment now says what is true and
+// names the consumer that cares (lisp/x/debugger's stepper, which steps on
+// MacroID changing).  TestCopyDuplicatesTheMacroExpansionID pins it.
+//
+// LIVE OR LATENT: latent, on the same population argument #467 records.  The
+// parser's Meta writes all happen on nodes it has just built, before anything
+// can have copied them, and no non-test caller of LVal.Copy writes through
+// Meta or MacroExpansion afterwards.  So the writes these tests perform are
+// the TESTS' writes; they demonstrate the mechanism.  What is observed rather
+// than demonstrated is the SHARING, which the tests assert directly and which
+// failed on 95e2e1a.
+//
+// Tests marked GUARD pass before the fix as well as after.  They pin
+// behaviour the fix must not break; they are not catches.
+
+package lisp_test
+
+import (
+	"strings"
+	"testing"
+
+	"github.com/luthersystems/elps/formatter"
+	"github.com/luthersystems/elps/lisp"
+	"github.com/luthersystems/elps/parser/rdparser"
+	"github.com/luthersystems/elps/parser/token"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// readFormatting parses src in format-preserving mode, which is the only mode
+// that populates Meta at all.
+func readFormatting(t *testing.T, src string) []*lisp.LVal {
+	t.Helper()
+	sc := token.NewScanner("copy-meta.lisp", strings.NewReader(src))
+	exprs, err := rdparser.NewFormatting(sc).ParseProgram()
+	require.NoError(t, err)
+	require.NotEmpty(t, exprs)
+	return exprs
+}
+
+// metaSource is the probe from issue #466: one form with a leading comment and
+// an inline trailing comment, so Meta carries both a []*token.Token and a
+// single *token.Token.
+const metaSource = "; lead\n(+ 1 2) ; trail\n"
+
+// TestCopyDoesNotAliasSourceMeta is the reproduction from issue #466.
+// CATCH: both assertions failed on 95e2e1a.
+func TestCopyDoesNotAliasSourceMeta(t *testing.T) {
+	orig := readFormatting(t, metaSource)[0]
+	require.NotNil(t, orig.Meta, "premise: format-preserving parsing populates Meta")
+
+	cp := orig.Copy()
+	require.NotNil(t, cp.Meta, "the copy lost its formatting metadata")
+	assert.NotSame(t, orig.Meta, cp.Meta,
+		"a copy shares the original's *SourceMeta (#466)")
+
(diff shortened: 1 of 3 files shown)


iamsamwood (approved):

---

Issue #464, opened 2026-08-15
lsp emits BYTE columns where LSP specifies UTF-16 code units, and on protocol 3.16 cannot negotiate otherwise


sam-at-luther:
Found by **reading**, during the neighbourhood audit for #449 (PR #462); the byte-column half is confirmed by running.

## The gap

LSP defines a `Position.character` as an offset in **UTF-16 code units**. Before 3.17 that is not negotiable — it is simply the encoding.

This server emits byte offsets everywhere. `Scanner.LocStart` sets `Col` to `startPos - s.startLinePos + 1`, a byte offset, and `elpsToLSPPosition` passes it straight through:

```go
line := loc.Line
col := loc.Col
if line > 0 { line-- }
if col > 0 { col-- }
```

`position.go` splits content on `"\n"` and slices lines by byte throughout, and the inbound direction has the same assumption in reverse — a `character` the client sent in UTF-16 units is used directly as a byte index.

There is no conversion in either direction and no `positionEncoding` capability: the server is built on `glsp/protocol_3_16`, whose `ServerCapabilities` struct has no such field (it was added in LSP 3.17). So the escape hatch that would let a server legitimately declare "I speak UTF-8" is not available on the protocol version in use.

## Consequence

For any line containing a non-ASCII character, every column the server sends and every column it receives is displaced, in both directions, by `bytes - utf16units` for the text to the left of the position. Concretely, on a line beginning `(defun é`:

- a range the server sends lands to the **right** of where it means, growing with the number of multi-byte characters before it;
- a cursor position the client sends is read as a byte index, so `symbolAtPosition` looks in the wrong place and hover/definition/completion answer for the wrong token or for nothing.

Non-BMP characters (emoji, many CJK extension characters) are the worst case: 4 bytes against 2 UTF-16 units.

## Scope

This is not fixable in any one handler — `elpsToLSPPosition` and its inbound counterpart are the two choke points, and a conversion there needs the document text, which they are not currently given. It is also not fixable by changing a length: PR #462 keeps semantic-token lengths in **bytes** precisely so that a length agrees with the column it is added to, rather than mixing units within one token. Whatever the fix is, it has to move start and end together, and inbound and outbound together.

Worth deciding at the same time whether to move to `protocol_3_17` and advertise `positionEncoding: "utf-8"`, which is the cheap correct answer if the clients in use support it, versus converting to UTF-16 at the boundary, which is correct for every client.

## Not the same as #463

[...]

Issue #463, opened 2026-08-15
token.TokenEnd counts EndCol in RUNES onto a byte Col, so every lsp range is short for a non-ASCII name — and rename silently corrupts the source


sam-at-luther:
Found by **running**, during the neighbourhood audit for #449 (PR #462).

## The mismatch

`Scanner.LocStart` computes `Col` as a **byte** offset within the line:

```go
Col:  startPos - s.startLinePos + 1,
```

`token.TokenEnd` computes `EndCol` by ranging over the token text, which iterates **runes**:

```go
for _, ch := range tok.Text {
	if ch == '\n' { line++; col = 1 } else { col++ }
}
```

So `EndCol == Col + runeCount(text)` — a rune width added to a byte base. On a token that is itself pure ASCII the two agree by accident; on any token containing a multi-byte rune, `EndCol` is in neither unit and is short by `len(text) - runeCount(text)`.

`EndPos` in the same function is `tok.Source.Pos + len(tok.Text)`, i.e. bytes, and is correct. It is only `EndCol` that is mixed.

## Measured

Document `(defun éx (a) a)\n(éx 1)\n`. The symbol `éx` is 3 bytes and 2 runes, starting at byte column 8 (1-based).

```
SYMBOL "éx" Line=1 Col=8 EndLine=1 EndCol=10 Pos=7 EndPos=10
  elpsToLSPRange -> start.Character=7 end.Character=9   (byte-correct end is 10)
```

`Col=8` is a byte column, `EndPos-Pos=3` is the correct byte width, but `EndCol=10` = `8 + 2 runes`. The range is one byte short.

## Who is affected

`elpsToLSPRange` prefers `EndCol` whenever it is set, so every span-first caller inherits this: `call_hierarchy.go`, `code_actions.go`, `definition.go`, `highlight.go`, `hover.go`, `linked_editing.go`, `references.go`, `rename.go`. `position.go`'s `locContainsCol` and `selection_range.go`'s `locContainsPosition` compare against `EndCol` the same way, so a cursor on the last byte of a non-ASCII name is treated as outside the name.

Hover on the definition above returns `[0:7..0:9]` for text occupying bytes 7..10.

## This is not cosmetic in rename

`textDocument/rename` builds its `TextEdit` ranges from the same helper. Renaming `éx` to `zz`:

```
EDIT file:///…/rename.lisp [0:7..0:9] -> "zz"
EDIT file:///…/rename.lisp [1:1..1:3] -> "zz"
```

[...]

PR #497, opened 2026-08-16
diagnostic: measure the caret underline in terminal cells, not bytes

### LINKED ISSUE #469: diagnostic renderer sizes its underline in BYTES but indents it in display columns, so carets overrun a non-ASCII span
Found by **running**, during the consumer audit for #463 (PR for #463). Cosmetic — terminal diagnostic output only, no data loss — and **distinct from #463**: `diagnostic.Span` is its own type and no producer in the tree bridges `token.Location.EndCol` into it, so the #463 fix does not touch this.

## The mismatch

`diagnostic/renderer.go` `renderSpan` computes the two halves of the underline in different units:

```go
underLen := endCol - col + 1          // BYTES: span.Col is a byte column and
                                      // detectEndCol returns a byte index

prefix = source[:col-1]
displayCol := displayWidth(prefix)    // DISPLAY COLUMNS: displayWidth ranges
                                      // over RUNES (tabs -> 4)

underPad := strings.Repeat(" ", displayCol)
underline := strings.Repeat("^", underLen)
```

The indent is rune-counted, which is right — a terminal advances one cell per rune, not per byte. The length is byte-counted, which is not. On any span containing a multi-byte rune the caret run is longer than the text it points at.

## Measured

Rendering a span with `EndCol` unset (so `detectEndCol` supplies it) on current `main` (`4737835`):

```
SRC "(set! é 1)"
 1 |  (set! é 1)
   |        ^^          <- 2 carets under a 1-column token

SRC "(f 加算 1)"
 1 |  (f 加算 1)
   |     ^^^^^^        <- 6 carets under a 2-rune token, spilling over " 1)"

SRC "(set! abc 1)"      <- ASCII: correct, byte == rune == column
 1 |  (set! abc 1)
   |        ^^^
```

The CJK case is the clearest: `加算` is 6 bytes, so it gets six carets, and they run past the token and underline the following text as well.

Strictly the right unit for `underLen` is neither bytes nor runes but **display width** — `displayWidth` of the spanned text — since the indent it has to line up with is already computed that way. East Asian wide characters occupy two cells, so `加算` should get four carets, not two and not six; `displayWidth` does not currently know about width classes either, which is a second, smaller gap in the same function.

## A latent trap next to it

`underLen := endCol - col + 1` is **inclusive** arithmetic. `token.Location.EndCol` is documented (as of the #463 fix) as **exclusive**. Nothing bridges the two today — `cmd/diagnostic.go` and `repl/diagnostic.go` both set `Col` only and leave `Span.EndCol` zero for auto-detection — but the two types now have same-named fields with opposite conventions, and the first person to wire an analyser's end position into a rendered span will get an underline one caret too long. Worth either renaming or documenting the convention on `diagnostic.Span.EndCol` while this is being looked at.

## Suggested fix

Compute `underLen` as `displayWidth(source[col-1 : endCol-1])`, bounds-checked, so both halves of the underline are measured the same way. `detectEndCol` already advances by `utf8.DecodeRuneInString`, so it returns a valid byte boundary and the slice is safe.

sam-at-luther:
Fixes #469

## Reproduced

On `95e2e1a`, rendering spans with `EndCol` unset so `detectEndCol` supplies it:

```
SRC "(set! é 1)"     1 |  (set! é 1)
                       |        ^^          2 carets under a 1-cell token

SRC "(f 加算 1)"      1 |  (f 加算 1)
                       |     ^^^^^^        6 carets under a 4-cell token, spilling over " 1)"

SRC "(f 𝛼𝛽 1)"       1 |  (f 𝛼𝛽 1)
                       |     ^^^^^^^^      8 carets under a 2-cell token

SRC "(f é́ 1)"        1 |  (f é́ 1)
                       |     ^^^^          4 carets under a 1-cell token (e + U+0301)

SRC "(set! abc 1)"   1 |  (set! abc 1)
                       |        ^^^        ASCII: correct
```

## Bytes, runes, or cells

The issue asks the question and this is the answer: **cells**, and it matters.

`underLen` was bytes and `displayCol` was a rune count. Fixing only the mismatch — measuring both halves in runes — would have stopped the reported spillover and still been wrong, because the underline has to line up with what a **terminal** draws and a terminal does not advance one cell per rune. An East Asian wide character advances two; a combining mark advances none. Under a rune count `加算` gets two carets under a four-cell token, and `é` written as `e` + U+0301 gets two under one cell.

So both halves now go through `displayWidth`, which is the one place that decides what a cell is, and they agree by construction rather than by coincidence. The CJK rows in the tests are what separates the two candidate fixes: six is the byte answer, two is the rune answer, **four** is right.

**What the repo already had:** `displayWidth` in this file, counting one per rune with tabs expanded to four — no width classes anywhere in the tree. `github.com/mattn/go-runewidth` was already in the module graph as an indirect dependency via the direct `muesli/reflow`, so this only moves it from indirect to direct; `go.sum` is unchanged and no module is added.

It is used through an **explicit `Condition`** rather than the package-level default, which matters: go-runewidth's `init` reads `LC_ALL`/`LC_CTYPE`/`LANG` and sets `EastAsianWidth` from it, under which the ~1000 codepoints in Unicode's Ambiguous width class (`→`, `①`, Greek, Cyrillic, box drawing) become two cells wide. Left on the default, the same diagnostic would be underlined differently depending on the locale of the shell that ran `elps`, and this package's own golden tests would pass or fail on the same grounds. Ambiguous-as-narrow is what essentially every modern terminal does.

`tabWidth` and `tabExpansion` are now named constants. The source line and the caret line are laid out by two separate pieces of code and only line up because these agree; `TestTabExpansionMatchesTabWidth` pins that.

The out-of-range clamping is new. `Span` is an exported type filled in by callers and nothing validates it; the fix computes a slice from `Col`/`EndCol`, and a slice is the kind of thing that panics.

## The latent trap next to it

`diagnostic.Span.EndCol` is used **inclusively**; `token.Location.EndCol` is documented (since #463) **exclusive**. Nothing bridges them today — `cmd/diagnostic.go` and `repl/diagnostic.go` both set `Col` only. So this is not a defect and nothing is renamed: an exported field is not worth breaking for a convention that can be written down. The convention is now on the type, with the conversion (`EndCol: loc.EndCol - 1`) and a pointer at #463, and `TestSpanEndColIsInclusive` makes it executable so a silent switch fails here rather than in someone's terminal.

**This is deliberately not #464.** That is the LSP server emitting byte columns where LSP 3.16 specifies UTF-16 code units, it is another agent's, and it is in `lsp/` — untouched here. `diagnostic` is a standalone package with no LSP consumer, and "the two ends of one underline are counted in the same unit" is a property under any answer #464 gets.

## Red then green

`diagnostic/underline_width_test.go`. Against `95e2e1a` (`renderer.go` restored, the two new constants stubbed so the file compiles):

```
--- FAIL: TestUnderlineCoversExactlyTheSpannedText
    --- FAIL: /two-byte-rune          --- FAIL: /cjk-wide
    --- FAIL: /astral-narrow          --- FAIL: /combining-mark
    --- FAIL: /emoji                  --- FAIL: /tab-then-cjk
    --- FAIL: /wide-before-the-span
    --- PASS: /ascii-GUARD            --- PASS: /tab-indent-GUARD
--- FAIL: TestUnderlineIsMeasuredInTheSameUnitAsTheIndent   (7 of 8 source lines)
--- FAIL: TestDisplayWidthCountsCells                       (cjk, hangul, fullwidth, emoji, decomposed accent)
--- FAIL: TestUnderlineSurvivesOutOfRangeSpans
--- PASS: TestSpanEndColIsInclusive                         (GUARD)
--- PASS: TestTabExpansionMatchesTabWidth                    (GUARD)
```

A failure prints the two lines one above the other, which is the form the defect is obvious in:

```
  source: (f 加算 1)
  got:       ^^^^^^
  want:      ^^^^
  (6 carets, wanted 4)
```

`wide-before-the-span` is worth noting: on `main` it fails on the **indent**, not the length — `displayWidth` counted `加算` as two, so the carets started two columns left of the token. Fixing only `underLen` would have left that.

`TestUnderlineIsMeasuredInTheSameUnitAsTheIndent` states the same defect as a property over every rune-boundary start column of eight source lines, so it cannot be satisfied by writing the wrong string into a `want`.

All nine existing tests in `renderer_test.go` still pass unchanged, including `TestRenderErrorGolden`.

## Neighbourhood audit

- `detectEndCol` already advances by `utf8.DecodeRuneInString`, so it returns a valid rune boundary and the new slice is safe. Unchanged.
- `displaySource`'s tab replacement and `displayWidth`'s tab handling are the only other place two measurements have to agree in this file; they now share a constant.
- `lsp/`, `lint/`, `analysis/`, `mcpserver/` all carry their own position types and none constructs a `diagnostic.Span`. `lsp/position.go`'s byte-vs-UTF-16 question is #464 and is untouched.
- No other caller of `displayWidth` exists.

## Gates

`go build`, `go vet`, `go test -count=1 ./...`, `go test -race ./...`, `golangci-lint run ./...`, `elps fmt -l ./...`, `elps doc -m`, `make go-test`, `make race`, `make test-elpscheck`, `make test-examples`, `CI_GATES_REQUIRE_SHELLCHECK=1 ./scripts/ci-gates-test.sh`, `./scripts/confidentiality-guard.sh`, `./scripts/fuzz-budget-check.sh` — all clean.

No benchmark touches `diagnostic`.

---

### DIFF
--- a/diagnostic/diagnostic.go
+++ b/diagnostic/diagnostic.go
@@ -28,11 +28,31 @@ func (s Severity) String() string {
 }
 
 // Span identifies a region of source code to highlight in the diagnostic.
+//
+// UNITS AND CONVENTION.  Col and EndCol are 1-based BYTE columns, and EndCol
+// is INCLUSIVE -- it names the column of the span's last byte, not the one
+// after it.  A span covering "false" at columns 7 through 11 is Col: 7,
+// EndCol: 11.
+//
+// That is the OPPOSITE of parser/token.Location.EndCol, which is documented
+// (as of the #463 fix) as an EXCLUSIVE byte column.  The two types have
+// same-named fields with opposite conventions, and nothing in the tree bridges
+// one into the other today: cmd/diagnostic.go and repl/diagnostic.go both set
+// Col and leave EndCol zero for auto-detection.  The first caller to wire an
+// analyser's end position straight into a Span would get an underline one
+// caret too long, which is why the convention is written down here rather than
+// left to be inferred (issue #469).  Converting is `EndCol: loc.EndCol - 1`.
+//
+// The RENDERER measures the underline in terminal CELLS rather than in bytes,
+// so the carets line up with what a terminal draws even when the span holds
+// multi-byte, East Asian wide, or combining characters -- see displayWidth in
+// renderer.go.  The byte convention here is about how a caller ADDRESSES the
+// source; it is not the unit the carets are counted in.
 type Span struct {
 	File   string // path for reading source; display name if unreadable
 	Line   int    // 1-based line number
-	Col    int    // 1-based start column
-	EndCol int    // 1-based end column (0 = auto-detect from source)
+	Col    int    // 1-based start byte column
+	EndCol int    // 1-based end byte column, INCLUSIVE (0 = auto-detect from source)
 	Label  string // text shown under the underline
 }
 
--- a/diagnostic/renderer.go
+++ b/diagnostic/renderer.go
@@ -11,6 +11,8 @@ import (
 	"strconv"
 	"strings"
 	"unicode/utf8"
+
+	"github.com/mattn/go-runewidth"
 )
 
 // Renderer formats diagnostics as Rust-style annotated source snippets.
@@ -129,10 +131,26 @@ func (r *Renderer) writeSpan(ew *errWriter, span Span, p palette) {
 
 	// Source line with line number
 	// Replace tabs with spaces for consistent alignment
-	displaySource := strings.ReplaceAll(source, "\t", "    ")
+	displaySource := strings.ReplaceAll(source, "\t", tabExpansion)
 	ew.printf(" %s%s |%s  %s\n", p.boldBlue, lineStr, p.reset, displaySource)
 
-	// Underline
+	// Underline.
+	//
+	// Both halves of it -- how far in the carets start, and how many there
+	// are -- have to be measured in the SAME unit, and that unit has to be
+	// TERMINAL CELLS, because a terminal is what lays the two lines out on
+	// top of each other.  Issue #469 was that they were not: the indent was
+	// computed with displayWidth (cells) and the length with `endCol - col +
+	// 1` (bytes), so any span containing a multi-byte rune got more carets
+	// than the text it pointed at and they ran on over what followed.  On
+	// "(f 加算 1)" the two-rune token got six carets.
+	//
+	// Bytes are not the unit and neither are runes.  A rune count is right
+	// for the common case only because most runes happen to occupy one cell;
+	// an East Asian wide character occupies two and a combining mark
+	// occupies none, so 加算 needs FOUR carets rather than two or six.  Both
+	// measurements below therefore go through displayWidth, which is the one
+	// place that decides what a cell is.
 	col := span.Col
 	endCol := span.EndCol
 	if col <= 0 {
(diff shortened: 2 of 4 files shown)


iamsamwood (approved):

---

PR #500, opened 2026-08-16
fuzzwatch: consult Starved() at both caps so a stalled window is never a hang (#488)

### LINKED ISSUE #488: fuzzwatch.Check returns Hung for a window it simultaneously classifies as Starved, so Inconclusive is unreachable above 25% CPU share
Found by **running**, during the neighbourhood audit for #454. Reproduced live, then confirmed as arithmetic.

## What happened

Running `./lisp/x/debugger/` under 400 spinners on 4 cores (the load PR #447 established as the house condition), `TestDebugSeedsTerminate` failed 4/4 runs, 0/4 unloaded. Three of the failures came from the fuzzwatch-denominated watchdog in `debugEvalBudgeted`:

```
debug_fuzz_test.go:516: debugged evaluation did not terminate within 30s of SCHEDULED time
despite a 2s context deadline and a 200000-step budget; a pause with no resume looks exactly
like this (wall 1m24.561s, of which 30.561s scheduled and 54s lost to scheduler stall
(longest single stall 10.7s))
```

Read that report carefully. `fuzzwatch` **did** see the starvation this time — 54s lost, longest single stall 10.7s, only 36% of wall clock scheduled. It reported the process as starved and then failed the test anyway.

## Why

`Budget.Check` tests the arms in this order:

```go
switch {
case r.Scheduled() >= b.total:
	return Hung, 0, r
case r.Wall >= b.hardWall:
	if r.Starved() {
		return Inconclusive, 0, r
	}
	return Hung, 0, r
}
```

The first arm wins unconditionally, and it never consults `Starved()`. `Inconclusive` lives only inside the second arm, which is reached only once `Wall >= 4 * total`.

So `Inconclusive` is unreachable whenever scheduled time crosses the budget before wall clock crosses four times the budget — that is, **whenever the process gets more than ~25% of the CPU**. Between 25% and 100% share, a starved process is reported as a hang.

Checked against the observed numbers (`Report{Wall: 84.561s, Lost: 54s}`, `total = 30s`):

```
Scheduled()=30.561s  total=30s  Scheduled>=total = true
Starved()  = true    (Lost*2=1m48s >= Wall=1m24.561s)
hardWall   = 2m0s;   Wall >= hardWall = false
=> Check() takes the FIRST arm and returns Hung, never consulting Starved()
   share of wall actually scheduled: 36.1%
```

`Starved()` is `true` and the verdict is `Hung`. Those two should not be able to co-occur — `Starved()` exists precisely to mean "do not blame the code under test".

## Why this matters more than the headroom argument suggests

The package doc's defence of the nine watchdogs is that they are sized 20,000x–90,000x above mean work, so no plausible slowdown reaches them. That argument holds for the *budget*, but this failure did not come from the work being slow. It came from **`Check` reaching a verdict it had the evidence to avoid**. Once stalls are long enough to be visible at all — 10.7s here, far above the 400ms floor — the accounting works and the interpretation throws the answer away.

This is complementary to #453, not a duplicate. #453 is about stalls fuzzwatch **cannot see** (CPU share, sub-400ms). This is about a stall it **did** see and then discarded. PR #483 documents #453's boundary and deliberately makes no behaviour change; this one is a behaviour question and is filed separately so it can be judged on its own.

## Suggested direction

Consult `Starved()` in the first arm too — a window that is more than half stall should not return `Hung` regardless of which cap was hit. Something like:

```go
case r.Scheduled() >= b.total:
	if r.Starved() {
		return Inconclusive, 0, r
	}
	return Hung, 0, r
```

That is a real semantic change (some genuine hangs on a busy machine become Inconclusive rather than failures) so it wants a deliberate decision about which error the project prefers, plus the red-then-green under load that PR #447 set the pattern for. **Please do not simply widen the budget** — #443/#452 and #435/#447 both established that raising a threshold is the wrong answer here.

## Reproducing

```
for i in $(seq 1 400); do ( while :; do :; done ) & done
go test -run 'TestDebugSeedsTerminate' -count=1 ./lisp/x/debugger/
pkill -f 'while :'
```

4/4 failures loaded, 0/4 unloaded, on a 4-core box. The arithmetic above needs no load at all — it is a unit-level property of `Check`.

sam-at-luther:
Fixes #488

## The defect

`Budget.Check` tested the scheduled-time cap first and returned `Hung` from it unconditionally:

```go
switch {
case r.Scheduled() >= b.total:
	return Hung, 0, r          // never consults Starved()
case r.Wall >= b.hardWall:
	if r.Starved() {
		return Inconclusive, 0, r
	}
	return Hung, 0, r
}
```

`Starved()` exists precisely to mean "do not blame the code under test". It and `Hung` must not be able to co-occur, and they could.

## Red, by running

Two ways, because the arithmetic alone does not show that the monitor, the `Report` and the verdict line up end to end on a live process.

**1. The window #488 observed live**, replayed against the real `Check` (`internal/fuzzwatch/fuzzwatch_test.go`):

```
--- FAIL: TestCheckDoesNotCallAStarvedWindowAHang
    verdict = Hung for a window Check itself classifies as starved
    (wall 1m24.561s, of which 30.561s scheduled and 54s lost to scheduler stall)
--- FAIL: TestHungAndStarvedNeverCoOccur
    wall=30s lost=16.5s: verdict Hung for a starved window
    wall=40s lost=22s:   verdict Hung for a starved window
    ... 26 more points across the share space
```

**2. A real scheduler stall**, not a synthetic `Report` (`internal/fuzzwatch/stall_probe_unix_test.go`). The test re-executes itself as a child, the child arms a real `Budget` on the real process-wide heartbeat, and the parent freezes it mid-window with `SIGSTOP` and resumes it with `SIGCONT`:

```
against main's Check:
  attempt 1: verdict=hung          wall=4.616s lost=2.9s scheduled=1.716s starved=true (37% of wall scheduled)
  FAIL: a real scheduler stall produced verdict=hung -- Starved() means
        'do not blame the code under test' (#488)

on this branch:
  attempt 1: verdict=inconclusive  wall=4.704s lost=3s   scheduled=1.704s starved=true (36% of wall scheduled)
  PASS
```

36% of wall clock scheduled is the number #488 reported from the live failure, arrived at independently.

`SIGSTOP` rather than CPU load is deliberate and is the only honest way to do this here: PR #483 measured that contention does **not** produce stalls this instrument can see (200 spinners on 4 cores, `lost=0s`), because starvation is microseconds at a time and the 400ms floor swallows it. A frozen process — VM steal, cgroup throttling, an SMR pause — is the failure mode the package doc names, so it is the real condition for this instrument, and it is deterministic.

The harness asserts the invariant on **every** window the child produces, whatever its shape; and if it cannot build the #488 shape in three attempts it **fails**, because "the harness could not run" is not the same as "the check passed".

## The change

```go
spent := r.Scheduled() >= b.total
outOfWall := r.Wall >= b.hardWall
if spent || outOfWall {
	if r.Starved() {
		return Inconclusive, 0, r
	}
	return Hung, 0, r
}
```

Both caps now consult `Starved()`. `Continue` and its re-arm arithmetic are untouched.

### Why the two caps can share one branch

The wall cap's old "not starved, therefore `Hung`" fallback was **unreachable code**. Reaching it required `Wall >= 4*total` *and* `Scheduled < total`, which gives `Wall >= 4*total > 4*Scheduled > 2*Scheduled` — the definition of `Starved()`. `TestTheWallCapIsAlwaysStarved` pins that arithmetic over the whole space so the collapse is not taken on trust. (It also means `TestCheckHungAtTheWallCapWhenNotStarved` never exercised the wall cap despite its name — with 40s wall and 1s lost the budget is long spent, so it took the first arm. Its comment now says what it actually covers; it is still the load-bearing negative and still passes.)

## Which verdicts become reachable, and what is lost

| verdict | before | after |
|---|---|---|
| `Continue` | unchanged | unchanged |
| `Hung` | any window reaching either cap, starved or not | only windows reaching a cap that are **not** starved |
| `Inconclusive` | only below ~25% scheduled share (`Wall >= 4*total` with `Scheduled < total`) | any window reaching a cap with scheduled share **at or below 50%** (`Starved()` is `Lost*2 >= Wall`) |

Nothing previously reachable becomes unreachable. The one branch that disappears — wall cap, not starved, `Hung` — was already dead.

`TestHungAndStarvedNeverCoOccur` sweeps the share space and asserts both directions (`Hung` never co-occurs with `Starved`, `Inconclusive` never occurs without it), and fails if the highest share yielding `Inconclusive` is still at or below 25%. It reports `Inconclusive reachable up to 45% scheduled share`.

**The cost is real and is the point of the issue.** A genuine hang on a machine that was stalling for more than half the window now reports `Inconclusive` instead of failing. That is the error this project prefers: fuzzing is a repeated-trial process, so a real hang is re-found by the next run or by the same input on a quieter machine, whereas a false `Hung` writes a crasher blaming an innocent input — an investigation wasted and the corpus polluted permanently. It is also confined to demonstrably starved windows: on a healthy machine no stall is recorded, `Starved()` is false, and `Hung` fires at exactly the configured budget as before, which is the path every one of the eleven live watchdogs takes today.

No threshold was widened and no budget was retuned. Per #488, and per #443/#452 and #435/#447 before it, raising a threshold is not the answer here. Per PR #483, #447's rejected scheduled-time budget is not re-proposed.

Every call site already handles `Inconclusive` by skipping the input (`t.Skipf("no verdict: the process was starved throughout")`), so nothing outside `internal/fuzzwatch` changes.

## Negative control

`internal/fuzzwatch/verdict_control_test.go`. The starvation guard is deleted from a **copy generated from the live `fuzzwatch.go` on every run** — never a fixture, so it cannot drift away from what it guards — the copy is compiled as a standalone module together with a generated probe replaying the #488 window, and the old verdict must come back.

Both directions run through the same rig, because either alone would be satisfied by a broken rig:

```
control (intact):   probe FAILED   verdict = inconclusive, want Hung
control (stripped): probe PASSED
negative control: REMOVING the starvation guard restores the #488 verdict
(Hung for a starved window), so the guard is load-bearing
```

If the guard can no longer be located the control fails loudly and names what it was looking for, rather than silently testing nothing. If no Go toolchain can be found it fails rather than skipping, for the same reason.

## Found while reproducing, filed separately, not folded in

The monitor is sampled asynchronously, so at the instant a freeze **ends** the heartbeat that will charge it has not run yet. A `Check` performed at that moment sees `lost=0s` and charges the entire freeze to the code under test. Measured with this harness before the child was given a settle: `wall=3.306s lost=0s scheduled=3.306s starved=false verdict=hung`, of which 3.0s was a real `SIGSTOP`. Filed rather than folded in here — it is a different mechanism from #488 (accounting lag, not arm ordering) and it wants its own judgement.

## Gates

`go build`, `go vet`, `go test -count=1 ./...`, `go test -race ./...`, `golangci-lint run ./...` (0 issues), `elps fmt -l ./...`, `elps doc -m`, `make go-test`, `make race`, `make test-elpscheck`, `make test-examples`, `./scripts/confidentiality-guard.sh`. `./scripts/fuzz-budget-check.sh` unchanged: 30 targets / 8 shards / 10 min headroom. `CI_GATES_REQUIRE_SHELLCHECK=1 ./scripts/ci-gates-test.sh` unchanged at 283. No workflow or `scripts/` file touched.

The two new tests add ~8s to `internal/fuzzwatch` (a ~4.6s frozen child, and two small `go test` compiles for the control).

---


iamsamwood (approved):

---

Issue #502, opened 2026-08-16
ci-gates-test.sh: assert_not_contains reports PASS when the command it runs does not exist or crashes


sam-at-luther:
Found **by running**, while adding coverage for #495 (PR #503). Filed rather than folded in — `scripts/ci-gates-test.sh` is contended today and this is a change to a shared helper, not to the govulncheck assertions.

## The helper

```bash
# assert_not_contains <needle> <description> <command...>
assert_not_contains() {
	local needle="$1" desc="$2"
	shift 2
	local out
	out=$("$@" 2>&1)
	if echo "$out" | grep -qF -- "$needle"; then
		bad "$desc — output unexpectedly contained '$needle'"
		echo "$out" | sed 's/^/        | /'
	else
		ok "$desc"
	fi
}
```

The exit status of `"$@"` is never consulted. A command that could not run produces no output, the needle is absent, and the assertion passes.

## Demonstrated

Driving the real helper, extracted from the live file:

```
$ assert_not_contains "REGRESSION" "demo: a command that does not exist" /nonexistent/script.sh
PASS  demo: a command that does not exist
pass=1 fail=0
```

I hit this for real: while developing against a copy with a wrong `SCRIPT_DIR`, every `assert_exit` and `assert_contains` in the section failed loudly with `No such file or directory`, and the single `assert_not_contains` in the same block sat there reporting `PASS`.

## Why it matters

This is the defect class the file exists to catch, in the file's own machinery. `assert_exit` and `assert_contains` both fail closed when the command cannot run — a missing script gives exit 127 and no output, which neither will accept. `assert_not_contains` is the one helper that reads "the thing did not happen" as "the thing was checked and did not happen".

Seven call sites use it (`scripts/ci-gates-test.sh`, current `main` plus PR #503). None of them is *only* covered by `assert_not_contains` — each sits next to an `assert_exit` or `assert_contains` on the same command, which is what kept this from mattering so far. That is a property of how the assertions happen to be written today, not a property of the helper, and it is not enforced anywhere.

Rated accordingly: this has not hidden a real defect, and I am not claiming it has. What it means is that a future `assert_not_contains` written on its own would be a check that cannot fail.

## Suggested fix

Capture and check the status, the way the other helpers effectively do:

```bash
out=$("$@" 2>&1)
rc=$?
if [ "$rc" -gt 125 ]; then
	bad "$desc — the command could not run (exit $rc), so nothing was asserted"
	...
```

[...]

PR #503, opened 2026-08-16
ci: stop the govulncheck gate un-failing the condition its own if: fires on (#495)

### LINKED ISSUE #495: The scheduled govulncheck gate step un-fails the exact condition its own `if:` fires on, so missing scan outputs go green
Found **by running**, during the audit requested alongside #493.

`.github/workflows/govulncheck-scheduled.yml`'s final step and the script it runs disagree about what an empty scan output means, and the script wins.

## The disagreement

The step's `if:` treats an empty output as **tripped** (fail-safe):

```yaml
- name: Fail if govulncheck reported a finding or a build break
  if: ${{ !cancelled() && (steps.scan.outputs.exit != '0' || steps.binscan.outputs.exit != '0') }}
  env:
    SOURCE_SCAN_EXIT: ${{ steps.scan.outputs.exit }}
    BINARY_BUILD_FAILED: ${{ steps.binscan.outputs.build_failed }}
    BINARY_VULNS_FOUND: ${{ steps.binscan.outputs.vulns_found }}
  run: bash scripts/govulncheck-fail-summary.sh
```

`'' != '0'` is **true** in Actions expression syntax, so a missing or empty `steps.scan.outputs.exit` *runs* this step. That is the correct, deliberate direction.

`scripts/govulncheck-fail-summary.sh` then treats the same empty value as **did not trip**:

```bash
SOURCE_SCAN_EXIT="${SOURCE_SCAN_EXIT:-0}"
BINARY_BUILD_FAILED="${BINARY_BUILD_FAILED:-0}"
BINARY_VULNS_FOUND="${BINARY_VULNS_FOUND:-0}"
```

`:-` substitutes `0` for the empty string, so every reason test falls through, `reasons` stays empty, and the script prints "No govulncheck condition tripped — nothing to fail" and exits 0.

The script's header documents this ("an unset/empty value counts as 'did not trip'"), so it is intentional in isolation — but it is the opposite of what the caller's `if:` assumes, and the net effect is that the one input meaning "the scan produced no output" is the one that passes.

## Demonstrated

```
$ SOURCE_SCAN_EXIT='' BINARY_BUILD_FAILED='' BINARY_VULNS_FOUND='' bash scripts/govulncheck-fail-summary.sh
No govulncheck condition tripped — nothing to fail.
>>> EXIT=0

$ env -u SOURCE_SCAN_EXIT -u BINARY_BUILD_FAILED -u BINARY_VULNS_FOUND bash scripts/govulncheck-fail-summary.sh
No govulncheck condition tripped — nothing to fail.
>>> EXIT=0

$ SOURCE_SCAN_EXIT=3 bash scripts/govulncheck-fail-summary.sh   # control
govulncheck failed for the following reason(s):
  - VULNERABILITY: the source-mode scan reported a finding.
>>> EXIT=1
```

So: the step runs *because* the outputs were missing, and then reports that nothing was wrong.

## Reachability

`steps.scan.outputs.exit` is written by `scripts/govulncheck-scan.sh` as `echo "exit=${status}" >> "${GITHUB_OUTPUT:-/dev/null}"`. It goes empty if that append fails, if `GITHUB_OUTPUT` is unset in the step's environment, if the output key is renamed on one side only, or if the step is replaced by one that does not emit it. Same class as #493: no route requires anyone to be careless, only for a refactor to move a name.

## Severity, stated honestly

Lower than #493, and I want to be clear about why rather than overstate it:

- This is the **scheduled** workflow, not a required PR check. It does not gate merging.
- The tracking-issue writer above it uses the same fail-safe expression (`FOUND: ${{ steps.scan.outputs.exit != '0' || ... }}`), so on this path `FOUND` is `true` and `scripts/govulncheck-drift.cjs` **still opens the drift issue**. The failure is therefore not silent — it is visible as an issue while the cron job itself reads green.

What is lost is the red signal on the run, and the job summary naming the condition. Anyone reading the Actions tab sees a passing scheduled scan.

## Suggested fix

Make the script agree with its caller: distinguish "did not trip" from "was not told". An unset variable and an explicit `0` are different inputs and should not collapse to the same branch — an empty `SOURCE_SCAN_EXIT` means the scan's status was never reported, which is a reason to fail, not a reason to pass. Exit 2 for "could not determine", matching the convention `scripts/fuzz.sh`, `scripts/fuzz-budget-check.sh` and `scripts/confidentiality-guard.sh` already use, would keep it distinguishable from a real finding.

Worth pairing with a behavioural assertion in `scripts/ci-gates-test.sh`, which currently has no coverage of this script at all beyond `bash -n` and shellcheck.

Not fixed here — filed rather than folded into #493's PR (#494) so that change is not widened.

sam-at-luther:
Fixes #495

## The disagreement

`.github/workflows/govulncheck-scheduled.yml`'s final step is fail-safe on purpose. `'' != '0'` is TRUE in Actions expression syntax, so a missing `steps.scan.outputs.exit` **runs** the step rather than skipping it.

`scripts/govulncheck-fail-summary.sh` then read the same empty value as an explicit zero, via `${VAR:-0}`, found no reason to fail, and exited 0. The step ran *because* the outputs were missing and then reported that nothing was wrong.

## Red, by running

The real script, real inputs, no mock:

```
$ SOURCE_SCAN_EXIT='' BINARY_BUILD_FAILED='' BINARY_VULNS_FOUND='' bash scripts/govulncheck-fail-summary.sh
No govulncheck condition tripped — nothing to fail.
>>> EXIT=0

$ env -u SOURCE_SCAN_EXIT -u BINARY_BUILD_FAILED -u BINARY_VULNS_FOUND bash scripts/govulncheck-fail-summary.sh
No govulncheck condition tripped — nothing to fail.
>>> EXIT=0
```

## Green

```
$ SOURCE_SCAN_EXIT= BINARY_SCAN_EXIT= BINARY_BUILD_FAILED= BINARY_VULNS_FOUND= bash scripts/govulncheck-fail-summary.sh
::error title=govulncheck gate could not run::Scan outputs were empty or unset
(SOURCE_SCAN_EXIT BINARY_BUILD_FAILED BINARY_VULNS_FOUND BINARY_SCAN_EXIT), so
nothing was verified. An unreported scan status is not a clean one — refusing to
summarise this run as passing. See #495.
govulncheck gate could not determine its inputs; unreported:
  - SOURCE_SCAN_EXIT
  - BINARY_BUILD_FAILED
  - BINARY_VULNS_FOUND
  - BINARY_SCAN_EXIT
>>> EXIT=2
```

Unset behaves identically. Real inputs are unchanged: all-zero still exits 0, a source finding still exits 1 with the same wording, a build break still exits 1 and still says "This is NOT a vulnerability finding".

**Exit 2, not 1**, following `scripts/fuzz.sh`, `scripts/fuzz-budget-check.sh` and `scripts/confidentiality-guard.sh`: "the gate could not determine its inputs" is a different outcome from "a condition tripped", and conflating them leaves CI unable to tell a broken gate from a real finding. Only an input that is explicitly `0` is now read as "this condition was clear".

A missing input does not hide a condition that *did* trip among the inputs that arrived — both are printed, and the exit is 2 because the run's verdict is not fully determined.

## The other half of the same disagreement

The step's `if:` also fires on `steps.binscan.outputs.exit != '0'`, and that output was **never passed to the script**. So a binary pass reporting an aggregate failure that neither `build_failed` nor `vulns_found` accounts for would again have been summarised as "nothing to fail" — the identical shape, reachable the moment `govulncheck-binary-scan.sh` grows a third failure condition (a scanner error that is neither a build break nor a finding is already distinguished in its log, just not in its outputs).

It is an input now. An aggregate failure with no specific condition set is reported as `UNATTRIBUTED FAILURE` and exits 1:

```
$ SOURCE_SCAN_EXIT=0 BINARY_SCAN_EXIT=1 BINARY_BUILD_FAILED=0 BINARY_VULNS_FOUND=0 \
    bash scripts/govulncheck-fail-summary.sh
govulncheck failed for the following reason(s):
  - UNATTRIBUTED FAILURE: the binary-mode pass reported exit=1, but neither
    build_failed nor vulns_found is set. It failed for a reason this summary
    cannot name — read the step log.
>>> EXIT=1
```

This is the same three lines of `env:` and the same script as the empty-input fix, so it is not a separate change to review; it is the second half of "the script must never say nothing failed when its caller's gate said something did".

## Coverage: 20 assertions where there were none

`scripts/govulncheck-fail-summary.sh` had **no behavioural coverage** in `scripts/ci-gates-test.sh` — only the `bash -n` and shellcheck sweep, which prove it parses, not that it works. Two independent halves are added, because either alone has a gap.

**(A) Behavioural.** The real script is run with every input class: all clear, each real condition, empty, unset, one-missing-among-reported, and the unattributed aggregate. The real-input cases come first so a script that merely fails on everything is not mistaken for one that works.

**(B) Wiring / structural.** #495 was a disagreement between a workflow expression and a script. Neither file is wrong on its own, only the pair is, and nothing else in the suite reads both. The check runs in two directions: every `steps.*.outputs.*` the step's `if:` decides on must be bound in its `env:`, and every variable the script reads must be bound by the step. The step is located by what it **runs**, not by its name, so renaming it cannot silently drop the check; a `run:` it cannot find, an `if:` reading no outputs, or a script from which no required inputs can be extracted all fail as "the check could not run" rather than as clean.

### What it does not catch, stated plainly

- A wrong *value* reaching the script. This checks that the wire exists, not what travels down it.
- The `if:` expression being wrong in a way that skips the step entirely — that path runs nothing, so there is nothing here to observe.
- `govulncheck-scan.sh` / `govulncheck-binary-scan.sh` failing to write `$GITHUB_OUTPUT` in the first place. That is the *cause* of the empty input; this PR makes the consequence loud rather than preventing it.
- Anything in `.github/workflows/govulncheck.yml` (the PR-triggered gate). It does not use this script.

## Negative controls

Two, one per half, both **generated from the live files** by reintroducing the defect — never kept as fixtures, because a fixture drifts away from the thing it stands for and then the control tests history instead of the guard. A mutation that cannot be applied fails loudly rather than skipping.

1. **Behavioural**: put the `${VAR:-0}` collapse back into a copy of the real script (two mutation sites, both required to apply) and require the vacuous pass to return. The rig also pins that the collapsed copy still passes on all-clear and still fails on a real finding, so a control that merely broke the script is not mistaken for a working one. **The wiring check is blind to this control by construction** — it reads the workflow and the script's variable *names*, neither of which the mutation touches. Only running the thing catches it, which is what proves (A) is load-bearing rather than decorative.
2. **Structural**: unbind `BINARY_SCAN_EXIT` in a copy of the live workflow — exactly the state `main` was in — and require the wiring check to object. It calls the same `gv_check_wiring` the real assertion calls, so the control cannot pass by exercising a different code path.

Verified red-then-green at the suite level. Run against `main`'s script and workflow, 14 of the 20 new assertions fail; the 6 that pass are the real-input cases, which is the point of having them:

```
PASS  govulncheck-summary: all four outputs reported and clear passes (exit 0)
PASS  govulncheck-summary: a source-mode finding fails (exit 1)
PASS  govulncheck-summary: a binary-mode BUILD FAILURE fails (exit 1)
PASS  govulncheck-summary: a binary-mode finding fails (exit 1)
PASS  govulncheck-summary: a build break is not described as a security finding
FAIL  govulncheck-summary: EMPTY scan outputs refuse to report clean — want exit 2, got 0
FAIL  govulncheck-summary: UNSET scan outputs refuse to report clean — want exit 2, got 0
FAIL  govulncheck-summary: ONE missing output is enough to refuse — want exit 2, got 0
FAIL  govulncheck-summary: the undetermined case SAYS nothing was verified
FAIL  govulncheck-summary: the undetermined case NAMES the outputs it was not told
FAIL  govulncheck-summary: no longer claims nothing tripped when it was told nothing
FAIL  govulncheck-summary: an UNATTRIBUTED binary-pass failure fails — want exit 1, got 0
FAIL  govulncheck-summary: the unattributed case says so rather than reporting clear
FAIL  govulncheck-summary: the wiring check could not run (rc=2) — it verified nothing
FAIL  negative control could not reinstate the empty-input collapse
FAIL  negative control could not unbind BINARY_SCAN_EXIT from a copy of the workflow
```

On this branch all 20 pass.

## Severity, unchanged from the filing

This is the **scheduled** workflow, not a required PR check; it does not gate merging. The tracking-issue writer above it uses the same fail-safe expression, so `scripts/govulncheck-drift.cjs` still opens the drift issue on this path. What was lost is the red run and the job summary naming the condition — anyone reading the Actions tab saw a passing scheduled scan. That is worth fixing and is not worth more than that.

## Found while writing this, filed separately

`assert_not_contains` in `scripts/ci-gates-test.sh` reports PASS when the command it runs does not exist or crashes — it never consults the exit status. Filed as #502 rather than fixed here: it is a change to a shared helper in a file several people are editing today. No current call site depends on it alone.

## Contended file

`scripts/ci-gates-test.sh` is touched. The new section is inserted immediately after the `require-jobs-succeeded` block rather than appended at the end, to keep clear of other work landing today.

---


iamsamwood (approved):

---

Issue #514, opened 2026-08-21
Distributed eval-path timing drag after per-env formals (#513): characterize, then reclaim via #379 interning


sam-at-luther:
Follow-up recorded at merge time for #513 (the #363 fresh-formals fix), so the cost story stays measured rather than remembered.

## What CI observed that local measurement could not

PR #513's benchmark comparison (interleaved arms, same runner, n=10 — [the gate report](https://github.com/luthersystems/elps/pull/513)) showed **35 statistically significant timing regressions in `lisp`**, all below the 15% gate, **geomean +1.95%**:

| row | delta |
|---|---|
| `EnvFunCallKey` | +8.22% (p=0.000) |
| `EnvGet` | +7.08% (p=0.000) |
| `EnvFunCallVar` / `TailRec` / `Pos` / `Opt` | +5.3–6.7% |
| `FormatString*`, `Alias*` clusters | +2.5–7% |

The authoring measurement said the eval path was flat; that box drifts ±30% between sessions and could not resolve deltas of this size. CI can, and did.

## Why this is probably two effects, not one

Two signatures point away from "the copy is on the hot path" (it is not — allocation metrics are **byte-identical, all samples equal, on every regressing row**):

1. **GC scan tax**: each env now keeps ~56–96 KiB of extra pointer-dense live formals; every GC cycle during a benchmark scans it. The #513 authoring data showed `GOGC=off` removes roughly half the env-construction wall-clock cost, consistent with a GC component.
2. **Code layout**: a new file (`lisp/defformals.go`) shifts function addresses, and the row pattern is *mixed* — `EnvFunCallBuiltin` got **7.2% faster**, `AliasCDR` **7.9% faster**, with identical allocation profiles. Swings in both directions at ±8% with no allocation change is the classic instruction-cache-alignment signature in Go microbenchmarks.

How much of the +1.95% geomean is each effect is exactly what needs characterizing before anyone tries to "fix" it.

## Plan

1. **Characterize**: an A/B with layout controlled (e.g., pad or merge `defformals.go` into `env.go` on a throwaway branch; and a `GOGC=off` arm) to split layout noise from GC scan tax. A null comparison (same commit both arms) on the CI runner bounds the floor.
[...]

PR #530, opened 2026-08-24
libjson: cut the encoder's allocation cost (#379 item 6)

sam-at-luther:
Refs #379 (item 6: "libjson allocation profile ... buffer reuse / streaming paths are the classic wins").

Experiment-first: profile the libjson benchmarks, then implement only what the profile named.

## The profile

`-memprofile` over the heavy shapes — `BenchmarkPackage/{$load,load-github,dump-github,load-nested,dump-nested}` plus `BenchmarkEncodeOwnMessage*` — 17.9 GiB and 179M objects total.

| site | alloc_objects | alloc_space |
|---|---|---|
| `lisp.String` (inline) | 24.0% | 25.6% |
| `encoding/json.unquote` | 20.3% | 6.8% |
| `lisp.QExpr` (inline) | 15.0% | 16.1% |
| **`libjson.SortedMap.Entries`** (flat) | **13.6%** | 2.1% |
| `encoding/json.(*decodeState).literalInterface` | 10.0% | 1.5% |
| **`bytes.growSlice`** | 0.9% | **17.5%** |

Attributing the two inlined leaves to their callers is what makes the picture legible:

* **`sortedMapEntries` → `Map.Entries` is 44% of all objects and 34% of all bytes**, and 89% of that is `libjson.SortedMap.Entries`. Almost all of `lisp.String` and `lisp.QExpr` above are its per-entry boxing.
* **`bytes.growSlice` is 17.5% of all bytes**, all of it the encoder's output buffer.
* On the shape substrate actually runs (`BenchmarkEncodeOwnMessageMedium`, a JSON-RPC envelope around one `json:dump-message` native), the split is `bytes.growSlice` 52% / `encoding/json.Marshal` 41%.

Everything the *decode* side spends is `encoding/json`'s own `unquote` and map growth plus one `lisp.String` per decoded string — irreducible without replacing `encoding/json`. **The load path is not touched, and every load row below is unchanged on all three metrics. That is the control.**

## The three changes

**1. Map entry boxing** — 40% of all objects, 31% of all bytes. Both sorted-map implementations built three heap objects per entry (the pair `LVal`, its two-element `Cells` slice, the key `LVal`) for entries the encoder reads once and drops on the floor. There was even a `TODO: Cache map entries slices` sitting on `encodeSortMap`.

Those are now carved out of arrays sized once from `Len()`: `lisp.sortedmap` costs **3 allocations for a map of n entries instead of 3n**; `libjson.SortedMap` costs **n+2** (`lisp.QExpr` is the only way to set the `quoted` flag from outside package `lisp`, so the pair `LVal` stays per-entry).

Nothing observable changes: each pair is still a distinct quoted `LVal` whose `Cells` slice is **capped to its own two slots**, and each key is still a distinct `LVal`. `elpstest.AssertSortedMap` grew a conformance case for exactly that cap — a pair carrying spare capacity would let an append on one entry overwrite the next entry's key, corruption no other assertion in that file would catch. It runs against both implementations.

**2. The output buffer** — 17.5% of all bytes. A `bytes.Buffer` doubled its way up from empty on every encode, for a buffer then discarded. Encoders are pooled. `json:dump-string` — the busiest entry point, and the one every `dump-*` row goes through — converts the pooled buffer straight to its string and keeps the buffer, leaving one allocation of exactly n bytes (the string, which is what the caller asked for).

The paths whose `[]byte` escapes **donate** the array instead and put the encoder back with an empty one. Copying to keep the buffer there was measured and is worse: it only pays for itself if the recycled buffer survives, and on exactly the large documents where recycling would matter most it does not, because a process allocating hundreds of KB per encode empties the pool at every GC. Measured on `EncodeOwnMessageLarge`, copying cost +1.6% B/op; donating costs nothing and still pools the encoder struct.

**3. `json.Marshal` on the native path** — 41% of the bytes on the per-response shape. `Marshal` ends with `append([]byte(nil), e.Bytes()...)` to hand its caller an owned slice, and this package's only use for that slice was `enc.buf.Write(b)`. `json.Encoder` writes from the same pooled `encodeState` directly into the buffer with the same options (`NewEncoder` defaults `escapeHTML` to true, which is what `Marshal` hard-codes), so the copy disappears. The trailing newline `Encode` adds is truncated, and `checkLoadable` now inspects the bytes in place and rolls back on failure — both invisible, since an encode that errors discards its buffer.

## Results

Interleaved benchstat, n=12, alternating prebuilt base/PR binaries:

```
                          │     base     │                   pr                   │
                          │     B/op     │     B/op      vs base                  │
Package/dump-github         51.03Mi ± 0%   36.33Mi ± 0%  -28.81% (p=0.000 n=12)
Package/dump-array          807.6Ki ± 0%   635.9Ki ± 0%  -21.26% (p=0.000 n=12)
Package/dump-nested         2.835Mi ± 0%   2.606Mi ± 0%   -8.05% (p=0.000 n=12)
Package/dump-object         1.735Mi ± 0%   1.598Mi ± 0%   -7.89% (p=0.000 n=12)
EncodeOwnMessageLarge       628.0Ki ± 2%   307.1Ki ± 3%  -51.09% (p=0.000 n=12)
EncodeOwnMessageMedium      33.72Ki ± 2%   17.64Ki ± 3%  -47.69% (p=0.000 n=12)
EncodeOwnMessageSmall       2.517Ki ± 0%   1.812Ki ± 0%  -27.98% (p=0.000 n=12)
EncodeNativeSmall             892.0 ± 1%     769.0 ± 2%  -13.79% (p=0.000 n=12)
EncodeNativeLarge           45.65Ki ± 0%   41.56Ki ± 0%   -8.95% (p=0.000 n=12)
Encode                      11.32Ki ± 0%   11.30Ki ± 0%   -0.18% (p=0.000 n=12)
Encode_stringNumbers        1.297Ki ± 0%   1.313Ki ± 0%   +1.24% (p=0.000 n=12)
Package/$load               824.1Ki ± 0%   824.2Ki ± 0%        ~ (p=0.235 n=12)
Package/load-github         32.74Mi ± 0%   32.74Mi ± 0%        ~ (p=0.702 n=12)
Package/load-nested         2.154Mi ± 0%   2.154Mi ± 0%        ~ (p=0.989 n=12)
Package/get-nested-baseline 36.53Mi ± 0%   36.53Mi ± 0%        ~ (p=0.551 n=12)
Load/*, LoadIntegers/*                                    ~ (all samples are equal)
geomean                     147.5Ki        130.7Ki       -11.42%

                          │  allocs/op   │  allocs/op    vs base                  │
Package/dump-github          387.5k ± 0%   150.5k ±  0%  -61.15% (p=0.000 n=12)
Package/dump-nested          37.06k ± 0%   22.06k ±  0%  -40.47% (p=0.000 n=12)
Package/dump-object          23.04k ± 0%   15.04k ±  0%  -34.71% (p=0.000 n=12)
Package/dump-array          10.033k ± 0%   8.034k ±  0%  -19.92% (p=0.000 n=12)
EncodeOwnMessageSmall/Medium  17.00 ± 0%     9.00 ±  0%  -47.06% (p=0.000 n=12)
EncodeOwnMessageLarge         18.00 ± 6%    10.00 ± 10%  -44.44% (p=0.000 n=12)
EncodeNativeSmall             17.00 ± 0%    15.00 ±  0%  -11.76% (p=0.000 n=12)
Encode                        231.0 ± 0%    226.0 ±  0%   -2.16% (p=0.000 n=12)
every load row                                                  ~
geomean                       1.447k        1.200k       -17.08%
```

Timing moves the same way where it moves at all — `Package/dump-github` **-22.0%** (p=0.000), `EncodeOwnMessageLarge` **-13.4%** (p=0.000), `EncodeOwnMessageMedium` **-8.3%** (p=0.002), geomean -2.1%. No timing regression; the rest is sandbox noise.

Two bad-direction moves, both accounted for:

* `Encode_stringNumbers` **B/op +1.24%**, allocs/op unchanged. Batching n `LVal`s into one array can round up one size class where n separate ones did not (3×112 = 336 → the 352 class). Bounded at 16 bytes per array, and it is exactly what buys the allocation counts above.
* `Load/exactIntegers` showed **sec/op +30.75%** in the table above and `~ (p=0.977)` in an earlier identical run. It is a 16 µs pure-decode row whose B/op and allocs/op are byte-identical ("all samples are equal") and whose code is untouched. Re-measured in isolation at `-benchtime 20000x`, n=15: `16.48µ ± 5% → 17.41µ ± 7%, ~ (p=0.217)`. The sandbox artifact is `-benchtime 30x` = 0.5 ms of measurement per sample; CI's `-benchtime=100ms` gives that row ~6000 iterations.

The `lisp` package's own benchmarks were checked separately for fallout from the `maps.go` change: `Equal/{map,vector,nested}` and `String/{map,vector,nested}` go **-46% / -24% allocs/op** for **+3.2% B/op** (the same size-class rounding). The one row the gate flagged there, `PackageGetFunParallel B/op +23.5%`, is `-benchtime 100x` setup noise — at `-benchtime 20000x`, n=15 it is `0.000 B/op` on both arms, all samples equal.

## New guards

* **Pooled-encoder state hygiene** — a stale `stringNums` changes a document's bytes; a stale `nestedDeep`/`wroteNative` silently mislabels `json:dump-message`'s loadability verdict, which is the quiet one. Red-proofed: deleting any single reset in `getEncoder` fails a case.
* **Donated buffers do not alias** — a later dump must not write through a slice an earlier caller still holds.
* **Batched struct literals still match their constructors** — the literals in `Entries` are copies of `lisp.String`/`lisp.QExpr` bodies, so both are pinned with `reflect.DeepEqual`, which reads unexported fields and therefore fails on a field libjson could not have seen.
* `TestEncodeDoesNotAllocateForCycleTracking`'s exact counts each drop by exactly one — the pooled encoder struct — and no further, because every map in that test has one entry, where batching three objects into three arrays is a wash.

## Test plan

* `go test ./...` — pass
* `go test -tags elpscheck ./lisp/...` — pass
* `go test -race ./lisp/lisplib/libjson ./lisp` — pass
* `make elpsvet` — pass (the one deliberate mutation in the new `elpstest` conformance case carries an `//elps:mutates` justification)
* `make static-checks` — only the two pre-existing `parser/token/token.go` nolintlint findings
* `./elps fmt -l ./...` clean; `./elps lint ./...` byte-identical to base; `./elps doc -m` clean
* `scripts/confidentiality-guard.sh` — clean
* `-fuzz` 45 s each on `FuzzDumpJSON`, `FuzzCheckLoadableMatchesLoad`, `FuzzDumpExactIntegers` — no crashers (these cover the native/loadable paths change 3 touches)

No change to JSON semantics, error messages, or the public API.

https://claude.ai/code/session_013WFdAfZ8LNMmGBockuMuJq

---
_Generated by [Claude Code](https://claude.ai/code/session_013WFdAfZ8LNMmGBockuMuJq)_

### DIFF
--- a/elpstest/sortedmap.go
+++ b/elpstest/sortedmap.go
@@ -25,6 +25,9 @@ import (
 //	Calling m.Get() with a key from m.Entries() returns a value consistent
 //	with that entry.
 //
+//	Each pair from m.Entries() owns its two cells: appending to one pair
+//	cannot reach into another's storage.
+//
 // AssertSortedMap does not test any of the following properties:
 //
 //	Success/Failure of insertions or deletions -- m must already be
@@ -51,6 +54,46 @@ func AssertSortedMap(t *testing.T, m lisp.Map) bool {
 	}
 	if !testGetEntries(t, m) {
 		return false
+	}
+	if !testEntryPairsAreIndependent(t, m) {
+		return false
+	}
+	return true
+}
+
+// testEntryPairsAreIndependent is the conformance test for an optimisation
+// both in-tree implementations now make (issue #379, item 6): a map of n
+// entries carves its n two-element Cells slices out of ONE backing array
+// rather than allocating n of them.  That is invisible only for as long as
+// every pair's slice is capped to its own two slots.  A slice that carries
+// spare capacity lets an append on one pair overwrite the next pair's key,
+// which is a data-corruption bug no other assertion here would catch --
+// entries would still read back correctly until something appended.
+func testEntryPairsAreIndependent(t *testing.T, m lisp.Map) bool {
+	t.Helper()
+	entries, err := mapEntries(m)
+	if !assert.NoError(t, err) {
+		return false
+	}
+	before := make([]*lisp.LVal, len(entries.Cells))
+	for i, pair := range entries.Cells {
+		if !assert.Equal(t, 2, cap(pair.Cells),
+			"entry %d has spare capacity, so appending to it reaches storage it does not own", i) {
+			return false
+		}
+		before[i] = pair.Cells[0]
+	}
+	// Appending to every pair in turn must leave every other pair's key
+	// exactly where it was.
+	for i := range entries.Cells {
+		//elps:mutates the pairs were built by the m.Entries call above into a buffer this function owns, and appending to them IS the property under test -- it is how a pair that shares storage with its neighbour is detected
+		entries.Cells[i].Cells = append(entries.Cells[i].Cells, lisp.Int(i))
+	}
+	for i, pair := range entries.Cells {
+		if !assert.Same(t, before[i], pair.Cells[0],
+			"entry %d's key was overwritten by an append to another entry", i) {
+			return false
+		}
 	}
 	return true
 }
--- a/lisp/constructors_test.go
+++ b/lisp/constructors_test.go
@@ -2,7 +2,10 @@
 
 package lisp
 
-import "testing"
+import (
+	"reflect"
+	"testing"
+)
 
 // noopBuiltin is an LBuiltin used as a stand-in body for constructor
 // tests — none of the cases below invoke it.
@@ -65,4 +68,41 @@ func TestFunInPackage_SetsPackage(t *testing.T) {
 			t.Errorf("Package() = %q, want empty", got)
 		}
 	})
+}
+
+// TestBatchedEntryValuesMatchTheirConstructors is the guard on sortedmap's
+// entry batching (issue #379, item 6).
+//
+// Entries builds its pair and key LVals as struct literals carved out of
+// arrays instead of calling QExpr and String n times, because the per-entry
+// allocation was the single largest object-count site in the libjson
+// benchmark suite.  Those literals are copies of the constructors' bodies, so
+// they are correct only for as long as the constructors stay what they are
+// today.  A field added to String or QExpr and not added here would produce
+// map entries that differ from every other value of their type in a way
+// nothing else would notice.
+//
+// reflect.DeepEqual reads unexported fields, so the comparisons below cover
+// every field of the struct, not just the ones a literal names.
+//
+// Red-proof: dropping `quoted: true` from the pair literal in
+// sortedmap.Entries, or adding an initialised field to String, fails this.
+func TestBatchedEntryValuesMatchTheirConstructors(t *testing.T) {
+	for _, s := range []string{"", "k", "a key", "\x00<&>"} {
+		batched := LVal{Type: LString, Str: s}
+		if want := String(s); !reflect.DeepEqual(&batched, want) {
+			t.Errorf("a batched LString for %q is not what String builds:\n got %#v\nwant %#v",
+				s, batched, *want)
+		}
+	}
+
+	cells := []*LVal{String("k"), Int(1)}
+	batchedPair := LVal{Type: LSExpr, quoted: true, Cells: cells}
+	// mklist copies its arguments into a slice of its own, so the two pairs
+	// hold different slices with equal contents; DeepEqual compares the
+	// contents, which is what matters.
+	if wantPair := mklist(cells[0], cells[1]); !reflect.DeepEqual(&batchedPair, wantPair) {
+		t.Errorf("a batched entry pair is not what mklist builds:\n got %#v\nwant %#v",
+			batchedPair, *wantPair)
+	}
 }
--- a/lisp/lisplib/libjson/encode.go
+++ b/lisp/lisplib/libjson/encode.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"math"
 	"strconv"
+	"sync"
 	"unicode/utf8"
 
 	"github.com/luthersystems/elps/lisp"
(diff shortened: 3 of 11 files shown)


sam-at-luther:
Reconciling the PR description with the CI benchmark run: the description says "two bad-direction moves", but the CI gate reports **25** (still 0 at or above the gate). They are not two different stories — the CI run simply exposes the batching trade on rows the local `./lisp/`-scoped comparison didn't cover.

Every one of the extra rows is the same shape: **a few bytes of size-class rounding bought with a large drop in allocation count.**

| row | B/op | allocs/op | bytes per allocation removed |
|---|---|---|---|
| `Equal/map` | +63 B (+3.23%) | 26 → 14 (**−46%**) | 5.3 |
| `String/map` | +32 B (+2.48%) | 25 → 19 (−24%) | 5.3 |
| `libelpspath TypeCheck` | +129 B (+2.15%) | 78 → 42 (**−46%**) | 3.6 |
| `libelpspath BuiltinSetCopy` | +256 B (+1.60%) | 231 → 159 (−31%) | 3.6 |

Batching *n* entry objects into one array sized from `Len()` rounds that array up to a size class, so total bytes tick up by a small constant while object count falls sharply. At 3.6–5.3 bytes per object removed the trade is strongly favourable — a Go heap object costs more than that in header and GC scan work alone, which is the same lesson #514 landed on (object count and pointer density dominate raw bytes).

The upside is also wider than the description claims. `sortedMapEntries` is not libjson-only, so **`libelpspath` gets −17.92% allocs/op and −4.47% sec/op geomean** without touching that package, and the `lisp` allocs/op geomean moves −5.63%.

One row does want a second look before merge: **`Encode_stringNumbers` sec/op +6.57% (p=0.000)** — the description diagnoses its B/op (+1.24%, size-class rounding) but not its timing. Below the 15% gate and on a 2.7 µs benchmark, so plausibly the same rounding plus layout effects, but it is deterministic rather than noise.

---

sam-at-luther:
Answering my own question above: **`Encode_stringNumbers` was not a regression — it was a benchmark artifact, and it is now a 19% win.** Pushed as `5709f2b`.

`testEncode` and `testEncode_stringNumbers` took an encoder from the pool and dropped it:

```go
enc := getEncoder(true)
... assertions ...
// no putEncoder
```

So every iteration missed an always-empty pool: `sync.Pool` lookup, a fresh allocation through `New`, the field resets — and never once the reuse the pool exists for. No caller behaves that way (`dumpString` does `defer putEncoder`, `dump` calls it explicitly), so the two benchmarks running through these helpers were measuring a lifecycle that does not occur in the language.

Reproduced the CI move locally first on a different CPU (+7.65%, p=0.000, n=12 — CI saw +6.57%), so it was deterministic rather than runner noise, then added the missing `putEncoder` and re-measured. Interleaved, n=12, `benchtime=300000x`:

| benchmark | before | after |
|---|---|---|
| `BenchmarkEncode` | +1.50% | **−16.40%** (p=0.000) |
| `BenchmarkEncode_stringNumbers` | +6.57% | **−19.01%** (p=0.000) |

The assertions are untouched and the string is materialized before the encoder goes back, so this changes what the benchmarks *measure*, not what the tests *check* — and what they measure now is the path `json:dump-string` actually takes.

Worth stating plainly since it cuts the other way too: this means the encoder-pooling win is **larger** than the PR describes, not smaller. The remaining below-gate B/op rows are the size-class trade in the comment above, which stands as described.

`go test`, `-race`, and `-tags elpscheck` on `./lisp/lisplib/libjson/` are all clean on the new head; `gofmt` clean.

---

PR #536, opened 2026-08-24
feat(lisp): sealed per-file load cache hook (Runtime.LoadCache)

### LINKED ISSUE #368: Provide a sealed per-file load-file cache hook so embedder parse caches never hold raw ASTs
Follow-up to the parse/cache boundary work. The remaining hard seam is `load-file`.

## The seam

`Reader` and `LocationReader` are the only interfaces an embedder can implement to influence parsing, and both return a raw slice of AST nodes (`lisp/loader.go:14-27`):

```go
type Reader interface {
	Read(name string, r io.Reader) ([]*LVal, error)
}

type LocationReader interface {
	ReadLocation(name string, loc string, r io.Reader) ([]*LVal, error)
}
```

`load-file` reaches them through `Runtime.Reader`: `builtinLoadFile` (`lisp/builtins.go:462`) calls `LEnv.LoadFile`, which calls `env.Runtime.Reader.Read(name, r)` and evaluates the result directly (`lisp/env.go:238-300`, `1251-1300`).

So an embedder that wants a parse cache has exactly one place to put it, and that place requires it to take custody of raw `*LVal` AST nodes and hand the same nodes to every environment. That is not a misuse of the API. It is the only shape the API offers.

`luthersystems/substrate` does precisely this: `cacheReader` implements both interfaces over a process-global hash to `[]*lisp.LVal` map and returns the cached slice aliased, with no copy (`internal/substrate/elpsutil/elpsutil.go:60-110`). Tracked on that side as luthersystems/substrate#375.

## Why lisp.Program does not close it

`lisp.Program`, added on the unmerged branch `claude/lval-hermetic-full` in commit [d1d8a8b](https://github.com/luthersystems/elps/commit/d1d8a8b), seals parse output behind an opaque type whose only exit is `Detach()`, with producers (`ReadProgram`, `ReadLocationProgram`, `(*LEnv).ParseProgram`) placed so that raw-slice custody never leaves package `lisp`. It deliberately adds no `NewProgram([]*LVal)`, on the grounds that a seal whose caller keeps the raw slice seals nothing.

That covers the parse-once/load-many entry point an embedder calls directly. It does not cover `Runtime.Reader`, which is the seam `load-file` drives, and `load-file` is how a phylum actually loads its own sources. The `Program` commit message says as much: full sealing is `Program` at the boundary plus the eval-side leak seals, and the `Reader` path is untouched by design because the change was additive.

## What is wanted

An elps-owned, sealed hook for per-file `load-file` results, so that an embedder supplies a cache policy (key, lookup, store, eviction) without ever receiving or retaining `[]*LVal`. Sketch of the requirements rather than a design:

- The cached object is opaque to the embedder, in the way `Program` is opaque.
- Handing a cached entry to an environment is elps's decision, so elps can choose aliasing or copying per environment and change that choice later without an embedder migration.
- `Reader`/`LocationReader` stay for embedders that genuinely want to replace the parser, but they stop being the only way to get a cache.

The payoff is that the cache aliasing class becomes an elps-internal decision, enforceable by construction inside the one repository that owns the AST type, instead of a convention each embedder has to keep.

## Related

- luthersystems/substrate#375, the embedder-side instance and its measured constraints.
- #362 and #363, the value-level sharing hazards this seam multiplies.

sam-at-luther:
Closes #368.

## What this is

`lisp.Program` closed the parse/cache boundary an embedder drives *directly*. It leaves `Runtime.Reader` untouched by design, and `Runtime.Reader` is the seam `load-file` drives — which is how a phylum loads its own sources. An embedder that wants a per-file parse cache has had exactly one place to put it, and that place requires it to take custody of `[]*lisp.LVal` and hand the same nodes to every environment.

`Runtime.LoadCache` is the elps-owned hook for that path. The embedder supplies **policy** and never touches **data**:

```go
type LoadCache interface {
        Load(key string) (*CachedSource, bool)
        Store(key string, src *CachedSource)
}
```

`*CachedSource` is opaque exactly as `Program` is — only elps mints one, and no exported member yields a `*LVal`. Every `Load*` entry point funnels through `(*LEnv).readCached`:

- **miss** — read the stream, derive the key, parse, admit through **`newProgram`** (the same single admission point the `Program` constructors use), `Store`, hand the sealed tree to this load. The first loader gets the cached nodes too; keeping a pristine copy back is exactly the cost this hook removes.
- **hit** — hand that same sealed tree to the next environment **by reference**.

`Reader`/`LocationReader` are untouched; the hook composes in front of them. A nil `LoadCache` is byte-identical to the historical path — the reader gets the caller's own `io.Reader`, and the funnel allocates nothing.

New files: `lisp/loadcache.go` plus five test files. Modified: `lisp/env.go` (four read sites), `lisp/runtime.go` (one field), `lisp/export_test.go` (test bridges), `lisp/ownership_check_elpscheck.go` (allowlist commentary), `docs/sealed-ast.md` §2.9, `docs/embed.md`, `CLAUDE.md` (fuzz inventory).

## The refund, measured

`lisp/loadcache_bench_test.go`, three arms plus an `env-only` control, one machine, one process, `-benchtime=100ms -count=10`, measured on this branch's merged head.

`benchstat -col /arm`, base = the arm an embedder can actually write:

```
                                 │ reader-cache-copy │           load-cache-alias           │
                                 │      sec/op       │    sec/op     vs base                │
LoadIntoEnv/size=small-1KiB-4            212.0µ ± 7%   144.5µ ± 11%  -31.86% (p=0.000 n=10)
LoadIntoEnv/size=library-78KiB-4         5.891m ± 5%   2.009m ±  5%  -65.90% (p=0.000 n=10)
geomean                                  1.118m        538.7µ        -51.80%

                                 │ reader-cache-copy │           load-cache-alias           │
                                 │       B/op        │     B/op      vs base                │
LoadIntoEnv/size=small-1KiB-4           226.2Ki ± 0%   147.3Ki ± 0%  -34.88% (p=0.000 n=10)
LoadIntoEnv/size=library-78KiB-4        5.429Mi ± 0%   1.605Mi ± 0%  -70.43% (p=0.000 n=10)
geomean                                 1.095Mi        492.0Ki       -56.12%

                                 │ reader-cache-copy │          load-cache-alias           │
                                 │     allocs/op     │  allocs/op   vs base                │
LoadIntoEnv/size=small-1KiB-4            2.202k ± 0%   1.284k ± 0%  -41.69% (p=0.000 n=10)
LoadIntoEnv/size=library-78KiB-4         64.31k ± 0%   18.70k ± 0%  -70.92% (p=0.000 n=10)
geomean                                  11.90k        4.900k       -58.82%
```

All four arms, absolute (`env-only` is environment construction with no load — subtract it to see the load by itself):

| arm | 78 KiB sec/op | B/op | allocs/op |
|---|---|---|---|
| env-only (control) | 0.105 ms | 118 KiB | 915 |
| no-cache | 16.89 ms | 9169 KiB | 235,036 |
| reader-cache-copy (embedder-shaped) | 5.89 ms | 5559 KiB | 64,308 |
| load-cache-alias (this hook) | 2.01 ms | 1644 KiB | 18,700 |

The small-file arm is not decoration: the hook's own cost is a SHA-256 of the source plus a map probe, and 1 KiB is where a cache could plausibly lose. It does not — it still beats both other arms.

`TestBenchmarkArmsAgree` is the benchmark's anti-vacuity gate: all three arms must reach the same environment state, and the cached arms must parse exactly once. A benchmark arm that silently failed its load would otherwise post a beautiful number for doing nothing.

## Plan premises that were wrong, with evidence

Three, and two of them changed the work.

**1. "Extend the admission mechanism used by shared sealed formals and `AddPackage`'s sealed-singleton class." No new admission class is needed, and adding one would have been wrong.**

`checkOwnership` already exempts sealed nodes outright (`lisp/ownership_check_elpscheck.go`):

```go
if v == nil || rt == nil || isSingleton(v) || v.sealed {
    return
}
```

Allowlist entry 2 names the topology this PR implements as the *first* of its sanctioned ones — "substrate's warm parse cache hands one sealed tree to many runtimes". Because `newCachedSource` routes every parse through `newProgram`, a cache entry is sealed throughout by construction and is already inside that exemption. The only change to that file is a line added to its list of sanctioned topologies, per the file's own rule that a deliberate design gets documented there. Both directions are still driven, in `lisp/loadcache_elpscheck_test.go`.

**2. "The debugger stamps macro-expansion metadata onto macro-expanded nodes; if it can write into shared cached nodes, hand the env a private `Detach()`-style copy." It cannot, so there is no fallback and debug mode pays nothing.**

Where the stamping writes: `stampGuarded` (`lisp/macro.go`), and it returns before any write at the first sealed node —

```go
if v.sealed {
    return
}
...
if v.source == nil || v.source.Pos < 0 {
    v.source = callSite
    if ctx != nil {
        v.macroExpansion = &macroExpansionInfo{...}
    }
}
```

A sealed node's descendants are all sealed, so the whole subtree is skipped. The debugger packages themselves only *read* expansion metadata (`engine.go:997`, `inspector.go:126`); the other debug-mode change, globally disabling TCO, is stack behaviour and writes no nodes.

There is a second, independent guard: rdparser gives every node it emits a real location, so the stamp's `source == nil || Pos < 0` condition is false for an ordinary parse whether or not the seal exists. That makes an end-to-end test over ordinary source *insensitive* to the guard actually under discussion, which I checked rather than assumed — see the red proofs below. `TestLoadCacheDebuggerDoesNotStampSharedNodes` therefore uses a `Reader` that emits a location-less node (legal: the `Reader` interface promises nothing about locations, and the cache admits whatever it returns), which is the case where the sealed skip is the only guard left.

**3. Keying: "substrate keys its cache by a hash of the code — mirror that so migration is natural." Mirrored in shape, deliberately not in content.**

`parseCache` keys on `sha512(content)` alone and ignores name and location. That is safe for its topology (a file's name and location are stable across loads) but not in general: two files with identical text share an entry, and the served tree carries the *first* file's parse locations, so every error raised from the second names the wrong file. elps digests the content **and** the name **and** the location, each length-prefixed so no concatenation of one tuple can collide with another. On the motivating topology this costs nothing and hits identically. `TestLoadCacheKeySeparatesIdentityFromContent` pins it.

Two smaller corrections: the issue's line numbers are stale throughout (the funnel is `LoadFile` → `LoadLocation`/`LoadLocationContext` → `Reader`, at `lisp/env.go:339/383/1566/1585`), and the downstream `cacheReader` no longer aliases at all — its `Get` returns a detached copy, which is what turns this issue from a correctness argument into a cost argument, as the status comment says.

## Red proofs, each demonstrated failing and restored

**(a) An unsealed node smuggled into a cached tree trips the ownership checker.**

`TestLoadCacheUnsealedEntryTripsOwnership` (`-tags elpscheck`) builds an entry via a test-only bridge that skips the admission walk, loads it in runtime A, then in runtime B. Sensitivity check — mint the same entry through the real `newCachedSource` instead:

```
--- FAIL: TestLoadCacheUnsealedEntryTripsOwnership
    expected an ownership-violation panic, got none — the gate cannot fail
```

So the violation is present exactly when the admission is bypassed, and absent when it runs. The green half (`TestLoadCacheSealedEntryIsExemptFromOwnership`) proves a legally admitted entry serves two runtimes without tripping — without it, an embedder on the supported topology could not use checked mode at all.

**(b) A lisp-level write through the shared literal raises `modify-literal-error`, and the other environment sees nothing.**

`TestLoadCacheWriteThroughSharedTreeIsRefused`. With the sealed guard in `builtinSortStable` neutered:

```
--- FAIL: TestLoadCacheWriteThroughSharedTreeIsRefused
    Error: Not equal: expected: "'(3 1 2)"  actual: "'(1 2 3)"
    Messages: the other environment must not observe the attempted write
    Error: Not equal: expected: 0x615dbe298191e59b  actual: 0xcab760edc51a5095
    Messages: the shared sealed tree changed; a guard failed and one
              environment rewrote every other one's program
```

Restored, it passes, and the sanctioned remedy `(stable-sort < (copy (literal)))` works in both environments and leaves the shared tree alone.

**(c) The debugger stamp.** `TestLoadCacheDebuggerDoesNotStampSharedNodes`, with `if v.sealed { return }` removed from `stampGuarded`:

```
--- FAIL: TestLoadCacheDebuggerDoesNotStampSharedNodes
    Error: Not equal: expected: 0x380307148dd00dd1  actual: 0xda53008353e84a7c
    Messages: macro expansion (debugger=true) wrote into the SHARED cached tree
    Messages: macro expansion (debugger=false) wrote into the SHARED cached tree
```

The first version of this test used ordinary parsed source and did **not** fail under that mutation — the parser's real locations were carrying it. That is why it now uses a location-less node, and why the honest claim is "two independent guards, one of which the test isolates".

**(d) The fuzz target is non-vacuous.** With the hit branch of `readCached` disabled, `FuzzLoadCacheMultiEnv` fails on its first seed: `the source was parsed 2 times across 2 environments; a shared cache must parse it once`.

**(e) `CachedSource`'s opacity guard is non-vacuous** by construction: `TestCachedSourceIsOpaque` asserts the expected accessor surface exists as well as that none of it reaches a `*LVal`, so a mass deletion cannot pass as "no leaks found".

## Test plan — real results

Re-run on the merged head after taking `origin/main` (v1.53.0 + #535) in.

| gate | result |
|---|---|
| `go build ./...` | ok |
| `go test ./...` | all packages ok |
| `make test` (go test -cover ./... + example lisp files) | all packages ok; `lisp` 83.7% |
| `go test -tags elpscheck ./lisp/...` | all packages ok |
| `go test -race ./lisp/` | ok, 81.4s |
| `golangci-lint run ./...` (v2.6.2, CI's pin) | **0 issues** |
| `go run ./cmd/elpsvet -test=false ./...` | clean, no new suppressions |
| `./elps fmt -l ./...` | clean |
| `./elps lint --workspace=. --exclude grammar --include _examples ./...` | pre-existing warnings only, exit 0 |
| `./elps doc -m` | All functions and exports are documented |
| `scripts/confidentiality-guard.sh` | clean (743 files) |
| `go test -fuzz=FuzzLoadCacheMultiEnv -fuzztime=60s ./lisp/` | 23,442 execs, no crashers |
| same, re-run on merged head at 45s | 17,991 execs, no crashers |

golangci-lint found 5 real issues on the first pass (errorlint, nilerr, testifylint, 2× unconvert) — all fixed; the nilerr one produced `admitCachedSource`, which reads better than the version that tripped it.

## Notes for the reviewer

- **`make fuzz-budget-check` now reports "room for 0 more target(s) before this gate goes red".** The sweep still fits with 10 min headroom, but this target consumed the last slot (35 → 36 targets; it was "room for 1" before). The next fuzz target added anywhere in the repo needs `timeout-minutes` raised in `.github/workflows/fuzz.yml`. Flagging rather than pre-emptively raising it.
- **The hook sits at the read funnel, so it sees `LoadString` and `Load`, not only `LoadFile`.** That is deliberate (it matches the reach of the embedder-side caches it replaces, so migration is a straight swap) and documented on the interface, including that a host evaluating many distinct one-off strings through an unbounded cache accumulates an entry per string. Retention is the implementation's job.
- **Scope I did not take:** concurrent evaluation of one cached tree from several goroutines is untested here, exactly as it is untested in `FuzzSharedProgramMultiEnv`. Environments run one after another, which is the production topology. The `-race` watchdog covers the concurrent case for sealed nodes generally.
- **`FuzzLoadCacheMultiEnv` is a sibling of `FuzzSharedProgramMultiEnv`, not an extension of it.** They share a property but not a code path: the existing target shares a `Program` the test itself holds, so everything between "the bytes" and "the expressions the evaluator sees" — key derivation, admission, store, hit, alias-vs-copy — is stubbed out there and is the whole subject here. It reuses that target's seed corpus, its determinism control, its budgeted harness and its divergence-confirmation discipline.
- **Nothing here changes `Reader`, `LocationReader`, `Program`, or any existing behaviour with `LoadCache` unset.**

https://claude.ai/code/session_013WFdAfZ8LNMmGBockuMuJq


sam-at-luther:
## Adversarial review (custody / API lens): would not merge as-is

An independent adversarial pass broke the admission boundary five ways, each with a runnable repro against head `375fa62`. Recording here so the findings are tracked; a second pass (evaluation-semantics lens) is still running and will be appended. **Holding this PR until these are fixed with a red-proof per finding.** The design shape is sound — concurrency, immutability, opacity, lifetimes, and key-*mismatch* trust all held — but the admission is not yet the boundary the design claims.

### Must-fix before merge

1. **Admission trusts the seal *flag*, not the node *type* — and `-tags elpscheck` shares the blind spot.** `program.go:231` (`firstUnsealed`) and `ownership_check_elpscheck.go:237` both key off `sealed`/`v.sealed` alone. A closure minted in Runtime A, laundered into a cache entry, invoked from Runtime B, ticks its captured counter `1→2→3` across two runtimes sharing one `*LEnv` — the exact cross-runtime sharing the admission exists to stop — and the checked build *exempts* the laundered node for the same untrusted-flag reason. Fix: conjoin `sealableNodeType(v.Type)` in both predicates.

2. **The key identifies the input bytes; the entry is the reader's output; nothing binds the two.** Two Runtimes with different `Reader`s sharing one cache serve each other's parses; a single Runtime whose `Runtime.Reader` is swapped between loads re-serves the stale parse; and elps's own `Load` (`Read`) vs `LoadLocation` (`ReadLocation`) collide on the same `(name,"",src)` key. Fix: fold a reader identity/epoch and a Read-vs-ReadLocation discriminator into the digest, or record the producing reader on the entry.

3. **The admission walks are unbounded, and this PR is what puts them on the `load-file` path.** `loader.go:121` and `firstUnsealed` have no cycle guard, depth cap, or memo: a cyclic reader tree → unrecoverable `stack overflow`; an interned-subtree reader → exponential (measured 5.2 s at sharing-depth 28), both only with a cache installed. `sealfp.go:158` already models the node/depth budget to copy.

### Contract text / release note (before embedders build on the wording)

4. **Installing a cache can change lisp semantics.** For a non-sealing Reader, admission's copy path runs `SealAST`, so guarded mutation sites begin raising `modify-literal-error` on code that worked cache-less (repro: `(stable-sort < (lit))` with `WithFormatPreserving`). This is a migration hazard, not a doc nit — deserves a release note. Also: any wrapping Reader that synthesizes even one node forces the whole file down the copy-and-seal path, so the "zero-copy hit" story is conditional.

5. **`io.ReadAll` changes the outcome of erroring/streaming readers** (`loadcache.go:312`) — a stream that delivers a full program then a non-EOF error succeeds cache-less, fails with a cache. Undocumented, cache-gated, either direction.

6. **Fast-path custody (SUSPICIOUS).** The `allSealed` branch (`program.go:205`) stores the reader's own nodes; a reader that retains its output can rewrite the cached tree afterward (checked build catches it loudly; production is silent). Either copy on the fast path too, or state in the `LoadCache` contract that a reader must not retain output, enforcement being checked-builds-only.

7. **Re-entrancy (SUSPICIOUS).** A `Load` that re-enters the load path (a warming cache) → unrecoverable `stack overflow`, or deadlock with an ordinary mutex. Every other embedder mistake degrades to "no cache"; this one kills the process. Add an in-flight guard or at minimum a contract sentence.

8. **Guest lisp mints entries** — `load-string`/`load-bytes` are builtins, so semi-trusted phylum source can populate the cache (repro: 501 entries from one guest file). The retention doc note covers host call sites only; name the guest surface.

Held clean: nil/zero-entry → miss; key-mismatch trust (no bypass); TOCTOU on key-vs-content; concurrency under `-race -tags elpscheck` including debugger-attached; entry immutability; opacity; lifetimes.

Repros preserved in the reviewer's scratch dir. The `-tags elpscheck` build verifies only "sealed bytes never change" — it is blind to findings 1, 2, 5, and the wrong-program-served class generally; the PR's VERIFICATION bullet should be narrowed to match.

---

sam-at-luther:
## Custody-review fixes pushed (`c319ba3`)

All findings from the custody/API adversarial pass are addressed on this branch, each MUST-FIX with a committed regression test shown failing on the prior head (`375fa62`) and passing after. No new PR, no merge; nothing changed with `LoadCache` unset.

### MUST-FIX

**1 — admission trusted the seal FLAG, not the node TYPE.** `firstUnsealed` (`program.go`) and the `checkOwnership` exemption (`ownership_check_elpscheck.go:236`) now conjoin `sealableNodeType(v.Type)`. A laundered `LFun` closure (flag set, type mutable) is routed to the copy path and rejected at admission, and is no longer exempt from the cross-runtime gate.
- Red-proof `TestReadCachedRejectsLaunderedNonSealableNode`: on the old `firstUnsealed` → `FAIL` ("a laundered non-sealable node was admitted into the cache (1 entries)"); after → PASS.
- Red-proof `TestLoadCacheLaunderedClosureTripsOwnership` (`-tags elpscheck`, the reviewer's closure-across-runtimes shape): on the old exemption → `FAIL` ("expected an ownership-violation panic, got none"); after → panics as required.

**2 — the key bound the bytes but not the producing reader.** `loadCacheKey` now folds in reader identity (Go type by default; optional `lisp.ReaderIdentity` hook for a reader that varies its parse behind one type) and a `Read`-vs-`ReadLocation` discriminator. Type-based identity is deliberate: per-Runtime readers of the same type still share entries (the warm-cache topology), while distinct readers/methods never serve each other's parses. Enforced entirely by elps — no embedder cooperation for the default.
- Red-proofs (each served the WRONG program on the old key, correct after): `TestLoadCacheDifferentReadersDoNotServeEachOther` (expected 2, got 1), `TestLoadCacheSwappedReaderReparses` (expected 2, got 1), `TestLoadCacheReadVsReadLocationDoNotCollide` (expected 20, got 10). Plus `TestLoadCacheReaderIdentityInterfaceSeparates` for the optional hook.

**3 — unbounded admission walks, newly on the load-file path.** `checkLoaderExpr` (`loader.go`) is now a bounded walk — cycle/sharing memo + depth cap + node budget, mirroring `sealfp.go` — refusing a cyclic, interned-shared, or over-deep tree with `errReaderTreeUnbounded`. Because it runs first, `firstUnsealed` and `Copy` only ever see a finite strict tree. Such output fails the load (it is unsafe to evaluate) rather than falling back.
- Red-proof `TestLoadCacheCyclicReaderOutputIsBounded`: old walk → `fatal error: stack overflow` (process dies); after → bounded error. `TestLoadCacheInternedSubtreeIsBounded` (depth-40 DAG): old walk → does not terminate (20s timeout); after → bounded error.

**7 — a re-entrant `Load` killed the process.** `readCached` guards with `Runtime.loadCacheActive`; a cache that warms itself is treated as a miss instead of recursing.
- Red-proof `TestLoadCacheReentrantLoadIsGuarded`: without the guard the warmer recurses unbounded (fails after ~9s); after → collapses to a single warm.

### Contract / docs

**4 (release note / migration hazard)** — installing a cache runs `SealAST` at admission, so a **non-sealing** reader's guarded mutation sites begin raising `modify-literal-error` on previously-working code, and the zero-copy hit is conditional on the reader sealing its whole output. Written up as a migration note in `docs/embed.md` and `docs/sealed-ast.md §2.9`.
**5** — `io.ReadAll` changes the outcome for a streaming reader that errors after a complete program: documented on `LoadCache` and `readCached`.
**6 (decision: contract, not copy)** — the fast path stores the reader's own nodes. Chosen the contract route ("a `Reader` must not retain and later mutate returned nodes") over copy-on-fast-path, because copying would add a deep copy to **every miss** even though no compliant reader needs it — the cost this hook exists to remove. Enforcement (checked-build fingerprint re-verification after every load) is already present; the contract is now stated explicitly.
**8** — `load-string`/`load-bytes` are builtins, so guest phylum can mint entries; named in the retention note.

**VERIFICATION narrowed** (in `loadcache.go` and `sealed-ast.md`): `-tags elpscheck` proves exactly *"the sealed bytes never change after admission"* — so no load rewrites a cached node, and (with fix 1) a laundered non-sealable node still trips the gate. It does **not** prove the *right* program was served; that wrong-program-served class is closed by the key and the flag-vs-type conjunction, not by the checked build.

### Verification (real numbers)
`go build ./...` ok · `go test ./...` all ok · `go test -tags elpscheck ./lisp/...` ok · `go test -race ./lisp/` ok (104s) · `make test` ok · `golangci-lint run ./...` **0 issues** · `elpsvet` clean · `elps fmt -l` clean · `elps doc -m` "All functions and exports are documented" · confidentiality guard clean (743 files) · `make fuzz-budget-check` unchanged (room for 0, still fits; **no fuzz target added**).

One note on fuzzing: a brief `FuzzLoadCacheMultiEnv` run flagged a *slow* seed under CPU contention — the mutator's runaway variant of the committed `tail-recursion-bounded` seed (`(countdown …)` with `(- 1)` looping to `MaxTailIterations`, ~8s). It passes deterministically and is a pre-existing eval-speed property of the seed corpus, independent of the load path (my changes don't touch eval). Not committed as a crasher.

Reviewers' repro shapes are preserved as the regression tests above, so both attackers can be re-run against the patched head.


---

PR #544, opened 2026-08-25
lisp: two seal-adjacent allocation cuts — Array dims copy, per-fork metadata rebuilds

sam-at-luther:
Two independent allocation cuts plus adversarial-review follow-ups, one commit each. Both cuts came out of the #379 disposition profiling; the second is the elps-core remainder identified in substrate#440.

## Commit 1 — `lisp: stop deep-copying dims Array built for itself` (`1ccfbf8`)
- `lisp.Array(nil, cells)` built its own one-element dims list and then deep-copied it before storing — a defensive copy of a list nothing else can hold. The copy now happens only for **caller-supplied** dims, where it is load-bearing: `builtinSelect`/`builtinReject` resize a vector's stored dims in place, so aliasing a caller's list would be observable.
- Uses a separate `stored` variable instead of reassigning the `dims` parameter: the obvious one-liner defeats escape analysis and regresses the caller path 8→9 allocs/op; this way the caller path stays at 8 and the self-built path drops 8→6.
- `TestArrayDoesNotAliasCallerDims` pins the aliasing guard (fails if the remaining copy is removed).
- CI benchmark gate (interleaved, n=10, same runner): 0 rows at gate. Highlights: `libjson Load/default` −4.44% allocs; `lisp EnvFunCallPos/Key/Opt` −8.7…−9.1% allocs; `libelpspath` delete-compaction rows −20…−26% allocs, package geomean −10.7% sec/op.

## Commit 2 — `perf(lisp): stop rebuilding dead and empty per-fork metadata` (`aa9f9f7`)
Applying "prove immutable, then share" to Fork's per-env metadata found rebuilds of things carrying nothing:
- **`LEnv.funName` deleted** — dead state with a `TODO(elps2)`; no site in the repository writes or reads it, yet every env allocated it and every fork rebuilt it.
- **`copyLocation` per forked env dropped** — `LEnv.loc` is eval's location register, rebound on every step; a quiescent template holds only the leftover position of the last node it evaluated. A fork now starts with an empty register, exactly as it starts with an empty call stack and zero `TotalSteps`. Strictly more hermetic than the copy.
- **`symbolDocs` lazily allocated** — 10 of 13 packages in a loaded env document no symbol; `setSymbolDoc` is the sole writer and allocates on first write, all readers nil-safe.
- **`funNames`/`externals` sharing measured and declined** (recorded in code): both have live template-side writers, so sharing is the issue #397 fatal-concurrent-map hazard, and `Exports` sorts `externals` in place. Worth −45/−14 allocs/op; not worth the CoW machinery.
- Red-proofs executed for every guard (reinstating the location copy, sharing the location pointer, dropping the nil-map guard, sharing `symbolDocs` — each fails its test); no gate weakened, no exemption added. New `lisp/fork_metadata_test.go` carries the guards.
- Fork-mode env construction −2.20% allocs on the in-repo 9-env template; the two deleted per-env sites dominate downstream (substrate templates have ~557 envs; see substrate#440/substrate#441). Bonus: `EnvGet` −7.81% allocs — the `funName` removal lands on the evaluator's own path, helping environments nobody forks.

## Commit 3 — `review(lisp): copy-free Array error paths, fork.md loc row, test-helper honesty` (`f4cc506`)
Follow-ups from an adversarial review of the two commits above, which confirmed **no correctness bugs** (it specifically attacked and ruled out: a shared/interned `Int` in the uncopied dims, a missed `symbolDocs` writer, a nil `loc` deref on a fork's pre-first-eval error path):
- `Array` defers the caller-dims deep copy until every validation check passes, restoring copy-free error paths; error precedence unchanged.
- `docs/fork.md`'s what-is-shared table now records the evaluator location register as fresh (the embedder-visible half of commit 2).
- The `forkedEnvs` test-helper comment now states exactly what the helper walks instead of overclaiming parity with `forker.env`.
- Noted for a follow-up, deliberately not taken here: ~9 hot list-to-vector builtins still pass self-built dims as caller dims and could use the copy-free path (~3 allocs/vector), but that interacts with `select`/`reject`'s dims resizing and deserves its own measured change.

## Test Plan
- [x] `go test ./...` clean on all three commits; `-race ./lisp/` and `-tags elpscheck` clean
- [x] `make elpsvet` clean (both passes); `golangci-lint run ./lisp/...` 0 issues; `elps fmt -l` / `elps doc -m` clean
- [x] Red-proofs for every new guard executed and restored (details above)
- [x] CI benchmark gate on the PR: 0 rows at or above threshold (timing 15%, allocation 5%)

CI note: the mid-review red checks on this PR were cancellation artifacts of checkpoint pushes and a shard re-run racing each other in the same concurrency group. The one genuine fuzz finding (shard 6, `FuzzSharedProgramMultiEnv` crasher `f34b03570d1e03ac`) was triaged in the comments: a budget-bounded runaway input that replays green and identically on head and base; the worker death was wall-clock during minimization.

https://claude.ai/code/session_01EgShxM3EoPmeVnq1cmpcPK

### DIFF
--- a/docs/fork.md
+++ b/docs/fork.md
@@ -124,6 +124,7 @@ the same reasoning that exempts the nil/true/false singletons).
 | Limit configuration (`MaxAlloc`, stack bounds, step budget, ...) | copied |
 | `Profiler`, `Debugger` | do not travel (fork starts with none) |
 | Call stack, condition stack, step accounting | fresh |
+| Evaluator location register (`Source()`) | fresh — a fork starts with no position, so an error raised before its first evaluation reports `<native code>`, not the template's last position |
 | Env-ID and gensym counters | **continued** past the template's |
 
 `LoadCache` is shared for the same reason `Reader` is, and because
--- a/docs/sealed-ast.md
+++ b/docs/sealed-ast.md
@@ -253,7 +253,7 @@ writes — so those fields are unexported (#362 for `source`, #382 for
   environment, in every function value that closed over it — without
   passing `Put`, the seal, or elpsvet; `env.Loc = loc` aliased a caller's
   mutable location into every error and frame stamped afterwards (the #362
-  class).  `scope`, `funName`, `parent` and `loc` are unexported; reads go
+  class).  `scope`, `parent` and `loc` are unexported; reads go
   through `Bindings()` (an `iter.Seq2` over the immediate scope),
   `NumBindings()`, `Parent()` and `Source()` (a location *copy*).
   `Runtime` and `ID` stay exported — neither is a container an embedder can
--- a/lisp/array_test.go
+++ b/lisp/array_test.go
@@ -80,6 +80,29 @@ func TestArray(t *testing.T) {
 	elpstest.RunTestSuite(t, tests)
 }
 
+// TestArrayDoesNotAliasCallerDims pins the constraint that decides which
+// Array calls copy their dims.  An array rewrites its own cardinality in
+// place -- builtinSelect and builtinReject size a vector's dims to the
+// element count once the predicate has run -- so a caller-supplied dims list
+// must be copied or that write lands in a value the caller still holds.  Dims
+// Array constructs for itself (the dims == nil entry) are reachable from
+// nothing else, so they are stored directly and cost no copy.
+func TestArrayDoesNotAliasCallerDims(t *testing.T) {
+	dims := lisp.QExpr([]*lisp.LVal{lisp.Int(3)})
+	arr := lisp.Array(dims, []*lisp.LVal{lisp.Int(1), lisp.Int(2), lisp.Int(3)})
+	require.NotEqual(t, lisp.LError, arr.Type, "constructing the array: %v", arr)
+	require.NotSame(t, dims, arr.Cells[0], "the array stored the caller's dims list")
+	arr.Cells[0].Cells[0].Int = 2 // the resize select/reject perform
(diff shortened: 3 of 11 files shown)


sam-at-luther:
**Fuzz shard 6/10 failure triaged: not this PR's.** The failing check is `FuzzSharedProgramMultiEnv` — the worker died during minimization ("fuzzing process hung or terminated unexpectedly while minimizing: EOF") after finding crasher `f34b03570d1e03ac`:

```
(defun countdown(n)(if(= n 0)e(countdown(-  1)))) (countdown 10000)
```

A tail-recursive runaway (`(- 1)` never reaches 0) that the eval budget correctly stops — but each evaluation burns the full budget, and this target evaluates every input many times (baseline + determinism control + shared runs × up to 3 envs × 3 reps), so on a loaded shard the minimization loop's re-executions ran into the harness watchdog and the worker process died.

Attribution evidence:
- The captured crasher input **replays green on both arms**, deterministically and identically: PASS in 4.95s on this PR's head (`1ccfbf8`) and 4.93s on base (`48a04a9`).
- The input constructs no arrays; this PR's diff touches only `lisp.Array`'s dims handling.
- The failure mode is a worker-process death mid-minimization (wall-clock), not an assertion failure; there is no fix to port.

Re-running the failed job once per the flake protocol; a second failure would be treated as real. If this recurs, the durable fix is target-shaped, not diff-shaped: the per-input evaluation count (envs × reps × control) times the step budget needs to stay comfortably under the watchdog on a loaded runner — worth an issue against the fuzz harness if it reproduces.

---

PR #567, opened 2026-08-29
libelpspath: take the selector parser upstream, and make every path algorithm linear

### LINKED ISSUE #563: elpspath: no spelling for an open-ended range, though the engine already supports it
## Summary

`rangePath` already implements an open-ended slice — `validateRange` takes an
`implicitTo` flag and, when set, resolves the end to the sequence length:

```go
// lisp/lisplib/libelpspath/path.go:1262
func validateRange(n int, from int, to int, implicitTo bool) (int, int, error) {
	if from < 0 {
		from = n + from
	}
	if implicitTo {
		to = n
	}
	...
```

`Range(from, to, implicitTo)` exposes the flag. But **nothing in elps ever
passes `true`.** The positional-step parser hardcodes it shut:

```go
// lisp/lisplib/libelpspath/query.go:145
case "range":
	if len(cells) != 3 {
		return nil, fmt.Errorf("range requires exactly 2 arguments (from to), got %d", len(cells)-1)
	}
	...
	return Range(from.Int, to.Int, false), nil
```

So `'(range 1)` is a parse error, and there is no way to write "from index 1 to
the end" through the `?` family. The only caller that constructs an
`implicitTo` path today is an embedder's own jq-string front end, which builds
`Path` values directly rather than going through `ArgsToPath`.

Every other slice shape already round-trips: `'(range 0 2)` (open start via an
explicit 0), negative `from`, negative `to`, `'*`, identity. Open **end** is the
single gap.

## Proposal

Accept a one-argument `range` step:

```lisp
(? scores '(range 1))    ; elements [1, len)
(? scores '(range -2))   ; last two elements
```

`(range from to)` keeps its current meaning. `(range)` and `(range a b c)` stay
errors, with the arity message updated to say 1 or 2.

## Why this spelling

This is not an invention — it is what the design elpspath's positional-step
surface was ported from already does. cljq (`q3cpma-json-utils`,
<https://git.sr.ht/~q3cpma/cl-json-utils>) is the model: same `?` operator name,
same string-key / integer-index / `*` wildcard positional steps, same negative
indices, and the same selector name `range`. Its docstring and implementation:

```lisp
   (:RANGE start [end])  Convenient syntax for (:AND [start..end[) on arrays

   (:range
    (destructuring-bind (start &optional (end (length json))) (cdr selector)
```

`(end (length json))` is exactly `implicitTo`. elps ported the selector and
dropped the optional arm.

## Secondary defect: `String()` misreports an implicit end

```go
// lisp/lisplib/libelpspath/path.go:1258
func (s *rangePath) String() string {
	return "[" + strconv.Itoa(s.from) + ":" + strconv.Itoa(s.to) + "]"
}
```

It ignores `implicitTo`, so a path built as `Range(1, 0, true)` renders as
`.[1:0]` — which reads as an empty slice and is not the path that would parse
back to it. It should render `.[1:]`. This is reachable today by any embedder
constructing such a path, independent of the parser change above.

## Scope

- `query.go`: accept the 1-argument form, update the arity error.
- `path.go`: `rangePath.String()` honours `implicitTo`.
- Docstrings: `packageDoc` step table and the `?` builtin doc.
- Tests: the existing `"range wrong arg count"` case uses `(range 1)`, which
  becomes valid — it needs to move to a genuinely wrong arity.

### LINKED ISSUE #564: elpspath: the jq-string selector parser belongs upstream, next to the engine it targets
## Summary

One repository owns the *syntax* of a path language whose *semantics* live in
another.

`lisp/lisplib/libelpspath` owns the path engine: the `Path` interface, its six
implementations, the deep-copy helpers, the cycle guard, and the
positional-step builtins. `luthersystems/substrate`'s
`internal/substrate/shiro/libelpspath/parse.go` owns a jq-style selector
parser — 150 lines that turn `.a.b[0]["x y"][1:3]` into
`Root(Chain(Dot("a"), Dot("b"), Index(0), Dot("x y"), Range(1, 3, false)))`.

That parser touches nothing substrate-specific. Every symbol it names —
`Root`, `Chain`, `Dot`, `Index`, `Iter`, `Range` — is already exported from
here. It is pure translation into this package's constructors, and the only
reason it lives downstream is that it arrived with the deprecated jq-string
*builtins*, which upstream deliberately declined to take.

The builtins were the right thing to leave behind. The parser was not.

## Why it matters

**The corpus is the asset.** substrate's `parse_test.go` (723 lines) and
`parse_regexp_test.go` (215 lines) are a comprehensive behavioural corpus for
the engine — ~90 cases crossing every selector shape with all seven
operations, checked against jq's documented behaviour and its playground.
Roughly a fifth of those cases test the *engine*, not the parser: iterator
collapse, range splice arity, delete-of-root, in-place vs copy semantics. That
coverage is being spent downstream on code that lives here.

**The syntax cannot be evolved from where it is.** issue #563 is the worked
example. It added a one-argument `(range from)` step and fixed
`rangePath.String()`, which had been rendering `Range(1, 0, true)` as
`[1:0]` — an empty slice, a different path. The downstream parser has emitted
`implicitTo: true` paths since the beginning, from `[1:]` and `[:]`; it was
the only caller that ever did. So a printing defect in this package was
reachable only through a parser this package cannot see, and the round-trip
test that would have caught it could not be written on either side of the
split.

**A downstream consumer reaches an unexported gate through a hack.** See
below.

## Proposal

Add the parser here as a Go-level API. Nothing lisp-visible changes.

```go
// lisp/lisplib/libelpspath/selector.go
func ParseSelector(selector string) (Path, error)
```

Not bare `Parse` — this repository has a lisp reader, and an exported `Parse`
in a lisp package reads as that.

Port `parse_test.go` and `parse_regexp_test.go` with it, adapted rather than
thinned: they are most of the value of the move.

Also export the value gate. substrate's README already asks for this by name:

> Upstream's value gate (`okSimpleType`) is unexported, so the v1 builtins
> reach it through the one exported entry point that is exactly "run the gate
> and return the value unchanged", `BuiltinQueryGet` with no path steps.

That is a real hack in production code, and its comment says so. The gate is
not optional — it is what refuses a value that contains itself, and every
unguarded recursive walk over such an LVal grows the goroutine stack until the
runtime kills the process, which `recover()` cannot intercept (#393). A
downstream embedder must be able to call it directly:

```go
func OKSimpleType(v *lisp.LVal) error
```

## Explicitly out of scope

The v1 builtins — `get-path`, `set-path!`, `set-path`, `del-path!`,
`del-path`, `nil-path`, `nil-path!` — do **not** move. They are deprecated in
favour of the positional-step family, they are kept alive only because phylum
code builds path strings at runtime and on-chain envelopes persist them, and
declining them was a deliberate decision. Taking them now would re-introduce
deprecated lisp surface upstream. This issue is the Go-level parser only.

## Follow-up downstream

substrate's `Parse` becomes a thin alias once a release carries this, and
`checkSimpleType`'s indirection is replaced with a direct `OKSimpleType` call.
Separate PR; it has to wait for the release.

One doc correction falls out of #563 and should ride along: substrate's README
says "One v1 form has no v2 spelling — the open-ended slices `[1:]` and `[:]`,
because a v2 range step needs both bounds." That is no longer true. `[1:]` is
`'(range 1)` and `[:]` is `'(range 0)`.

### LINKED ISSUE #565: elpspath: Chain construction is exponential in the number of iterator steps — 24 `'*` steps cost 4.7s
## Summary

`Chain` is exponential in the number of `Iter` steps it is given. It is not
the path *evaluation* that is slow — it is building the `Path` value, before
any document is touched.

Measured on `feature/elpspath-parse-selector` (the numbers are the same on
`main`), `ArgsToPath` with N `'*` steps and nothing else:

| N iterator steps | time to construct the Path |
|---|---|
| 4 | 32µs |
| 8 | 126µs |
| 12 | 1.6ms |
| 16 | 21.7ms |
| 20 | 313ms |
| 22 | 1.22s |
| 24 | 4.74s |

Each additional step roughly doubles it. Interleaving keys makes it worse
(24 `'*`/key pairs: 8.98s).

## Reachable from shipped surface

This is the `?` family, not an internal API:

```lisp
(? obj '* '* '* '* '* '* '* '* '* '* '* '*
      '* '* '* '* '* '* '* '* '* '* '* '*)   ; 24 steps, ~4.7s, no document needed
```

Downstream the same path is `.[][][]...`, a **49-byte string**, through
substrate's `get-path`. Nothing bounds the step count: `?` is `&rest`, and a
phylum that builds a path string at runtime can reach any length. A handful
more steps takes the process out for minutes and then hours.

## Cause

Two compounding problems in `path.go`:

1. `normalizePaths` prepends one element at a time —
   `curChain = append([]Path{path}, curChain...)` — which is O(n²) on its own.
2. On an iterator it does `curChain = []Path{Iter(curChain...)}`, and `Iter`
   is `&iterPath{path: Chain(paths...)}`, so it re-enters `Chain` →
   `normalizePaths` over a chain that is *already normalized*. The tail is
   re-normalized once per iterator, and the tail itself contains iterators.
   That is the doubling. The code notes the mutual recursion —

   ```go
   // NOTE: there can be mutual recursion here between Iter/Chain.
   curChain = []Path{Iter(curChain...)}
   ```

   — without noting its cost.

## Why the fuzzers have not found it

`FuzzPathEngine` generates its steps under `pathGenBudget`, so it does not
reach step counts in this range. There is no fuzz target over the selector
parser, and #564 deliberately does not add one for exactly this reason: a
watchdog target over `ParseSelector` goes red on its first campaign, and it
would be reporting *this* defect as a parser hang.

## Scope

- `normalizePaths`: build the slice in one pass instead of prepending.
- `Iter`/`Chain`: do not re-normalize a chain that is already normalized.
- Then add `FuzzParseSelector` with a `fuzzwatch` budget — the natural
  regression guard, and the reason to do these together. See the "NO FUZZ
  TARGET YET" note in `selector.go`.

A benchmark over step count would make the fix legible; there is none today.

Found while doing #564.

### LINKED ISSUE #566: elpspath: a jq selector may contain at most one bracketed key — `.["a"]["b"]` is a parse error
## Summary

`reArrayKey` matches a quoted map key greedily across the entire remaining
selector, so any selector with **two** bracketed keys fails to parse.

```go
// lisp/lisplib/libelpspath/selector.go
reArrayKey = regexp.MustCompile(`^\[\s*("(?:\"|[^"])*")\s*\]\s*(\?)?`)
```

In a Go raw string `\"` is a backslash followed by a quote, which the regexp
engine reads as an *escaped quote* — i.e. a literal `"`. So the alternation
`(?:"|[^"])` matches **every character**, the group is `"` + `.*` + `"`, and
it is greedy.

Given `["a"]["b"]` it captures `"a"]["b"` and hands that to `strconv.Unquote`,
which rejects the interior quote:

```
.["a"]                          ok
.["a"].b                        ok      (the second key is a bare one)
.a["b"]                         ok
.["a\"b"]                       ok      (one key, escapes intact)
.["a"]["b"]                     invalid syntax
.["first name"]["last name"]    invalid syntax
.["a"][0]["b"]                  invalid syntax
.["a"][]["b"]                   invalid syntax
```

The bracketed form is the only spelling for a key that is not a bare
identifier (`[A-Za-z_][A-Za-z_0-9]*`), so this bites precisely the paths that
have no alternative: two nested keys with spaces, dollar signs, digits or
hyphens. `.["first name"]["last name"]` is an ordinary thing to want.

## It rejects rather than mis-parses

Every case observed fails loudly. The reason it should generalise: a span
reaching across two keys necessarily contains the first key's closing quote,
which is unescaped in its own literal, and `Unquote` refuses every such input.
That is an argument rather than an exhaustive proof — worth stating plainly
rather than relying on.

## A second symptom, which is how it was found

`dotPath.String()` renders a key with `%q`, so **every** key prints as a
bracketed one. A two-key path therefore does not survive its own printer:

```
.a.b     ->  String()  .["a"]["b"]  ->  reparse: invalid syntax
.a[].b   ->  String()  .["a"][]["b"] ->  reparse: invalid syntax
```

`TestParseSelectorRoundTrip` (#564) excludes exactly these, with a comment
saying why.

## Fix

Give the body proper Go string-literal syntax:

```go
reArrayKey = regexp.MustCompile(`^\[\s*("(?:\\.|[^"\\])*")\s*\]\s*(\?)?`)
```

Verified: with that one change all four rejected selectors above parse to the
obvious path, `.["a\"b"]` still parses correctly, and the rest of the suite is
green. Because the current behaviour on these inputs is *an error*, the change
cannot alter the meaning of a selector that parses today — it only turns
errors into parses.

## Scope

- `selector.go`: the regexp.
- `selector_test.go`: invert `TestParseSelectorTwoQuotedKeysIsBroken` (it is
  written to go red on the fix, and that is its red proof), and move `.a.b`
  and `.a[].b` into `TestParseSelectorRoundTrip`'s selector list.
- `selector.go`: drop the third wart from `ParseSelector`'s doc comment.
- Downstream: substrate's `parse.go` carries the same regexp today, so a
  phylum using `get-path` with two special-character keys is affected until it
  picks up the release.

Preserved deliberately by #564 so that the move stayed a move.

sam-at-luther:
Closes #563. Closes #564. Closes #565. Closes #566.

elps now owns both ends of the selector language — the parser and the printer.
Once it did, three defects became visible that neither repository could see
alone, and a complexity sweep of the package found three more.

Read commit by commit; each is self-contained and the series is bisectable.

## Part 1 — the language

**`c847c766` open-ended range step (#563).** `rangePath` could always express
"from here to the end" — it carries an `implicitTo` flag and `validateRange`
resolves the end against the input length — but nothing in elps ever set it, so
there was no way to write it:

```lisp
(? scores '(range 1))    ; elements [1, len)
(? scores '(range -2))   ; the last two elements
```

Not an invented spelling. The positional-step surface was modelled on
[cljq](https://git.sr.ht/~q3cpma/cl-json-utils) — same `?` operator name, same
string-key/int-index/wildcard steps, same negative indices, same selector name
`range`. Its range takes an optional end defaulting to the sequence length,
which *is* `implicitTo`. Also fixes `String()` ignoring the flag, which rendered
`Range(1,0,true)` as `.[1:0]` — an empty slice.

**`dc67e617` the parser moves upstream (#564).** `ParseSelector(string) (Path,
error)`, ported from substrate with its full corpus (79 cases + 12 regexp cases,
none dropped). One repo owned the syntax of a language whose semantics lived in
another.

Verified behaviour-preserving against substrate's parser at that commit: 23,684
comparisons over `String()`, result JSON and exact error text, zero mismatches.
Later commits diverge deliberately — see **Downstream**, which measures how far.

Scope held deliberately tight: **no new lisp builtins** (the v1 jq-string
builtins are deprecated and stay downstream), and **no new exported Go API**.
`okSimpleType` was briefly exported for a future downstream call site and has
been put back — every call site is inside the package, `ParseSelector` does not
use it, and exported API should be added by the change that needs it.

**`991f2e5c`** removes a third-party header. ⚠️ **Its commit message contains a
factual error, corrected [in a comment
below](https://github.com/luthersystems/elps/pull/567#issuecomment-5461154948)**
— the parser is a near-verbatim port (107 of 113 lines identical), not a
rewrite. The removal was a deliberate owner decision; the justification given
for it was wrong, and whether to restore the notice or add a `NOTICE` file is a
call for review.

## Part 2 — the printer emitted things the parser could not read

**`daa7d942` greedy quoted key (#566).** `reArrayKey`'s body was
`(?:\"|[^"])*`. In a regexp `\"` is a plain escaped quote, so the alternation
read "a quote OR not a quote" — every character — and ran greedily to the last
quote in the selector.

The input restriction was the smaller half. `String()` brackets every map key,
so `.a.b` printed as `.["a"]["b"]` — **the parser emitting output it could not
read back**. The round-trip test had to exclude every two-key selector, which is
most of them.

Now `(?:\\.|[^"\\])*` — the ordinary string-literal grammar. Verified purely
additive by exhaustive generation over an adversarial alphabet, and
independently in review over **every string of length ≤ 7 over `{[ ] " \ a b
space newline}` — 2,396,745 inputs**: zero regressions, 28 newly-accepted.

**`fca8e353` the identity selector.** `ParseSelector(".")` returned bare
`Chain()`, whose `String()` is the empty string — the same defect in a second
spot. Now `Root(Chain())`, matching `ArgsToPath`. **Found by the fuzzer on its
first seed.**

**`c3c434ba`, `aeda08d8` the fuzz target.** `FuzzPathEngine` generates steps as
lisp values and never sees a selector string, so the grammar was upstream of
everything it exercised. The invariant is not "does not panic":

> a Path this parser produces must print to a selector this parser reads back
> as the same path

Compared *behaviourally* across three non-mutating operations and six documents,
because equal printed forms do not imply equal paths — `.[1:0]` prints as itself
and means the wrong thing. Verified to catch what it was written for: with #566
restored it fails on seed #2, with the root spelling restored on seed #0.

## Part 3 — every algorithm is linear

Three superlinear algorithms, all live in released code, none introduced here.

**`a2a6a245` exponential construction (#565).** `normalizePaths`' iterator
branch called `Iter` → `Chain` → `normalizePaths`, so every iterator
re-normalized the whole tail — one re-entry per iterator, each over what the
previous rebuilt, is 2ⁿ.

| iterator steps | 8 | 12 | 16 | 20 | 24 |
|---|---|---|---|---|---|
| before | 93µs | 1.2ms | 21ms | 292ms | **4.7s** |
| after | 9µs | 8µs | 11µs | 14µs | **5µs** |

24 steps allocated **134,217,706** times; now **201**. Reachable from a 45-byte
selector and from the shipped `?` builtin.

**`d440366d` quadratic assembly (#565).** The chain was built by prepending,
copying everything already placed on every step. One slice per step either way,
so the allocation *count* stayed linear and a count-based test could not see it;
the *bytes* were O(n²).

| dot steps | 100 | 200 | 400 | 800 | 1600 | 3200 |
|---|---|---|---|---|---|---|
| before | 91µs | 677µs | 685µs | 2.3ms | 7.9ms | **38.2ms** |
| after | 12µs | 18µs | 29µs | 60µs | 132µs | **260µs** |

2000 steps allocated **34,068,984** bytes; now **231,064**.

**`95f8b2d5` quadratic rendering.** Composites rendered by materialising each
child's full string — one full-length copy per nesting level.

| nested iterators | 100 | 200 | 400 | 800 |
|---|---|---|---|---|
| before | 14.6µs | 32.8µs | 166µs | **706µs** |
| after | 2.6µs | 3.3µs | 6.2µs | **13.6µs** |

The three composites share one builder via an unexported `stringAppender`.
`Path` does **not** require it, so an embedder's own implementation still works
through a `String()` fallback.

### Nothing else is superlinear

Measured at head, per doubling: iterator construction 1.65×/2.17×/1.99×,
dot-step construction 1.65×/1.85×/2.52×, rendering 2.84×/1.59×/2.16×, parsing
1.75×/2.11×/2.33×. Quadratic would show ~4×. Selector parsing is flat in
ns/byte across nine shapes; deep `Get`/`Set`/`Delete`/`Nil`, wide iteration and
the range splice are all linear — the splice allocates exactly 2n+c.

### Practical inputs do not regress

Against `main`, n=10, `benchstat` — every one statistically indistinguishable,
with deterministic allocation counts down (`ArgsToPath` −21%, `String` −11%):

```
PracticalGet2         2.835µ ± 12%   2.849µ ±  8%   ~ (p=0.436)
PracticalGet3         2.896µ ±  7%   3.034µ ±  8%   ~ (p=0.363)
PracticalGetIndex     2.816µ ± 11%   3.030µ ± 10%   ~ (p=0.190)
PracticalGetIter      6.009µ ±  4%   6.006µ ±  5%   ~ (p=0.971)
PracticalGetRange     5.660µ ±  6%   5.511µ ±  4%   ~ (p=0.529)
PracticalSet          6.199µ ± 12%   5.793µ ±  8%   ~ (p=0.105)
PracticalDel          5.104µ ±  9%   5.144µ ± 12%   ~ (p=0.494)
PracticalArgsToPath   333.8n ± 16%   266.6n ± 48%   ~ (p=0.052)
PracticalString       349.0n ± 19%   350.5n ±  4%   ~ (p=0.739)
geomean                                             -2.00%
```

**`0625aecc` commits benchmarks so the gate catches the next one.** The cost
unit tests assert allocations and bytes, so they fail hard on a *complexity*
regression and name it; nothing watched the *constant factor*. Each stage now
has a practical arm and a larger one, calibrated so a complexity regression
diverges rather than tracks: iterators 8→16 costs **1.88×** today and would cost
~256× exponential; keys 100→400 costs **4.06×** against 16× quadratic; nested
iterators 50→200 costs **4.09×** against 16×.

## `89a2682b` — acting on an adversarial review

The branch was independently attacked before finalising. Six issues, **five of
them defects in tests and comments this branch added** — each an assertion that
looked right and was not checked:

1. **`TestStringIsLinearInNesting` could not fail.** Its bound was 1MiB at depth
   600, where the quadratic renderer allocates ~782KB — *above* the cost it
   pinned. It was red-proofed when written and the red-proof was misread: a
   large byte count taken for a failure without checking a `FAIL` line appeared.
2. **`newSpliceOracle` still bailed on the open-ended range** — invariant 6,
   which *is* the splice-bounds property, while the comment on the *other*
   witness argued that bailing leaves a step unwatched.
3. **The generator change silently repurposed committed seeds.** Adding the
   open-range draw shifted byte consumption, so four seeds moved arms — taking
   the explicit-range witness for `?set!`, `?del!` and `?nil!` with them, which
   is where #471's splice bug lived. The coverage test classified both arms as
   "range" and stayed green. It now asserts coverage per operation **and** form.
4. **The second half of `TestNormalizePathsIsNotExponential` asserted nothing.**
5. **Five stale comments**, including a "NO FUZZ TARGET YET" block whose target
   and blocking fix are both on this branch.

What it attacked and could **not** break: the `normalizePaths` rewrite
(exhaustive over 11,111 sequences plus 200,000 randomised — identical output, no
shared slice backings across ~7,000 trees), the regexp change, the
`stringAppender` split, round-trip over all 65,536 two-byte keys, and substrate
drift over **979,194 selectors** — `BEHAVIOUR-DIFFERS: 0`.

## Test plan

- `go test ./...` and `make elpsvet` clean; every commit builds and passes alone.
- Every new test red-proofed against unmodified source. The two #565 *meaning*
  tests deliberately pass against the unfixed code — the old behaviour was
  correct and only slow; the cost tests are what go red.
- Cost regressions assert **allocations** or **bytes**, never wall time.
- Verified against CI's exact golangci-lint (v2.6.2, not the older local one):
  zero findings.
- `FuzzPathEngine` **5.1M** executions with the widened splice oracle,
  `FuzzParseSelector` **8.0M**, no crashers. Seed coverage 21 seeds × 7
  operations × 3 step forms.
- Full substrate suite against this branch: green except the two documented Git
  LFS pointer-stub failures.

## Downstream: what this changes for substrate

Making substrate's `Parse` a thin alias is **not** a pure refactor. A transcript
of 42 selectors × 7 documents × 4 operations, generated through **substrate's
`Parse` on released elps v1.59.0** and through **this branch**:

**44 of ~1,200 rows differ, and every one is intentional:**

| rows | what changed |
|---|---|
| 15 | `STRING` only — `.` prints `"."` not `""`; `.[1:]` prints `.[1:]` not `.[1:0]` |
| 28 | every operation on `.["a"]["b"]` — which **production cannot parse at all** |
| 1 | that selector's `PARSE-ERR` line disappearing |

**Zero value changes for any selector that already parsed.** The set of
selectors with a differing `get`/`set`/`del`/`nil` row is exactly the set
production rejected. Re-confirmed in review over 979,194 selectors:
`SUBSTRATE-ACCEPTS-ELPS-REJECTS: 0`, `BEHAVIOUR-DIFFERS: 0`.

Note for that PR: substrate reaches elps's value gate through a documented hack
(`BuiltinQueryGet` with no path steps). Exporting the gate would remove it, and
is worth doing *there*, where the call site is real.

Substrate's `README.md` and `packageDoc` also still claim the open-ended slices
have no v2 spelling. `[1:]` is `'(range 1)`, `[:]` is `'(range 0)`.

### DIFF
--- a/lisp/lisplib/libelpspath/libelpspath.go
+++ b/lisp/lisplib/libelpspath/libelpspath.go
@@ -149,10 +149,10 @@ var errCyclicValue = errors.New("cannot operate on a value that contains itself"
 //
 // It continues the walk g is already on rather than starting a fresh one.
 // Every nested check must pass g down; starting a new walk per level resets
-// the bound on every lap and it never fires.  okSimpleType is the entry
+// the bound on every lap and it never fires.  OKSimpleType is the entry
 // point that begins a walk.
 //
-// With okSimpleType this is the gate every builtin runs before touching a
+// With OKSimpleType this is the gate every builtin runs before touching a
 // value, and the rest of the package relies on what it rejects: copyLVal's
 // multi-dimensional array branch cannot construct a copy and says so, and
 // stays unreachable only because this refuses such an array first.  A cycle
@@ -226,18 +226,33 @@ func okSimpleContainerContents(in *lisp.LVal, g cycleGuard) error {
 	}
 }
 
-// okSimpleType ensures that the lval is a valid simple type compatible with
+// OKSimpleType ensures that the lval is a valid simple type compatible with
 // elpspath.
 // It sucks that we have to traverse the entire object checking the type,
 // but better to be safe.
-func okSimpleType(in *lisp.LVal) error {
+//
+// Every builtin in this package runs it before touching a value, and an
+// embedder driving the Path interface from Go must do the same. It is not a
+// nicety: it refuses a value that CONTAINS ITSELF, and every unguarded
+// recursive walk over such an LVal -- Get, a copy, a String() -- grows the
+// goroutine stack until the Go runtime kills the process, an abort recover()
+// cannot intercept and handler-bind never sees (issue #393). It also refuses
+// the multi-dimensional array copyLVal has no answer for, which is what
+// keeps that branch's "cannot construct a copy" unreachable.
+//
+// It is exported for exactly that embedder (issue #564). Before it was,
+// downstream reached the gate through the one exported entry point whose
+// only observable effect is the verdict -- BuiltinQueryGet with no path
+// steps, since ArgsToPath of an empty step list is the identity -- which is
+// a hack that had to be explained at every call site.
+func OKSimpleType(in *lisp.LVal) error {
 	var st cycleState
 	return okSimpleTypeGuarded(in, newCycleGuard(&st))
 }
 
-// okSimpleTypeGuarded is okSimpleType continuing a walk already in progress.
+// okSimpleTypeGuarded is OKSimpleType continuing a walk already in progress.
 // The guard is threaded through rather than re-created because the recursion
-// that overflows the stack is okSimpleType <-> okSimpleContainerType, so a
+// that overflows the stack is OKSimpleType <-> okSimpleContainerType, so a
 // bound either survives the round trip or does nothing.
 func okSimpleTypeGuarded(in *lisp.LVal, g cycleGuard) error {
 	if in.IsNil() {
--- a/lisp/lisplib/libelpspath/libelpspath.go
+++ b/lisp/lisplib/libelpspath/libelpspath.go
@@ -21,7 +21,7 @@ step, no string DSL. Path step types:
   string    map key         "foo"
   int       array index     0, -1
   symbol    iterate all     '*
-  list      array slice     '(range 1 3)
+  list      array slice     '(range 1 3), '(range 1)
 
 Functions ending in "!" mutate in place; those without "!" return
 a copy and leave the original unchanged. For ?set! and ?set, the
@@ -74,10 +74,13 @@ var builtins = []*libutil.Builtin{
 
 		Each arg is a path step: string for map key, int for array index,
 		'* to iterate all elements, '(range from to) for array slice.
+		The slice end is optional: '(range from) runs to the end.
 
 		(? obj "foo" 0 "bar")     => value at foo[0].bar
 		(? users '* "name")       => list of all user names
 		(? scores '(range 1 3))   => elements [1,3)
+		(? scores '(range 1))     => elements [1,end)
+		(? scores '(range -2))    => the last two elements
 		(? obj)                   => obj itself (no path steps)`),
 	libutil.FunctionDoc("?set!", lisp.Formals("val", lisp.VarArgSymbol, "steps-and-value"), BuiltinQuerySetMutate,
 		`Set value at a path specified by positional args, mutating the original.
--- a/lisp/lisplib/libelpspath/libelpspath.go
+++ b/lisp/lisplib/libelpspath/libelpspath.go
@@ -149,10 +149,10 @@ var errCyclicValue = errors.New("cannot operate on a value that contains itself"
 //
 // It continues the walk g is already on rather than starting a fresh one.
 // Every nested check must pass g down; starting a new walk per level resets
-// the bound on every lap and it never fires.  OKSimpleType is the entry
+// the bound on every lap and it never fires.  okSimpleType is the entry
 // point that begins a walk.
 //
-// With OKSimpleType this is the gate every builtin runs before touching a
+// With okSimpleType this is the gate every builtin runs before touching a
 // value, and the rest of the package relies on what it rejects: copyLVal's
 // multi-dimensional array branch cannot construct a copy and says so, and
 // stays unreachable only because this refuses such an array first.  A cycle
@@ -226,33 +226,26 @@ func okSimpleContainerContents(in *lisp.LVal, g cycleGuard) error {
 	}
 }
 
-// OKSimpleType ensures that the lval is a valid simple type compatible with
+// okSimpleType ensures that the lval is a valid simple type compatible with
 // elpspath.
 // It sucks that we have to traverse the entire object checking the type,
 // but better to be safe.
 //
-// Every builtin in this package runs it before touching a value, and an
-// embedder driving the Path interface from Go must do the same. It is not a
+// Every builtin in this package runs it before touching a value. It is not a
 // nicety: it refuses a value that CONTAINS ITSELF, and every unguarded
 // recursive walk over such an LVal -- Get, a copy, a String() -- grows the
 // goroutine stack until the Go runtime kills the process, an abort recover()
 // cannot intercept and handler-bind never sees (issue #393). It also refuses
 // the multi-dimensional array copyLVal has no answer for, which is what
 // keeps that branch's "cannot construct a copy" unreachable.
-//
-// It is exported for exactly that embedder (issue #564). Before it was,
-// downstream reached the gate through the one exported entry point whose
-// only observable effect is the verdict -- BuiltinQueryGet with no path
-// steps, since ArgsToPath of an empty step list is the identity -- which is
-// a hack that had to be explained at every call site.
-func OKSimpleType(in *lisp.LVal) error {
+func okSimpleType(in *lisp.LVal) error {
 	var st cycleState
 	return okSimpleTypeGuarded(in, newCycleGuard(&st))
 }
 
-// okSimpleTypeGuarded is OKSimpleType continuing a walk already in progress.
+// okSimpleTypeGuarded is okSimpleType continuing a walk already in progress.
 // The guard is threaded through rather than re-created because the recursion
-// that overflows the stack is OKSimpleType <-> okSimpleContainerType, so a
+// that overflows the stack is okSimpleType <-> okSimpleContainerType, so a
 // bound either survives the round trip or does nothing.
 func okSimpleTypeGuarded(in *lisp.LVal, g cycleGuard) error {
 	if in.IsNil() {
--- a/lisp/lisplib/libelpspath/libelpspath_test.go
+++ b/lisp/lisplib/libelpspath/libelpspath_test.go
@@ -8,11 +8,7 @@ package libelpspath_test
 import (
 	"testing"
 
-	"github.com/stretchr/testify/require"
-
 	"github.com/luthersystems/elps/elpstest"
-	"github.com/luthersystems/elps/lisp"
-	"github.com/luthersystems/elps/lisp/lisplib/libelpspath"
 )
 
 // TestPackage runs elpspath lisp tests.
@@ -29,45 +25,4 @@ func TestPackageCyclicValue(t *testing.T) {
 	r := &elpstest.Runner{}
 	defer r.Close()
 	r.RunTestFile(t, "libelpspath_cycle_test.lisp")
-}
-
-// TestOKSimpleTypeIsReachableFromAnEmbedder is a compile-time assertion as
-// much as a behavioural one: this file is package libelpspath_test, so it
-// can only reach the gate through the package's exported surface, which is
-// the entire point of exporting it (issue #564).
-//
-// Before the export, a Go embedder that wanted the gate had to call
-// BuiltinQueryGet with no path steps -- ArgsToPath of an empty step list is
-// the identity, so the call's only observable effect is the verdict. That
-// works, and substrate shipped it with a paragraph of comment explaining
-// why, but it needs an *LEnv the embedder may not have and it reads as a
-// path query rather than as a type check.
-//
-// The behavioural half covers the two verdicts that matter to a caller: an
-// unsupported type, and -- the reason the gate is not optional -- a value
-// that contains itself, which every unguarded recursive walk answers by
-// growing the goroutine stack until the runtime kills the process, an abort
-// recover() cannot intercept (issue #393).
-func TestOKSimpleTypeIsReachableFromAnEmbedder(t *testing.T) {
-	t.Parallel()
-
-	ok := lisp.SortedMap()
-	ok.MapSet("a", lisp.Vector([]*lisp.LVal{lisp.Int(1), lisp.String("two")}))
-	require.NoError(t, libelpspath.OKSimpleType(ok), "an ordinary nested document must be accepted")
-
-	require.NoError(t, libelpspath.OKSimpleType(lisp.Bool(true)),
-		"true and false are the two symbols the gate admits")
-	require.NoError(t, libelpspath.OKSimpleType(lisp.Symbol(lisp.FalseSymbol)),
-		"and false, spelled as the bare symbol")
-	require.Error(t, libelpspath.OKSimpleType(lisp.Symbol("not-a-boolean")),
-		"an arbitrary symbol is not a value the engine can hold")
-	require.Error(t, libelpspath.OKSimpleType(lisp.Native(struct{ X int }{1})),
-		"a native Go value is not a value the engine can hold")
-
-	// The cycle. Built here rather than borrowed from the in-package
-	// helpers so that this test stands on the exported API alone.
-	cyclic := lisp.Vector([]*lisp.LVal{lisp.Int(1)})
-	cyclic.Cells[1].Cells = append(cyclic.Cells[1].Cells, cyclic)
-	require.Error(t, libelpspath.OKSimpleType(cyclic),
-		"a value that contains itself must be REFUSED, not walked (issue #393)")
 }
--- a/lisp/lisplib/libelpspath/libelpspath_test.go
+++ b/lisp/lisplib/libelpspath/libelpspath_test.go
@@ -8,7 +8,11 @@ package libelpspath_test
 import (
 	"testing"
 
+	"github.com/stretchr/testify/require"
+
 	"github.com/luthersystems/elps/elpstest"
+	"github.com/luthersystems/elps/lisp"
+	"github.com/luthersystems/elps/lisp/lisplib/libelpspath"
 )
 
 // TestPackage runs elpspath lisp tests.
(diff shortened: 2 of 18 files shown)


sam-at-luther:
## Correction: the justification in `991f2e5c` is wrong

An adversarial review of this branch caught a factual error in a commit message,
and since the commit is pushed I am correcting the record here rather than
rewriting history.

`991f2e5c` removes the `Copyright (c) 2016 Matt Ho` / Apache-2.0 header from
`selector.go`, `selector_test.go`, `selector_regexp_test.go` and
`path_chain_test.go`. Its message justifies this by saying the parser

> has been rewritten around this repository's own Path constructors and **shares
> no implementation with the original**.

**That is not true.** Comparing `selector.go` against substrate's `parse.go`
with comments stripped and the `elpspath.` qualifier normalised:

```
substrate lines: 113   elps lines: 113
differing lines: 12    identical lines: 107
```

The entire difference is six changes: the `errors` import added, the `elpspath`
import removed, `reArrayKey`'s body (#566), `Parse` → `ParseSelector`, two
`fmt.Errorf` → `errors.New`, and `Chain(...)` → `Root(Chain(...))`. It is a
near-verbatim port, which is what `selector.go`'s own doc comment says — *"this
is a port of a parser with live downstream callers, not a rewrite"*. The commit
message contradicts the file it modified.

**What is actually true:** the header removal was a deliberate decision by the
repository owner, who holds the modifications copyright. That decision stands on
its own; it does not need — and is not supported by — the claim that the code is
unrelated to its origin.

**What a reviewer should decide.** `elps/LICENSE` is BSD-3-Clause, there is no
`NOTICE` file, and `grep -rl "Matt Ho" .` now returns nothing, so that header was
the only record of the third-party origin. Apache-2.0 §4(b)/(c) asks that such
notices be retained in derivative works. The options are to restore the header on
the ported files, to add a `NOTICE` or third-party provenance section, or to
leave it as is — but it should be an informed choice rather than a side effect of
an inaccurate commit message.

I have not changed any code for this. Flagging it rather than deciding it.

---

PR #568, opened 2026-08-29
libelpspath: parse-path, converting a string path into positional steps

sam-at-luther:
**#567 is merged**, so this now targets `main` directly. The diff is the four `parse-path` commits, the two performance commits under *Constant factors*, and the crash fix + open-range coverage at the end.

## What it does

A path that **arrives as a string** — from inside a document, a client request, or a persisted envelope — has to go through the v1 string operations today, which re-parse it on every call. `parse-path` converts it once into the steps the `?` family takes:

```lisp
(parse-path ".items[0].id")   => '("items" 0 "id")
(parse-path ".items[].id")    => '("items" * "id")
(parse-path ".items[1:3]")    => '("items" '(range 1 3))
(parse-path ".items[1:]")     => '("items" '(range 1))
(parse-path ".")              => '()

(apply ? (cons obj (parse-path sel)))
(apply ?set (concat 'list (list obj) (parse-path sel) (list v)))
```

Nothing else was missing — `apply` already splices a runtime-built step list into every `?` builtin, including `'*`, `'(range …)` and the mutating forms. It returns a **list**, not a vector, because `apply` takes a list and a vector argument raises.

**The open-ended range is why this can be lossless at all.** Until `[1:]` had a step spelling (#563, in the base PR), there was no positional form of it — a conversion would have silently dropped or mis-rendered exactly the selectors that persist on chain.

## Measured

`.items[0].id` against a small document, 12 samples, ±3–6%:

| route | µs/op | vs v1 |
|---|---|---|
| v1 today — parse the string every call | 5.554 | — |
| `parse-path` used **naively** (convert every call) | 5.974 | **7.6% slower** ⚠️ |
| `parse-path` **converted once** | 3.959 | **28.7% faster** |
| native v2 with literal steps (the ceiling) | 3.978 | 28.7% faster |

Converting once is statistically indistinguishable from native v2 — a string path performs as if it had been written positionally. Naive per-call use is genuinely slower than staying on v1, and the docstring says so.

Memoizing the conversion is the natural next step and deliberately **not** here: a cache belongs where the call sites are.

## One scan behind both surfaces

The selector scan is factored into `selectorPaths`, returning the flat leaf steps; `ParseSelector` wraps them in `Root(Chain(...))` and `SelectorSteps` renders them as lisp values. The two agree on the grammar *by construction* rather than by test — a second parser would drift, and the drift would only show on selectors nobody writes.

What is **not** free is that the steps *mean* the same thing: `SelectorSteps` renders each leaf back into a lisp value and `ArgsToPath` parses it again, so a rendering that did not invert `argToStep` would build a different path from the same string. `TestSelectorStepsMatchParseSelector` applies both routes to five documents across the grammar and requires the same value or the same error text.

**The refactor is behaviour-preserving**, verified two ways: reconstructing the pre-refactor `ParseSelector` and differential-testing over **173,501 selectors** (accept/reject, error text, `String()`, and `reflect.DeepEqual` including unexported fields — zero differences), and re-running the 91,867-selector substrate differential before vs after (**0 selectors changed behaviour**).

## Constant factors — the benchmark gate caught a real regression, then went past it

That refactor was behaviour-preserving and **not** cost-preserving, and CI said so before a human had to. Splitting the scan out of `ParseSelector` turned the step slice into a **return value**, so escape analysis could no longer keep its growth off the heap. Profiled at `-memprofilerate=1`, the delta was exactly that and nothing else — `selectorPaths` 3 allocations (cap 1→2→4, 112 B) where `ParseSelector` had 1 (64 B), every other site byte-identical. That is +2 allocs and +48 B on **every** practical selector, which the gate reported as +11.76% allocs/op and +6.88% B/op.

Two commits fix it, and the second is the same defect one frame up:

- **Presize the step slice.** `stepCapHint` takes the smaller of two upper bounds on the step count — the number of `.` and `[` bytes, exact for the shapes callers write, and `(len+1)/2`, which caps a quoted key full of them. This removes the growth rather than restoring the old escape, so the cost stops depending on where the slice is built.
- **`expandPaths` skips its copy when the path is already flat.** It built a fresh slice unconditionally, growing from nil, so a path with no nested chain or iterator paid log(n) allocations to copy a slice that was already normal — and that is the shape *every* parse and *every* `ArgsToPath` call produces. Returning the input is safe because nothing writes through the result and nothing stores it.

Worth recording: presizing `expandPaths`' accumulator instead was the first attempt and was **worse** — an iterator's inner chain recurses through it with one element that expands to none, and a presized accumulator allocates for that empty result where a nil one does not (44 extra allocations on a fifty-iterator selector). The comment says so.

Result on CI (n=10, both arms interleaved on one runner) — **no arm regresses on allocs/op or B/op anywhere in the repo**, and `libelpspath` geomeans move −3.67% sec/op, −4.19% B/op, −2.43% allocs/op:

| benchmark | sec/op | B/op | allocs/op |
|---|---|---|---|
| `ParseSelector/practical/dot` | −7.63% | −18.42% | 17 → 14 |
| `ParseSelector/practical/index` | −7.36% | −17.48% | 17 → 14 |
| `ParseSelector/practical/quoted` | −4.66% | −17.61% | 18 → 15 |
| `ParseSelector/practical/range` | −1.89% | −3.70% | 11 → 10 |
| `ParseSelector/keys/100` | −6.10% | −28.05% | 318 → 305 |
| `PathConstruct/practical/key.key.key` | −31.11% | −32.56% | 11 → 8 |
| `PathConstruct/keys/400` | −24.00% | −41.94% | 415 → 405 |
| `PathEndToEnd/get/.items[0].id` | ~ | −2.84% | 69 → 66 |

The one `libelpspath` move in the bad direction is `PathConstruct/practical/key.iter.key` at +4.60% sec/op (gate 15%) with identical allocations — the extra type-switch pass, on the one shape that allocates regardless. Gate report: `0 at or above the gate`.

Three new tests, each red-proofed against the mutation it names. `TestStepCapHintBoundsStepCount` pins the upper bound and that the scan never grows past it, and `FuzzParseSelector` asserts the same on everything it accepts. `TestStepCapHintIsTightOnPracticalSelectors` pins the other half — an upper bound alone is satisfied by `len(selector)`, which would trade the growth allocations for wasted bytes. `TestExpandPathsFastPathAgreesWithTheLoop` reimplements the loop unconditionally, because the fast path's one failure mode is drift: a composite type added to the loop but not its type switch would silently stop normalizing, leaving a valid path that is merely un-normalized.

## A crash found while testing the open range

Asked to cover `?set` on an open-ended range, the new tests turned up a panic that **predates this branch and reproduces on `main`**.

`toCells` asked whether an array had **more** than one dimension. A **zero**-dimensional array — dims `'()`, which `lisp.Array` builds for an embedder passing an empty dims list — answered no and was accepted as indexable. It has no cardinality slot, and `storeCells` writes one unconditionally, so every in-place write through it panicked on `dims.Cells[0].Int = len(vals)`:

```
runtime error: index out of range [0] with length 0
```

Ten combinations: `?set!`, `?del!` and `?nil!` over an index or either range spelling. Not reachable from lisp — every in-tree constructor derives dims as `[len]` — but `lisp.Array`'s dims parameter is documented as chiefly serving embedders, and an embedder calling `Path.SetMutate` is behind no recover at all, so the panic reaches the host. It is the same defect `storeCells`' own comment describes for lists ("lists have no dims cell"), in the one array shape that *has* the cell and leaves it empty.

The requirement is now exactly one dimension, checked in `toCells` and in `okSimpleContainerContents` so the gate and the accessor agree. Only zero-dimensional arrays change behaviour; the multi-dimensional wording is kept verbatim so anything matching that text still matches.

Four tests, each failing on its own with the guard reverted (54 panic reports across them): sweeps of degenerate shapes, of extreme range bounds (`MinInt`/`MaxInt` through `validateRange`'s negative-index rewrite), and of the builtin surface — plus `TestZeroDimensionalArrayIsRefused`, which pins the *fix* rather than the absence of the crash, since a guard that silently accepted the shape and wrote inconsistent dims would pass a no-panic sweep. They call the Go API with nothing recovering in between, because that is the surface with no evaluator behind it; the per-case recover exists only so one bad shape reports itself instead of aborting the run. Not hypothetical — the first version let a panic escape `t.Run` and took the binary down, and two sibling sweeps never reported.

## Open-range splice coverage

A range set is a **splice**, not an element-wise assignment: the replacement may be any length and the sequence grows or shrinks to suit. Two gaps around that, both on the open form, which reaches a different arm of `validateRange` than the closed one:

- **The exact-result battery had no open arm.** `rangeStep` always emitted a 3-cell `'(range from to)`, so all seventeen cases of `TestRangeSpliceResult` — the regression battery for #471's splice-through-the-source-array bug — ran on one arm. #563 was a defect confined to the *other* one. Thirteen open cases now cross the same longer/equal/shorter/empty axes, plus a negative `from`, where the negative-index rewrite and the implicit-end fill-in have to compose.
- **Nothing reused one open path across document lengths** — the property a closed range cannot express. The only test of that shape covered `Get`. `TestOpenRangeTracksInputLengthForEveryOp` builds the path once, outside the loop, and applies it to documents of length 0, 1, 2, 3, 5 and 9 through the Go API. Both sides of the `from=2` boundary are covered: an implementation that ignored `implicitTo` and spliced to a fixed 0 would still agree on `n <= 2`.

Red-proofed together: making `validateRange` ignore `implicitTo` fails 77 subtests.

The lisp file gains the same shapes at the user-facing surface — the one open-range `?set!` case it carried was equal-length, which invites exactly the wrong assumption about matching sizes.

## Acting on an adversarial review

The branch was independently attacked. Eight findings, no shipped behaviour wrong — the holes were in the tests guarding it.

**The test guarding the worst failure could not fail.** `(assert-nil (ignore-errors (parse-path bad)))` passes whether the builtin raises *or* silently returns an empty list, because `ignore-errors` yields `()` on a raise and `()` is also the identity selector's *successful* result. There was no Go test of the builtin at all.

Not harmless: no steps **is** the identity path, so a swallowed error turns a malformed selector into "the whole document" for the very idiom the docstring recommends. Mutating the builtin to swallow the error left the whole suite green — including the test named "parse-path rejects…". Fixed with a sentinel that separates the two cases, plus `parse_path_builtin_test.go` where the error type is directly observable. The string type check was untested too, and is load-bearing: `LSymbol` carries a string payload, so without it `(parse-path '.a)` silently succeeds. Both mutations are now caught by the Go **and** lisp suites.

Also fixed: two comments the first commit made false (`"nothing here is reachable from lisp"`, of a parser it exposed to lisp), a citation to a test that never existed, and a reject-test whose header claimed to compare error messages while comparing only error-vs-no-error.

## Error messages

`failed to parse: -type` on a client-supplied `.content-type` named neither the rule nor the fix — and these paths arrive as *data*, so that message is the only feedback the caller gets. It now appends:

```
failed to parse: -type (a bare key must match [A-Za-z_][A-Za-z_0-9]*;
bracket and quote anything else, e.g. .["my-key"])
```

**Appended, not replaced** — the existing text stays the prefix, so a `handler-bind` matching on it still matches. Over the substrate differential: 17,845 rows are the identical message plus the hint, **0 rows differ any other way that is not already the #566 grammar fix, and 0 accepted selectors changed their result**.

**It took two attempts, and the differential caught the first.** The initial rule was "hint unless the stall is a bracket", which misdiagnosed `.["a"]foo` — that stalls at `foo`, a perfectly good key name, and what is missing is the *dot*. 3,425 rows were being told their keys were misspelled when they weren't. It now fires only on a rune a bare key cannot contain (`-`, a leading digit, `$`, non-ASCII); a stall on `[A-Za-z_]` means a missing separator and stays quiet.

Note the key rule is **unchanged and always has been** `[A-Za-z_][A-Za-z_0-9]*` — byte-identical to substrate's shipped v1 regex, and jq's rule. Underscores have always worked; only the message changed.

## Test plan

- `go test ./...`, CI-exact golangci-lint (v2.6.2) **zero findings**, `elpsvet` (both passes, `-test=false` as CI runs it), `elps fmt -l`, `elps doc -m`, `elps lint` all clean.
- Every new test mutation-proofed; the mutations that previously escaped every test now fail both suites.
- `FuzzParseSelector` clean (~2M executions locally with the new capacity invariant in the loop); `FuzzPathEngine` clean over 880k executions after the dimensionality change; plus the 10 CI shards.
- Value hygiene checked: fresh pointers per call, no spare capacity to alias into, the identity result is a fresh empty list rather than `lisp.Nil()`'s singleton.

## ⚠️ Downstream: this one breaks a substrate tripwire

Unlike #567, adding a **builtin** trips `TestComposedPackageExports`, which pins the exact export list of the composed `elpspath` package:

```
expected: [... "nil-path!", "set-path", "set-path!"]        (14)
actual:   [... "nil-path!", "parse-path", "set-path", ...]  (15)
elpspath exports changed
```

That is the tripwire working as designed. It does not break substrate today (it pins v1.59.0), but **whichever substrate PR bumps the elps dependency must add `parse-path` to that list**. No name collision otherwise — substrate composes seven names, none of them this one.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EgShxM3EoPmeVnq1cmpcPK)_


sam-at-luther:
**Status: what landed after the description above was written** (the description covers the branch through the adversarial-review commits; two more changes followed).

## The selector parser is now an independent hand-written scanner (`33ac2d79`)

The regexp core that had been ported from substrate's `parse.go` (three regexp parsers plus the round-robin scan loop) is replaced by a single left-to-right scanner designed from the grammar: one byte dispatch per step, escape-aware quoted-key scanning, hand-rolled subscript parsing, no backtracking, no substring copies. Public API, grammar, and error-message prefixes are unchanged.

- **Behavior-preserving, measured twice:** a 28,413-input differential (every corpus plus exhaustive 2–3-token adversarial strings) between the old and new implementations found zero differences, and an independent review reconstructed the pre-rewrite parser from git history and ran a 622,620-selector differential — also zero differences.
- **Independence:** substantive line overlap with the substrate original dropped from 88% to 13% — the residue is contract-mandated one-liners (the required error strings, the `.`/empty checks, `TrimSpace`).
- **Performance (CI's interleaved n=10 vs `main`):** `ParseSelector` −67% to −90% sec/op with allocs down 20–67%; `libelpspath` geomeans −19.2% sec/op, −13.1% B/op, −8.3% allocs/op; benchmark gate: 0 at or above the gate.
- `selector_regexp_test.go` (capture-group pins for the deleted regexps) is replaced by `selector_scanner_test.go` — the same cases as (step, consumed-bytes) pairs, the non-overlap property restated as `TestSelectorStepFormsAreDisjoint`, plus coverage for the two whitespace sets and the `selectorBody` newline behavior, now documented and pinned rather than accidental.

## Four review findings fixed (`13dfa2c8`)

A high-effort review of the full diff surfaced four issues, all fixed:

1. **`parse-path` no longer silently truncates a bracket-led selector at a newline.** `parse-path(".[0]\n.password")` previously returned just `'(0)` — a silently widened write target for the docstring's recommended `?set` idiom. It now raises: `selector may not span lines: ".password" would be silently discarded`. The check lives in the builtin only; `ParseSelector` keeps the v1 wart byte-for-byte for downstream parity (now documented on every surface), and the ParseSelector/SelectorSteps grammar-agreement invariants are untouched.
2. **`keySpellingHint` now diagnoses the rune the scan actually stalled on** — `. -key` gets the bracketing hint, a stray `?` no longer draws a key-spelling misdiagnosis.
3. **The empty-args guard names the right parameter** — `parse-path requires a selector argument` (the seven document builtins' messages are unchanged).
4. **`TestOpenRangeTracksInputLengthForEveryOp` now covers `Nil` and `NilMutate` over every length including the error side** (30 → 42 subtests).

All 24 checks are green on `13dfa2c8`; the review fixes cost nothing measurable on any benchmark arm.

---

PR #572, opened 2026-09-03
Clone sorted maps structurally in Fork

### LINKED ISSUE #576: Fork de-aliases two LVals that share one *MapData
## Problem

`Fork` memoises copies per `*LVal` (`forker.vals`) but not per `*MapData`. Two distinct `LVal`s that share one `*MapData` become two independent maps in the fork. This is reachable from pure ELPS because `Quote`, `Splice`, `shallowUnquote` and `opQuasiquote` copy the struct (`*cp = *v`) and keep the same `Native`.

```lisp
(set 'a (sorted-map "k" 1))
(set 'b (quasiquote (unquote a)))   ; struct copy, same *MapData
(assoc! a "x" 99)
(get b "x")                          ; => 99 in the template
```

After `env.Fork()`, a write through the fork's `a` is not visible through the fork's `b`. Same behaviour on `main` (03b789a) and on #572, so this is pre-existing, not a regression of the sorted-map clone.

## Proposed fix

Add `f.maps map[*MapData]*MapData` to `forker` and memoise in `forker.mapData` before the value walk, the same way `f.vals` does for `*LVal`. Add a test that binds one `*MapData` under two `LVal`s (via quasiquote), forks, writes through one name and reads through the other.

Found during the adversarial review of #572.

sam-at-luther:
## Summary

- `forker.mapData` rebuilt every sorted map by enumerating its entries in sorted order and re-inserting them one by one: a sort, one pair list per entry, and incremental growth of the new map, for every map in every fork.
- The stock sorted map is a Go map plus a key-type map, so the fork now copies both maps directly and remaps the values through the fork walk (values must still go through `f.val` so mutable values are copied and aliases/cycles are preserved; the keys are immutable strings and copy as-is). `Set` stored exactly those two things per entry, so contents, key types, aliasing and cycles are unchanged. Embedder map implementations keep the generic entries path.
- The clone seeds the per-`*MapData` memo that #587 (fixes #576) added below this PR, before the entries are walked, exactly as the generic path does, so two headers over one map and a map that reaches itself come out the same through either path (`lisp/fork_mapalias_test.go` runs on this head too).
- The clone uses `make(..., len)` plus a copy loop, not `maps.Clone`, on purpose. Go maps never shrink after deletes and `maps.Clone` copies the table at its high-water mark, so a load-time map filled to 100k entries and pruned to 3 would cost every fork 3.7MB instead of 0.2MB. Found by adversarial review of an earlier revision; `TestForkSortedMapClonePrunedMapIsRightSized` now pins it.
- First of a stack of small, isolated fork/evaluator changes from a profiling study of the interpreter on a 72k-line production phylum running under substrate, where the per-VM fork dominates warm-VM cost. Study: https://claude.ai/code/artifact/05b3209b-890d-4f35-8b1b-fef71f432b74. This PR is the lowest-risk item: it touches one function in `lisp/fork.go` and has no language-semantic surface.

Stacked on #586: the bug-fix PRs (#580, #581, #583, #587, #586) land first, then this and the rest of the perf stack.

## Measurements

Production phylum forked under substrate (pinned CPU, 30 forks per sample, n=3):

| metric | base | this change |
|---|---|---|
| fork B/op | 5.78 MiB | 4.15 MiB (−28%) |
| fork allocs/op | 49.3k | 42.2k (−15%) |
| retained-fork pool wall time | 1.98 s | 1.68 s (−15%) |

New `BenchmarkForkSortedMap` (20 maps × 200 entries, `-cpu 1 -count 6`, interleaved): sec/op −44%, B/op −44%, allocs/op −29%. The allocation numbers are exact; the wall-time number needs a pinned CPU to reproduce. CI benchmark gate: 0 rows at or above the gate.

## Test Plan

- [x] `make test` passes
- [x] `golangci-lint run ./...` reports 0 issues (v2.6.2, matches CI's v2.6)
- [x] `./elps fmt -l ./...` reports no changes
- [x] `./elps lint ./...` reports only the pre-existing `editors/vscode/test/grammar/builtins.lisp` diagnostics, identical on `main`
- [x] `./elps doc -m` reports no missing docs
- [x] `make elpsvet` passes
- [x] `TestForkSortedMapClone` pins entry order, key types, value remapping (mutable value copied, aliased map stays aliased, self-reference points at the clone) and template/fork independence in both directions, including the key-type map (a shared typemap would be a concurrent write to the template under parallel forks)
- [x] `TestForkSortedMapClonePrunedMapIsRightSized` fails on a `maps.Clone` build (3.7MB/op) and passes here (0.25MB/op)
- [x] `lisp/fork_mapalias_test.go` (from #587) passes through the structural clone

### DIFF
--- a/lisp/fork.go
+++ b/lisp/fork.go
@@ -606,6 +606,51 @@ func (f *forker) mapData(md *MapData) *MapData {
 		f.maps[md] = cp
 		return cp
 	}
+	if sm, ok := md.mapBacking.(sortedmap); ok {
+		// The stock sorted map is a Go map plus a key-type map, so it can be
+		// cloned structurally: copy both maps and remap the values through
+		// the fork walk.  This is what Set would store for every entry, so
+		// the contents, key types, aliasing and cycles come out identical to
+		// the enumerate-and-reinsert path below, without the sorted entry
+		// list, the per-entry pair cells, or the incremental map growth.
+		//
+		// make(..., len(sm.m)) right-sizes the clone on purpose.  Go maps
+		// never shrink after deletes and maps.Clone copies the table at its
+		// high-water mark, so a load-time map filled to 100k entries and
+		// pruned to 3 would cost every fork 3.7MB instead of 0.2MB (pinned
+		// by TestForkSortedMapClonePrunedMapIsRightSized); the per-entry
+		// copy costs nothing measurable over maps.Clone.
+		//
+		// The key-type map is copied verbatim rather than rebuilt from the
+		// entries (as detach's mapData still does).  An entry in tm with no
+		// partner in m cannot exist today (Set and Del keep the two in
+		// step), so nothing is laundered here, and a future desync shows
+		// up in the fork instead of being hidden by the rebuild.
+		//
+		// Measured on a 72k-line production phylum whose template holds
+		// large load-time sorted maps (rating tables, templates): this one
+		// branch removed 28% of the bytes and 15% of the allocations of
+		// every fork, and 15% of the wall time of building a retained pool
+		// of forked VMs.  Embedder map implementations keep the generic
+		// path: their entry enumeration is the only contract they offer.
+		nm := sortedmap{
+			m:  make(map[interface{}]*LVal, len(sm.m)),
+			tm: make(typemap, len(sm.tm)),
+		}
+		// Memo before descending, as the generic path does (issue #576):
+		// an entry may reach back to md.  The Go maps inside nm are
+		// references, so filling them after the memo stores the *MapData
+		// is visible through it.
+		cp := NewMapData(nm)
+		f.maps[md] = cp
+		for k, v := range sm.m {
+			nm.m[k] = f.val(v)
+		}
+		for k, t := range sm.tm {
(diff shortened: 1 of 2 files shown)


sam-at-luther:
**Benchmark gate on 8b8d043: RUNNER-UNFIT (exit 3), not a regression from this diff.**

The failing check is `Benchmark Comparison` / `Required: benchmark`. Only one row tripped the gate:

| Row | Delta | Spread |
|---|---|---|
| `PackageGetFunParallel-2` | +39.66% | ±37% |

- That benchmark measures package symbol lookup. This PR changes only `forker.mapData` (the sorted-map clone inside `Fork`), so the code that benchmark runs is not on the diff.
- A ±37% spread is above the gate's own 30% unfit threshold. The gate says the runner could not measure that row, and `scripts/benchgate.sh` names this row as the repo's noise-floor exemplar.
- Every other row, including the new `ForkSortedMap` benchmark, is inside the gate.

Per the gate doc the remedy is one re-run of the failed jobs. Re-running now; a second RUNNER-UNFIT on the same row will be treated as real and investigated.

---

PR #573, opened 2026-09-03
Keep the scanner's one-rune peek buffer instead of reallocating it

sam-at-luther:
## Summary

- **The bug.** `Scanner.Peek` buffers a lookahead rune with `s.peek = append(s.peek, peek)`, and `ScanRune` consumed it with `s.peek = s.peek[1:]`. On a one-element slice, `[1:]` leaves length **and capacity** zero, so the next `Peek`'s append had no slot to reuse and allocated a fresh backing array — one allocation for every rune of source scanned.
- **The fix.** `ScanRune` shifts the buffer down in place, `s.peek = s.peek[:copy(s.peek, s.peek[1:])]`, which keeps the backing array the first `Peek` allocated. This form was chosen over a bare `s.peek[:0]` after adversarial review: today the buffer never holds more than one element (the sole append is inside `Peek` behind an early return on `len(s.peek) > 0`, and nothing else writes the field), but `[:0]` would silently drop runes if a multi-rune lookahead ever landed, with no test to catch it. No slot is pre-allocated in `newScannerBuf`: append to a nil slice allocates the one-element array on the first `Peek`, and a constructor-time `make` only added an allocation to scanners over empty input.
- **Scope.** One line of production code plus a regression test. No semantic surface: the scanner's observable behaviour is unchanged (differential harness of 1,260 inputs × 7 Peek/ScanRune interleave patterns across scanner, lexer, parser, fault-tolerant parser and formatter was byte-identical to base), and this is the lowest-risk item of the stack.
- Stacked on #572 — this PR targets `claude/elps-perf-optimization-jcgfd2`, not `main`.
- Background study: https://claude.ai/code/artifact/05b3209b-890d-4f35-8b1b-fef71f432b74

## Measurements

`go test -run '^$' -bench 'BenchmarkParse50KB$|BenchmarkParser' -benchmem -count=4 -cpu 1 ./lisp/ ./parser/rdparser/`, compared with `benchstat`. Machine: Intel Xeon @ 2.80GHz, linux/amd64.

### `lisp` — BenchmarkParse50KB

| metric | base | new | delta |
|---|---|---|---|
| sec/op | 10.163m | 7.210m | **-29.05%** (p=0.029 n=4) |
| B/s | 4.826Mi | 6.795Mi | +40.81% (p=0.029 n=4) |
| B/op | 4.449Mi | 3.665Mi | -17.62% (p=0.029 n=4) |
[...]

PR #574, opened 2026-09-03
Share a function's sealed cells across forks

sam-at-luther:
## Summary

- **What the fork did.** `forker.val` copies every unsealed value, and an `LFun` is unsealed: it copied the header, rebuilt the `funData` with the captured environment remapped, and then — in the generic tail — allocated `make([]*LVal, len(v.Cells))` and filled it with `f.val(c)` per child. A function's cells are `[formals, body...]`, and for a `defun`/`lambda` parsed from source every child is a sealed program node that `f.val` returns unchanged. So the fork allocated one slice per function purely to hold the same pointers.
- **The change.** In the `len(v.Cells) > 0` block, before the generic copy: for `LFun`, probe every child; if each maps to itself, share the template's backing array (`cp.Cells = v.Cells[:len:len]`) and return. Otherwise fall through unchanged. `f.val` memoises, so a child that is *not* sealed is copied on the probe and the generic loop below reuses that memoised copy — the fallthrough costs a comparison per cell, not a second allocation.
- **The share is a three-index reslice, not a bare header copy.** Adversarial review found that `cp.Cells = v.Cells` also shares the slice's spare *capacity*: an embedder that had grown a template function's cells with `append` (no ELPS-reachable trigger, every constructor here allocates exact capacity) would have had a fork-side `append` and a template-side `append` land in the same spare slot. This is the hazard `forker.val`'s len==0 branch and `libschema.markValidator` already guard against. The clamp is free (no allocation, no copy) and `TestForkSharedFunctionCellsClampCapacity` fails on the bare header copy with `cap 4, len 3`.
- **The invariant it relies on, and the audit.** A function's cells are written once at construction, over backing the constructor itself allocated, and not rewritten afterwards by anything in this repository.
  - **(a) Post-construction `Cells` writes.** Grepping the non-test Go sources for `.Cells = append`, `.Cells[i] = ` and `.Cells = ` across the repository turns up only construction-time sites and sites on non-function values: `env.Lambda`, `registrationFunValue`, `FunInPackage`/`MacroInPackage`/`SpecialOpInPackage` and `Copy` build the slice they install; `opExpr` fills a formals list it just made; `libschema.markValidator` reinstalls an exact-length slice it allocated while tagging the value; `macroDeftype` appends onto an `SExpr` it is assembling, not a function; `builtins.go`'s appends are on call forms and vectors; the `args.Cells = args.Cells[1:]` sites decap per-call arglists with fresh backing. Outside `lisp/`, only the parser appends, onto its own nodes. The review re-swept every write to `LVal.Cells` across `lisp/`, `lisp/x/`, `lsp/`, `analysis/`, `lint/`, `formatter/`, `minifier/`, `elpsutil/`, `elpstest/`, `cmd/`, `parser/`, `internal/`, `mcpserver/` and `astutil/`: none targets an `LFun`. A loaded env plus a fork of it held 271 reachable `LFun` values, all with exact capacity.
  - **(b) Builtins vs lambdas.** A builtin's cells are `[formals, docstring]`. The formals are sealed (`sealDefaultFormals` at package init), but `Cells[1]` is a fresh **unsealed** `String` built per registration. The probe therefore fails at cell 1 and builtins keep the generic copy — correct, since their docstring is mutable storage. Only lisp-defined functions whose formals and body came from the reader take the shared path.
  - **(c)** The zero-length branch is untouched.
- Stacked on #573 (scanner peek), which is stacked on the sorted-map clone in #572. **Base this PR on `claude/elps-perf-optimization-jcgfd2-scanner-peek`, not `main`.** Independent of the sibling #575 on the same base — this touches only `lisp/fork.go` and adds one test file.
- Pre-existing, not a regression (identical on `main`): `s:deftype` validators stop validating after a `Fork`, because `libschema.isValidator` keys on pointer identity of a native marker the fork walk re-headers. Tracked and being fixed in #579.
- Study: https://claude.ai/code/artifact/05b3209b-890d-4f35-8b1b-fef71f432b74

## Measurements

`benchstat` over two interleaved base/change rounds of `-count=5 -cpu 1` (n=10 per arm) on `./lisp/`:

```
                  │    base     │               change                │
                  │  allocs/op  │  allocs/op   vs base                │
ForkManyFunctions   3.112k ± 0%   2.612k ± 0%  -16.07% (p=0.000 n=10)
ForkSortedMap       20.83k ± 0%   20.83k ± 0%        ~ (p=0.526 n=10)

                  │     B/op     │     B/op      vs base               │
ForkManyFunctions   437.6Ki ± 0%   429.8Ki ± 0%  -1.79% (p=0.000 n=10)
ForkSortedMap       2.466Mi ± 0%   2.466Mi ± 0%       ~ (p=0.579 n=10)

                  │   sec/op    │   sec/op     vs base               │
ForkManyFunctions   506.2µ ± 5%   510.8µ ± 3%       ~ (p=0.529 n=10)
ForkSortedMap       5.130m ± 7%   4.842m ± 6%       ~ (p=0.165 n=10)
```

Exactly 500 allocations disappear from the 500-function template, which is the whole of the effect. `ForkSortedMap` is the control and is unmoved. Wall time is within noise on a shared runner. CI benchmark gate: 0 rows at or above the gate.

`./lisp/lisplib`, `-count=6 -cpu 1`:

```
EnvConstruction/mode=fork   1.242k -> 1.235k allocs/op   -0.56% (p=0.002 n=6)
```

Small here only because the in-repo template defines few lisp functions. Measured previously on a 72k-line production phylum (2.5k defuns, 1.3k lambdas) under substrate, this removes about 8% of fork allocations on top of the sorted-map clone.

## Test Plan

- [x] `go build ./...`
- [x] `go test ./lisp/...`
- [x] `go test -race -run Fork ./lisp/ ./lisp/lisplib/`
- [x] `go test -tags=elpscheck -run Fork ./lisp/`
- [x] `./elps fmt -l ./...` — no files listed
- [x] `./elps doc -m` — all functions and exports documented
- [x] `go test -run Fuzz ./lisp/` (seed corpora + recorded crashers)
- [x] Differential battery (review): 36 hand-written programs plus all 21 in-repo `.lisp` sources, each run in a fork of a warm template and in a fresh env, on HEAD and HEAD~1: no divergence

New tests in `lisp/fork_cells_test.go`:

- `TestForkSharesSealedFunctionCells` — a sealed `defun`, forked: the fork's cell slice shares the template's backing array and every cell pointer, has **no spare capacity**, and the function value, `funData` and closure env stay distinct. Fails on base with `fork allocated a new cell slice for an all-sealed function`.
- `TestForkSharedFunctionCellsClampCapacity` — grows a template function's cells with `append` (arming spare capacity), forks, and asserts the fork's slice is clamped and that appends on either side cannot reach the other. Fails on a bare header copy with `fork's cells have cap 4, len 3`.
- `TestForkSharedFunctionCellsConcurrentCalls` — eight forks calling shared-cell functions (positional, `&optional`, `&key`, `&rest`, docstring, macro, closure) in parallel under `-race`, results compared against the template.
- `TestForkCopiesUnsealedFunctionCells` — an unsealed `(lambda (x) x)` still gets a private slice; runs the existing `forkAuditor` over the whole reachable graph.
- `BenchmarkForkManyFunctions` — 500 sealed `defun`s, measuring `env.Fork()`.

### DIFF
--- a/lisp/fork.go
+++ b/lisp/fork.go
@@ -34,7 +34,9 @@ import (
 //     environment remapped onto the fork's copies so closures close over
 //     fork state, never template state.  Builtin Go funcs travel by
 //     reference (Go code is immutable).  Formals and lambda bodies are
-//     sealed parser output and therefore shared.
+//     sealed parser output and therefore shared, and when every cell of a
+//     function is shared the cell backing array is shared too, clamped to
+//     its length so neither side can append into the other (see val).
 //   - LNative payloads: SHARED by default.  A native handle cannot be
 //     hermetically copied by the kernel (the same reason detach rejects
 //     it), and most native payloads held by loaded environments are
@@ -504,6 +506,44 @@ func (f *forker) val(v *LVal) *LVal {
 		}
 	}
 	if len(v.Cells) > 0 {
+		if v.Type == LFun {
+			// A function's cells are [formals, body...], written once at
+			// construction over backing that the constructor itself
+			// allocated (env.Lambda, registrationFunValue, FunInPackage and
+			// friends) and not rewritten afterwards by anything in this
+			// repository.  So when every child maps to itself -- all sealed
+			// program nodes, which is the shape of every defun and lambda
+			// parsed from source -- the fork can share the template's
+			// backing array outright instead of allocating a slice per
+			// function only to refill it with the same pointers.
+			//
+			// The share is a three-index reslice, for the reason the
+			// len==0 branch below and libschema's markValidator both spell
+			// out: a slice header carries its spare CAPACITY, and an append
+			// on either side of a shared header writes the other side's
+			// backing array.  Every constructor here allocates exact
+			// capacity, so the clamp is a no-op today; it is what keeps
+			// "nothing appends to a function's cells" from being
+			// load-bearing for embedder Go code, which can reach the
+			// exported Cells field and is not under cmd/elpsvet.
+			//
+			// f.val memoizes, so a child that is NOT sealed (a builtin's
+			// unsealed docstring cell, say) is copied on the first probe
+			// here and the generic copy below reuses that memoized copy:
+			// the fallthrough costs a comparison per cell, not a second
+			// allocation.
+			shared := true
+			for _, c := range v.Cells {
+				if f.val(c) != c {
+					shared = false
+					break
+				}
+			}
+			if shared {
+				cp.Cells = v.Cells[:len(v.Cells):len(v.Cells)]
+				return cp
+			}
+		}
 		cells := make([]*LVal, len(v.Cells))
 		for i, c := range v.Cells {
 			cells[i] = f.val(c)
--- a/lisp/fork_cells_test.go
+++ b/lisp/fork_cells_test.go
@@ -1,0 +1,422 @@
+// Copyright © 2026 The ELPS authors
+
+package lisp
+
+import (
+	"errors"
+	"fmt"
+	"strconv"
+	"strings"
+	"sync"
+	"testing"
+)
+
+// Tests for the function-cell sharing branch in forker.val.  A function's
+// Cells are [formals, body...], written once by the constructor that
+// allocated them and never grown or rewritten afterwards, so a fork whose
+// children all map to themselves — the shape of every defun and lambda
+// parsed from source, whose formals and body are sealed program nodes —
+// can share the template's slice instead of allocating a fresh one to hold
+// the same pointers.
+//
+// Package lisp cannot import the parser (import cycle), so these tests
+// apply SealAST by hand exactly where the parser would, the same technique
+// TestForkOwnership_SealedSharedAcrossRuntimes uses.
+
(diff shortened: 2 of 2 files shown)


sam-at-luther:
Lint & Test is red on 4b52b7e, and the failure is not this PR's.

The failing test is `cmd/benchgate` `TestBurninTakesRealTime`, in the `make test-elpscheck` pass:

```
burnin_test.go:356: 4x the work took 7.448405ms against 7.91908ms for the smaller loop
```

It compares one wall-clock sample of a 2000-iteration loop against one sample of an 8000-iteration loop, both under 8ms, and asserts the ordering. A single scheduler stall during the small loop inverts it. The same test passed in the plain `go test ./...` pass of the same job, and this PR touches only `lisp/`. Locally the full suite passes on this head.

No fix exists on main yet. I have filed the flake with a proposed best-of-N patch and am re-running the job once.

---

PR #584, opened 2026-09-03
Memoise sorted maps, bytes and native payloads per payload in Fork so aliased headers stay aliased

### LINKED ISSUE #576: Fork de-aliases two LVals that share one *MapData

sam-at-luther:
## Summary

`Fork` memoised copies per `*LVal` only. Two headers over one payload -- the shape `Quote` (reached from quasiquote through `doUnquoteValue`), `Splice`, `shallowUnquote` and `FunRef` produce, since each copies the struct and keeps the `Native` -- were rebuilt once per header. After `(set 'b (quasiquote (unquote a)))`, a write through the fork's `a` was invisible through the fork's `b`: a program that behaved one way on the template and another way in every fork of it. Pre-existing on `main` (03b789a); found during the adversarial review of #572.

One commit per payload type:

1. **Sorted maps** (`*MapData`). `forker` gains a memo keyed on the `*MapData`, seeded before the entries are walked, on both `mapData` paths that exist here (generic enumerate-and-reinsert, degenerate no-backing); the structural clone #572 adds above this PR seeds the same memo. A map reaching itself through a second header showed the same thing one level down: the `*LVal` memo bounded the walk (finitely many headers) but not the clones, one per header, each nesting the next. Closes #576.
2. **Bytes and native payloads**, found by the adversarial review of commit 1. Two headers over one `*[]byte` were copied once per header; two headers over one `NativeCloner` payload were cloned once per header, so an accumulator the template held once became two independent accumulators in the fork. Both get a memo keyed on the payload. Bytes cannot form a cycle, so the memo is filled after the copy. Native payloads are memoised only when the payload is a pointer (identity is what aliasing means); a replacer now sees such a payload once per fork, which its documented contract allowed ("may be called more than once"), and the `NativeCloner` and `ForkWithNativeReplacer` docs say so.

The same de-aliasing exists in `copy`/`detach`, which memoise per `*LVal` too; tracked in #585 rather than widened into this PR.

Stacked directly on #583, ahead of the perf PRs: bug fixes land first. #586 sits on this; #572 and the rest of the perf stack sit on #586.

## Measurements

`BenchmarkForkSortedMap`, `-count=4 -cpu 1`, base vs commit 1:

| metric | base | change |
|---|---|---|
| allocs/op | 20833 | 20837 |
| B/op | 2.586 MB | 2.591 MB |
| sec/op | within noise | within noise |

Four allocations per fork for the memo tables. CI benchmark gate on commit 1: 0 rows at or above the gate.

## Test Plan

`lisp/fork_mapalias_test.go`, all four **fail on the parent commit** and pass here:

```
--- FAIL: TestForkPreservesMapDataAliasAcrossHeaders
    fork de-aliased the shared map: a=0xc0000931c0 b=0xc0000931a0
    fork write through a not visible through b: got (), want 7
--- FAIL: TestForkSelfReferenceThroughAliasedHeaderStaysAliased
    forked self entry is a different map (0xc000440a00) from its container (0xc000440a20)
--- FAIL: TestForkPreservesBytesAliasAcrossHeaders
    fork de-aliased the shared bytes: a=0xc00028ae40 b=0xc00028ae28
    fork write through a not visible through b: got 3, want 4
--- FAIL: TestForkClonesANativePayloadOncePerPayload
    fork de-aliased the shared payload: a=0xc0000a46e0 b=0xc0000a46e8
    payload cloned 2 times, want 1
```

The first and third are the ELPS-level shapes (template write visible through the alias, fork write visible through the fork's alias, template untouched by the fork's write). The second is the self-reference through a second header; the fourth counts clones of a `NativeCloner` reached through two headers.

- [x] `go build ./...`
- [x] `go test ./...`
- [x] `go test -race ./lisp/`
- [x] `go test -tags=elpscheck ./lisp/`
- [x] `golangci-lint run ./...` -- 0 issues
- [x] `./elps fmt -l ./...` -- no files listed
- [x] `./elps doc -m` -- all documented

Closes #576

### DIFF
--- a/docs/fork.md
+++ b/docs/fork.md
@@ -281,9 +281,10 @@ r := &elpstest.Runner{
   mutable values pointer-distinct with identical content, template
   aliasing and cycles reproduced in the fork.
 - `lisp/fork_mapalias_test.go` pins aliasing one level below the `*LVal`:
-  two headers over one `*MapData` (the shape `(quasiquote (unquote a))`
-  makes) fork to two headers over one clone, and a map that reaches itself
-  through such a header terminates (issue #576).
+  two headers over one `*MapData`, one `*[]byte` or one native payload (the
+  shape `(quasiquote (unquote a))` makes) fork to two headers over ONE clone,
+  and a map that reaches itself through such a header closes onto its own
+  clone rather than nesting a fresh one per header (issue #576).
 - `lisp/lisplib/fork_test.go` proves bidirectional isolation over real
   parsed programs two ways: observable mutations (neither side sees the
   other's writes) and a full-state fingerprint (structure-only hash of
--- a/lisp/fork.go
+++ b/lisp/fork.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"reflect"
 )
 
 // Template-environment forking (issue #380).
@@ -82,6 +83,10 @@ import (
 // Under `copy` the duplicate stays inside one environment, so a payload
 // written for that path alone still has to meet the stricter bar to be
 // fork-safe.
+//
+// Fork calls it once per payload: a pointer payload reachable through
+// several headers is cloned once and the clone shared by all of them, as
+// the payload was on the template (issue #576).
 type NativeCloner interface {
 	CloneNative() interface{}
 }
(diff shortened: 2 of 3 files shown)


sam-at-luther:
This PR was not merged by a person. GitHub marked it merged automatically when its base branch (`claude/elps-perf-optimization-jcgfd2-lambda-env`, #578) was force-pushed to a head that already contained these two commits, as part of reordering the stack to land the bug fixes first. A merged PR cannot be reopened or retargeted, so the same branch and commits continue in #587, based on #583.

---

Issue #579, opened 2026-09-03
Schema validators from s:deftype stop validating after Fork


sam-at-luther:
## Problem

`libschema.isValidator` (lisp/lisplib/libschema/libschema.go, around line 434) recognises a validator by pointer identity against `validatorMarker`, a package-level `lisp.Native(&validatorTag{})` stored in the validator's `Cells[2]`. `forker.val` gives every `LNative` a fresh `*LVal` header, so in a fork the marker cell is a different pointer and the credential is revoked. Every `s:deftype` validator defined in a template fails in the fork:

```lisp
(s:deftype "T" s:int)
(s:validate T 3)
```

| Where | Result |
|---|---|
| fresh env | `()` |
| fork of a template that ran the deftype | `bad-arguments: Value is not a schema constraint: #<builtin>...` |

Identical on `main` (03b789a) and on the perf stack (#572 to #578), so this is pre-existing. Any embedder using `s:` schemas from a fork pool (substrate's preheater) has silently broken validation.

## Proposed fix

Make the validator credential fork-transparent: either seal the marker so the fork walk shares it, or key `isValidator` off a property the walk preserves (for example the `validatorTag` native payload type rather than the header pointer). Add a test that defines a validator in a template, forks, and validates in the fork; it must pass, and must fail on the current code.

Found by the adversarial review of #574.

PR #589, opened 2026-09-03
Add elpstest.RunForkCheck, a template/fork oracle with a regression per fork bug that shipped

sam-at-luther:
## Summary

- An embedder such as substrate loads a program once into a template and runs every transaction on a fork of it. Three fork bugs reached that pattern with no test that models it: #576 (two headers over one sorted-map, bytes or native payload forked to two payloads, so a write through one name was invisible through the other, on forks only), #579 (a libschema validator lost its credential across a fork) and #381 (a fork shared the template's libtesting suite). Each was caught by hand and each fix carries a test of its own shape. This adds the model itself, so the next bug of the class fails a test before it reaches an embedder.
- `elpstest.RunForkCheck` takes a program and the transactions a caller would run, and holds the template/fork model to three properties, each against a reference that never calls `Fork`, so a `Fork` bug cannot hide in the oracle:
  - **parity**: a transaction on a fork gives the same result, and leaves the same reachable state, as on a cold environment that loaded the program itself;
  - **aliasing**: "same mutable payload" holds for the same pairs of reachable positions in template and fork;
  - **isolation**: no per-fork payload is shared with the template or with another fork, the template is untouched after a fork's transaction, a later fork is pristine.
- "Reachable" includes the environment a closure captured (its bindings and its parents'), walked through `internal/funraw`, since that is the state a fork must copy and a transaction mutates through the closure. Every check also runs one hop deeper, on a fork of a fork (#579 was a one-hop-only fix once). Payload identity follows `docs/fork.md` and Fork's own memo: sorted-map storage, bytes storage, list and vector cells, captured environments and pointer `NativeCloner`s are per-fork; sealed values and plain natives are shared by design and carry no identity. `ForkOptions` lets a check exercise `ForkWithNativeReplacer` / `ForkWithContext` the way the embedder does.
- `elpstest/forkcheck_test.go` carries one `ForkCheck` per shipped bug, a closure-state shape, and the loaded-program shape the existing fork tests use. Internal tests pin that each oracle can see what it exists to see, and that the walks are linear on shared (diamond) graphs.
- Docs: a bullet under `docs/fork.md` → Verification, including what the oracles cannot see (a native's contents, package metadata beside the symbol table).

Two commits: the harness, then the adversarial-review pass (closure environments in every oracle, seen-guarded walks, pointer-only native identity, fork-vs-fork check, `ForkOptions`, doc corrections, stronger tests).

Stacked on #587 (needs its per-payload memo to pass). Test-only: no change to `lisp/`.

## Verified to fail where the bugs lived

The same three files, dropped onto the pre-fix trees:

| Tree | Result |
|---|---|
| parent of #587 (a5c3e69) | `SortedMapAliasAcrossHeaders`, `BytesAliasAcrossHeaders`, `NativeAliasAcrossHeaders` fail: "alias structure differs from the template" |
| `main` (03b789a) | those three, plus `SchemaValidatorCredential` ("tx[0] fork: result differs from the cold run") and `TestingSuitePerFork` (duplicate registration on the shared suite) |
| this branch | all green |

## Test Plan

- [x] `go build ./...`, `go test ./...`
- [x] `go test -race ./elpstest/`
- [x] `golangci-lint run ./...` — 0 issues
- [x] `make elpsvet`
- [x] Internal anti-vacuity tests: two names over one map vs two equal maps render the same state and different alias signatures; nested aliases carry the same numbers under both bindings; a 40-deep diamond chain grows the signature linearly; an environment shares every payload with itself and none with a cold reload; a write moves the state rendering; a write through a closure moves it and the captured environment is a payload

### DIFF
--- a/docs/fork.md
+++ b/docs/fork.md
@@ -293,6 +293,19 @@ r := &elpstest.Runner{
 - `lisp/fork_ownership_elpscheck_test.go` pins the checker model: sealed
   cross-runtime sharing sanctioned, mutable cross-runtime leaks still
   panic.
+- `elpstest.RunForkCheck` is the embedder-facing harness: give it a program
+  and the transactions a caller would run, and it holds the template/fork
+  model to three properties against a reference that never calls `Fork`.
+  Parity: each transaction's result, and the whole reachable state after
+  it, must match a cold environment that loaded the program itself.
+  Aliasing: "same object" must hold for the same pairs of reachable
+  mutable payloads in template and fork. Isolation: no per-fork payload
+  shared with the template, the template untouched after a fork's
+  transaction, a later fork pristine. Every check also runs one hop
+  deeper, on a fork of the fork. `elpstest/forkcheck_test.go` carries one
+  `ForkCheck` per fork bug that shipped (#576 for sorted maps, bytes and
+  native cloners; #579; #381), each verified to fail on the tree it
+  shipped in. New embedder shapes go there.
 - Correctness at production scale (POC, issue #380): transaction results
   byte-identical between forked and fresh-loaded environments; a full
   phylum unit-test suite fork-served with identical results.
--- a/docs/fork.md
+++ b/docs/fork.md
@@ -296,16 +296,23 @@ r := &elpstest.Runner{
 - `elpstest.RunForkCheck` is the embedder-facing harness: give it a program
   and the transactions a caller would run, and it holds the template/fork
   model to three properties against a reference that never calls `Fork`.
-  Parity: each transaction's result, and the whole reachable state after
-  it, must match a cold environment that loaded the program itself.
-  Aliasing: "same object" must hold for the same pairs of reachable
-  mutable payloads in template and fork. Isolation: no per-fork payload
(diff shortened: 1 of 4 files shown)


sam-at-luther:
**Benchmark gate on cb1fdb8: not this PR's.** The one row at the gate is `PackageGetFunParallel-2` at +35.8%, base 7.48n ±2% vs PR 10.16n ±27%. This PR changes no code under `lisp/` (it adds `elpstest/forkcheck*.go` and a docs bullet), so the interpreter the benchmark measures is byte-identical on both arms.

The same benchmark is bimodal on the runner, in both directions: on #588 (also a non-`lisp/` change) the earlier run reported base 10.23n ±27% vs PR 7.52n, a "−26.5%" in the other direction, and the run before that reported base 7.57n ±42%. Its ±27% sits just under the 30% variance ceiling the gate uses to discount noisy rows, which is why it counted here.

Re-running the failed job once. If the gate keeps catching this row on `lisp/`-free PRs, the class fix is to drop `PackageGetFunParallel` from the gated set or raise the ceiling for parallel rows; that is a separate PR.

---

PR #597, opened 2026-09-04
Pre-size known-length results in slice, make-sequence, join and elpspath

### LINKED ISSUE #596: Pre-size known-length allocations in slice coercion, make-sequence, string:join and the libelpspath copy/iterator/range paths
## Problem

Several builtins build a result whose final size is known before the first element is produced, but grow it from a nil slice, an empty map or an ungrown buffer one element at a time. Each pays a chain of reallocations and copies that the size in hand would avoid. Follow-up to #591, which did the same for one site.

## Sites

- `libelpspath` `copyMapExcept` (behind `?set!`, `?nil`, `?del` on maps): copies into an empty stock sorted map; the entry count is the length of the entries list it walks.
- `slice` coercing a list window to `'string` or `'bytes`: one byte per cell into a nil byte slice.
- `make-sequence` with integer start, stop and step: the element count is computable before the loop.
- `libelpspath` iterator `Get` and range-path nil rewrite: one result per element of a known horizon or window.
- `string:join`: a buffer with no `Grow`; the joined length is one pass over the parts away.
- Also surveyed, to be kept only if they measure: the list arm of `select`/`reject` (the vector arm already sizes), and the package externals list grown per registration in `AddBuiltins`/`AddSpecialOps`/`AddMacros`.

## Acceptance criteria

- Each kept site shows a measurable benchstat gain (time or allocs, beyond a few percent, p < 0.05) on a Go benchmark that exercises it at realistic sizes; a site that measures negligible is not changed, and the numbers are recorded in the PR.
- Semantics are byte-identical for every program: same values, same error text, same evaluation order. Where a length is computed by a pass over the input, that pass must not observe or raise anything the original loop would not have.
- Any sized-map hook exported from package `lisp` stays tiny and honours the string-key-only contract of the existing `emptyForStringKeys` path.
- The production phylum harness in substrate passes under both runners and its cold-load/fork benchmarks do not regress.

sam-at-luther:
## Summary

Five sites built a result whose final size was known before the first element and grew it one element at a time. Each now starts at that size. One commit per site, each kept only because it measured; the ones that did not are listed below with their numbers.

- `libelpspath` `copyMapExcept` (behind `?set`, `?nil`, `?del` on maps): the copy is built into `lisp.SortedMapSized(n)`, a new exported constructor sized like `make(map, n)`. It sizes the entry table only and leaves the key-type map at its zero size, exactly as `newmap` does, so it is correct for symbol keys as well as string keys. A negative `n` is clamped to zero rather than panicking the way `make(map, n)` does. `emptyForStringKeys` now delegates to the same internal helper.
- `slice` coercing a list window to `'string` or `'bytes`: the byte slice starts at the window's cell count, except for an empty window, which keeps its nil slice (see Review fixes).
- `make-sequence` with integer start, stop and step: `makeSequenceCap` returns the capacity for a forward range, bounded both by the runtime's allocation limit and by an absolute `maxSequencePresize`. A backwards or empty range sizes nothing at all (see Review fixes).
- `libelpspath` iterator `Get` and range-path nil rewrite: result slices start at the horizon or window length.
- `string:join`: the type-checking pass also sums the lengths, and a `strings.Builder` is grown once to the joined size.

Each kept site ships a benchmark at 10/100/1000 elements.

## Review fixes

An adversarial review found three ways the "sizing is invisible" claim did not hold. Each is fixed in the commit that introduced it, and pinned by a test.

**1. `slice` turned an empty byte result from nil into empty non-nil.** `libjson`'s `encodeBytes` emits JSON `null` for a nil `[]byte` and `""` for an empty non-nil one — a stated v1.13.0 compatibility contract. `make([]byte, 0, 0)` is not nil, so `(json:dump-string (slice 'bytes '() 0 0))` changed from `"null"` to `"\"\""`, along with an empty window taken out of a non-empty input or a vector, those values nested inside maps and lists, and `json:dump-bytes`. Both coercion arms now leave the slice nil when the window is empty; the `'string` arm is kept symmetric even though `string(nil)` and `string([]byte{})` are both `""`. Pinned by `lisp/slice_bytes_empty_test.go` (7 of its cases were red before the fix).

**2. `make-sequence` sized a backwards range off a positively-wrapped difference.** `stop - start` wraps *negative* only when `stop` really is the larger of the two. Run backwards it wraps *positive*: `(make-sequence 1 -9223372036854775808)` has a difference of `MaxInt64`, so the size was computed for an empty sequence. At the default limit that reserved about 80MB for a call returning `()`; under a raised `WithMaxAlloc`, the capacity exceeded what a slice can hold and `make` panicked, failing a call that used to succeed. The direction is now checked before the subtraction. Pinned by `lisp/make_sequence_overflow_test.go`: one test asserts `()` with no error under a raised limit, the other measures bytes allocated per call at the default limit (84,019,804 before the fix, 133,712 after — the same figure as `main`).

**3. The capacity is now bounded absolutely, not just by `MaxAlloc`.** Clamping only to the allocation limit left the same divergence on the *forward* side: `MaxAlloc` is a number the embedder picks and nothing obliges it to be a count of elements that can exist, so `(make-sequence 0 9223372036854775807)` under `WithMaxAlloc(1<<50)` asked `make` for a capacity beyond what a slice can address and failed immediately as `internal-panic: makeslice: cap out of range`, where on `main` it grows until memory runs out. Different observable behaviour, which this PR is not supposed to have. `makeSequenceCap` now also clamps to `maxSequencePresize` (1<<20 elements): the capacity is only a hint, the loop grows past it, `append` doubles, and beyond a million elements the growth being saved has amortised to nothing — so the bound keeps every measured benefit while making it impossible to hand `make` a number derived from a limit rather than from a real element count. Measured pre- and post-bound over `SliceToString`/`SliceToBytes`/`MakeSequenceInt` at 10/100/1000: allocations identical sample for sample, sec/op unchanged (geomean +0.04%, every arm p >= 0.16). Pinned by `lisp/make_sequence_cap_test.go`, which hands every computed capacity to `make` and fails if `make` rejects it — the panic is pinned directly, without running a loop that does not terminate.

Also from the review: `lisp.SortedMapSized` is exported and its argument is the kind of number a caller computes from a subtraction, so a negative `n` is clamped instead of panicking inside `make`. Pinned by `lisp/sortedmap_sized_test.go`.

The reviewer's remaining finding — a possible `CopyMapWide/10` regression — did not reproduce; see Measurements.

## Tried and dropped

Measured the same way; below the bar the user set (no negligible-effect changes).

| Site | Result | Why dropped |
|---|---|---|
| `select`/`reject` list arm | lambda predicate: time ~, allocs -0.1% to -2.5%; only the builtin-predicate/10 arm reached -8.5% | With a real predicate the calls dominate; the vector arm's presize is not worth mirroring for this |
| `AddBuiltins`/`AddSpecialOps`/`AddMacros` externals list (`slices.Grow` once per pass) | `EnvInit` time ~ (p=0.97), allocs -0.25% | Eight regrows per env init are invisible |
| iterator `Set`/`Delete`/`Nil` result slices | `IterPath/nil` time ~ (p=0.12 to 0.28), allocs -0.07% to -2.7% | The per-element copy dwarfs the slice growth |

## Semantics

Every program produces the same values, the same error text and the same evaluation order as before.

- `copyMapExcept`: only the initial table size changes; the entries list is the same one it already walked, `Set` is called with the same keys and values in the same order, and a symbol key still records its key type.
- `slice` coercion: `appendBytes` validates every cell before the first byte is written, as before, so the sized slice cannot change which error is reported or what is emitted. An empty window keeps its nil slice so that `libjson`'s nil-vs-empty distinction survives.
- `make-sequence`: the loop, its per-element `CheckAlloc` and `x.Copy()` are untouched; the capacity is only a hint, so an over-limit sequence still fails at the same element with the same message (`allocation size 10485761 exceeds maximum (10485760)`, verified against base). Only a forward integer range is sized — a backwards or equal-endpoint range takes the old path, because the difference it would size from is a wrapped value, not a length — and the capacity is bounded so that no setting of `MaxAlloc` can turn a range that used to exhaust memory into one that fails at `make`. Non-integer arguments take the old path.
- iterator `Get` and range nil: the elements are produced by the same calls in the same order; when a nested iterator collapses, the size is a floor and `append` grows past it as before.
- `string:join`: the length pass is the existing type-check loop; it stops at the first non-string with the same message, and nothing had been written before that point in either version. `strings.Builder.String()` returns the same bytes `bytes.Buffer.String()` did.
- The reviewer's differential battery (40 programs: string/symbol-keyed and nested maps, iterator and range shapes, the JSON encoding of every empty-window coercion, integer-overflow and allocation-limit shapes, and 10 error paths) is byte-identical against `main` on stdout, stderr and exit status. It was **not** identical before the review fixes; one program diverged on eight lines.

## Measurements

Re-measured on the review head against current `main` (53a24ca), not carried over from the first push. Micro benchmarks, `-count=10 -cpu 1`, arms interleaved round by round, benchstat; `~` marks p >= 0.05. `make bench-burnin` reports the machine FIT against a ±10% ceiling.

| Benchmark | sec/op | allocs/op |
|---|---|---|
| `CopyMapWide/10,100,1000` | -4.9%, -9.7%, -13.5% | -3%, -20%, -35% |
| `SliceToString/10,100,1000` | -11.6%, -16.7%, -17.3% | -22%, -33%, -50% |
| `SliceToBytes/10,100,1000` | ~ (p=0.063), -17.8%, -19.2% | -11%, -33%, -50% |
| `MakeSequenceInt/10,100,1000` | -15.6%, -13.1%, -9.5% | -12%, -3%, -0.5% (B/op -4% to -5%) |
| `IterPathGet/10,100,1000` | -18.9%, -15.4%, -9.6% | -19%, -6%, -1% |
| `RangePathNil/10,100,1000` | -42.6%, -43.1%, -37.5% | -50%, -64%, -71% |
| `Join/10,100,1000` | -41.9%, -46.7%, -50.1% | -50%, -71%, -82% |

`benchgate` over these arms: 0 significant moves in the bad direction, 0 at or above the gate.

`CopyMapWide` was measured four separate times because the review reported a `+11.9%` regression at `/10` (n=6). It did not reproduce in any of them:

| Run | /10 | /100 | /1000 |
|---|---|---|---|
| dedicated, n=10, default benchtime | -5.1% (p=0.022) | ~ (p=0.17) | -11.9% (p=0.000) |
| dedicated, n=20, fixed 3000x iterations | ~ (p=0.063) | -10.1% (p=0.000) | -14.4% (p=0.000) |
| full 7-family sweep, n=10 | ~ (p=0.44) | ~ (p=0.15) | -7.8% (p=0.015) |
| final sweep vs 53a24ca, n=10 (the table above) | -4.9% (p=0.009) | -9.7% (p=0.000) | -13.5% (p=0.000) |

No run puts `/10` on the slow side of neutral, so the sizing is left ungated: a threshold would add a branch and a magic number to defend against an effect four measurements could not find. The allocation counts are deterministic and fall at every width (-1, -7 and -16 allocations per call).

Production phylum under substrate (regression guard; 3 interleaved rounds, `-cpu 1`, burn-in FIT at ±1%) — measured on the pre-review head, and unaffected by the review fixes, which only add branches on the empty-window and backwards-range paths that this workload does not take:

| Benchmark | base | branch | verdict |
|---|---|---|---|
| preheater cold, sec/op | 100.3ms | 89.1ms | ~ (p=0.10, n=3) |
| preheater fork, sec/op | 39.1ms | 33.4ms | ~ (p=0.10, n=3) |
| preheater cold, allocs/op | 220.6k | 220.6k | unchanged |
| preheater fork, allocs/op | 30.47k | 29.49k | ~ (p=0.10, n=3) |
| unit-test suite, 29 files x 3 rounds, sec/op | 221.9ms | 221.2ms | ~ (p=0.37, n=87) |
| unit-test suite, allocs/op | 1.631M | 1.631M | unchanged |

No regression; the phylum does not lean on these sites, which is the expected shape.

## Test Plan

Re-run on the review head, rebased onto current `main` (53a24ca), unless a row says otherwise:

- [x] `go build ./...`, `go vet ./...`
- [x] `go test ./...` (41 packages)
- [x] `go test -tags=elpscheck ./lisp/...`
- [x] `go test -race -count=1` on `lisp`, `libelpspath`, `libstring`, `libjson`
- [x] `go test -run Fuzz ./lisp/` (corpora replay)
- [x] `make check-golangci-config`; `golangci-lint run ./...` and `--build-tags elpscheck ./lisp/...` (v2.6.2, the CI pin): 0 issues
- [x] `./elps fmt -l ./...` reports nothing; `./elps doc -m` clean
- [x] `make bench-elpscheck-smoke` (every benchmark once under elpscheck, including the new ones)
- [x] `make bench-burnin` FIT; `make bench-gate-arms` over the measured arms: 0 significant moves in the bad direction
- [x] 40-program differential battery byte-identical against `main` on stdout, stderr and exit status
- [x] every commit builds and vets on its own
- [x] substrate suites on the production phylum pass under the shared and the isolated runner — **run on the pre-review head only**, not re-run after the review fixes or the rebase

Closes #596

---
_Generated by [Claude Code](https://claude.ai/code/session_018RgUjSmaNsv8hyZdDwqGNX)_

### DIFF
--- a/lisp/builtins.go
+++ b/lisp/builtins.go
@@ -1944,6 +1944,60 @@ func builtinZip(env *LEnv, args *LVal) *LVal {
 	return v
 }
 
+// maxSequencePresize bounds the capacity make-sequence reserves up front,
+// on top of the runtime's own allocation limit.
+//
+// The capacity is only ever a hint: the loop appends into it and grows past
+// it exactly as it did before any of it was sized.  Past roughly a million
+// elements the growth it saves has amortised to nothing -- append doubles,
+// so a list of any size beyond this reaches its length in a handful of
+// regrows -- which is why bounding here costs nothing measurable.
+//
+// What it buys is that the number handed to make is always derived from a
+// real element count and never from MaxAlloc, which an embedder sets and
+// which has no obligation to be a number of elements that can exist.  With
+// the limit raised (WithMaxAlloc(1<<50), say) an unbounded clamp asked for a
+// capacity beyond what any slice can address, and make rejected it: a huge
+// forward range then failed as "internal-panic: makeslice: cap out of range"
+// where it used to grow until it ran out of memory.  Different behaviour,
+// and avoidable.
+const maxSequencePresize = 1 << 20
+
+// makeSequenceCap returns the capacity to reserve for the integer sequence
+// [start, stop) by step, or 0 to reserve nothing.
+//
+// Only a forward range is sized.  The length is (stop-start) rounded up to a
+// multiple of step, and that subtraction wraps NEGATIVE -- the signal an
+// overflow is skipped on -- only when stop really is the larger of the two.
+// Run backwards it wraps POSITIVE: (make-sequence 1 MinInt64) has a
+// difference of MaxInt64, so sizing without checking the direction first
+// would reserve for a sequence that has no elements at all.  Checking
+// start < stop leaves the backwards and equal-endpoint shapes at 0, and
+// makes any genuine overflow wrap negative and be skipped.
+//
+// The result is clamped to the runtime's allocation limit, because a
+// sequence longer than that fails in the loop at the same element it always
+// did, and to maxSequencePresize, because the limit is not a number of
+// elements anyone promised can be allocated.
+func makeSequenceCap(start, stop, step, limit int) int {
+	if start >= stop {
+		return 0
+	}
+	d := stop - start
+	if d <= 0 {
+		// The true difference is positive, so this is an overflow.
+		return 0
+	}
+	n := (d-1)/step + 1
+	if n > limit {
+		n = limit
+	}
+	if n > maxSequencePresize {
+		n = maxSequencePresize
+	}
+	return n
+}
+
 func builtinMakeSequence(env *LEnv, args *LVal) *LVal {
 	start, stop := args.Cells[0], args.Cells[1]
 	if !start.IsNumeric() {
@@ -1972,6 +2026,11 @@ func builtinMakeSequence(env *LEnv, args *LVal) *LVal {
 		}
 	}
 	list := QExpr(nil)
+	if start.Type == LInt && stop.Type == LInt && step.Type == LInt {
+		if n := makeSequenceCap(start.Int, stop.Int, step.Int, env.Runtime.MaxAllocBytes()); n > 0 {
+			list.Cells = make([]*LVal, 0, n)
+		}
+	}
 	for x := start; lessNumeric(x, stop); x = addNumeric(x, step) {
 		if msg := env.Runtime.CheckAlloc(len(list.Cells) + 1); msg != "" {
 			return env.Errorf("%s", msg)
--- a/lisp/builtins.go
+++ b/lisp/builtins.go
@@ -2087,7 +2087,15 @@ func builtinSlice(env *LEnv, args *LVal) *LVal {
 		case LBytes:
 			return String(string(list.Bytes()))
 		default: // LSExpr
+			// One byte per cell, so the slice is sized once.  An empty
+			// window stays nil: libjson encodes a nil []byte as JSON null
+			// and an empty non-nil one as "", a v1.13.0 compatibility
+			// contract (encodeBytes in lisp/lisplib/libjson/encode.go), so
+			// make([]byte, 0, 0) would change what the value serializes to.
 			var b []byte
+			if n := len(list.Cells); n > 0 {
+				b = make([]byte, 0, n)
+			}
 			err := appendBytes(env, list, func(x byte) {
 				b = append(b, x)
 			})
@@ -2103,7 +2111,15 @@ func builtinSlice(env *LEnv, args *LVal) *LVal {
 		case LBytes:
 			return list
 		default: // LSExpr
+			// One byte per cell, so the slice is sized once.  An empty
+			// window stays nil: libjson encodes a nil []byte as JSON null
+			// and an empty non-nil one as "", a v1.13.0 compatibility
+			// contract (encodeBytes in lisp/lisplib/libjson/encode.go), so
+			// make([]byte, 0, 0) would change what the value serializes to.
 			var b []byte
+			if n := len(list.Cells); n > 0 {
+				b = make([]byte, 0, n)
+			}
 			err := appendBytes(env, list, func(x byte) {
 				b = append(b, x)
 			})
--- a/lisp/lisp.go
+++ b/lisp/lisp.go
@@ -753,6 +753,19 @@ func SortedMap() *LVal {
 	return SortedMapFromData(&MapData{newmap()})
 }
 
+// SortedMapSized returns an empty sorted-map whose backing is sized to
+// receive n entries, as make(map, n) sizes a Go map.  n is a capacity hint:
+// the map is empty, it grows past n as SortedMap's would, and a negative n
+// is treated as zero rather than panicking the way make(map, n) does.  Use
+// it where the entry count is known before the first Set, such as when
+// copying a map.
+func SortedMapSized(n int) *LVal {
+	if n < 0 {
+		n = 0
+	}
+	return SortedMapFromData(&MapData{newmapSized(n)})
+}
+
 // SortedMapFromData returns sorted-map with the given backing implementation.
 // Applications calling this function must make ensure the Map implementation
 // provided satisfies the semantics of Map methods.
--- a/lisp/lisp.go
+++ b/lisp/lisp.go
@@ -755,10 +755,12 @@ func SortedMap() *LVal {
 
 // SortedMapSized returns an empty sorted-map whose backing is sized to
 // receive n entries, as make(map, n) sizes a Go map.  n is a capacity hint:
-// the map is empty, it grows past n as SortedMap's would, and a negative n
-// is treated as zero rather than panicking the way make(map, n) does.  Use
-// it where the entry count is known before the first Set, such as when
-// copying a map.
+// the map is empty and it grows past n as SortedMap's would.  A negative n
+// is clamped to zero rather than passed to make, because the language
+// requires make's size argument to be non-negative and gc's tolerance of a
+// negative map hint is not something an exported constructor should oblige
+// its callers to rely on.  Use it where the entry count is known before the
+// first Set, such as when copying a map.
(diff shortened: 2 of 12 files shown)


sam-at-luther:
Second adversarial review, run against `ed8f681` with both worktrees pinned and every gate re-run independently: **clean — 0 blocking, 0 should-fix, 1 nit, 3 info.**

No program differs from `main`. 175 targeted single-form probes, 400 generated programs (4,800 forms) and a 336-line allocation matrix (14 `MaxAlloc` settings x 24 expressions) are byte-identical on stdout, error text, error order and exit status. The harness was validated by sabotage rather than trusted: reintroducing `make([]byte, 0, len(cells))` produces 12 diffs, and removing the `maxSequencePresize` clamp kills the raised-limit probe with an out-of-memory fatal error. `makeSequenceCap` was checked against an independent big.Int reference over 31,752 structured combinations plus 3,000,000 random ones, and against the real loop count for every small range.

The reported `CopyMapWide/10` regression is resolved as measurement noise, on the reviewer's own numbers: interleaved n=10 at `-cpu 1` gives **-8.14% (p=0.003)**, all 21 arms improve, none regress, and the allocation counts — which carry no noise — fall at every width.

The nit is fixed in `d577abf`, which is comment-only and changes no behaviour. The godoc for `SortedMapSized` claimed a negative `n` was clamped "rather than panicking the way `make(map, n)` does". That is false for gc: `make(map[K]V, n)` accepts every negative `n` down to `MinInt64` and treats the hint as zero — only `make([]T, 0, n)` panics. So the negative-`n` rows of `TestSortedMapSizedIsAHintOnly` pinned nothing they claimed to: deleting the clamp leaves the test green. The clamp stays, because the language requires make's size argument to be non-negative and an exported constructor should not oblige a caller who computes a size from a subtraction to depend on gc being more forgiving than the language is. The godoc and the test comment now say that outright, including that removing the clamp would not fail those rows.

The three info items are recorded and not acted on: a `strings.Builder.Grow` overflow in `string:join` needing ~18 GB of input and unreachable on 64-bit; an incomplete (not incorrect) comment at `path.go:1512` about a reservation being a floor when it can also over-estimate; and the pre-existing `(slice 'bytes '() 0 0)` -> `null` versus `(slice 'bytes "" 0 0)` -> `""` asymmetry, which is identical on both trees and is now pinned by this PR.

---

Issue #585, opened 2026-09-03
copy and detach de-alias two LVals that share one *MapData


sam-at-luther:
## Problem

`detacher` (`lisp/detach.go`, also behind the lisp `copy` builtin via `lisp/copy.go`) memoises per `*LVal` only, the same shape as #576 in `Fork`. Two headers over one `*MapData` (as `(quasiquote (unquote a))` produces) come out of `copy` as two independent maps:

```lisp
(set 'inner (sorted-map))
(set 'inner2 (quasiquote (unquote inner)))   ; second header, same *MapData
(set 'pair (copy (list inner inner2)))
(assoc! (nth pair 0) "q" 1)
(get (nth pair 1) "q")                        ; => () ; the original pair would give 1
```

`copy.go`'s doc disclaims backing-array aliasing but not `*MapData`.

## Proposed fix

Port the `*MapData` memo from `forker.mapData` (the #576 fix) to `detacher`, seeded before the entries are walked, or extend the `copy` doc to state the limitation. The same applies to `*[]byte` and `NativeCloner` payloads reached through two headers.

Found during the adversarial review of the #576 fix.

Issue #582, opened 2026-09-03
Macro-expansion stamp writes into unsealed runtime lists that are bindings


sam-at-luther:
## Problem

`stampMacroExpansion` (lisp/macro.go) writes the macro call site onto every unsealed, location-less node of an expansion. It cannot tell expansion output from storage that also belongs to a binding. The fix for the VALUE case (functions, native handles, sorted maps, vectors reached through `Get`) replaces the value with a stamped private header copy instead of writing to it. One residual case remains: an **unsealed syntax container that is itself a binding**.

```lisp
(set 'l (list 1 2))        ; a runtime list: unsealed SExpr, cells with no location
(defmacro m () l)          ; the macro body evaluates to the binding itself
(m)
```

After `(m)`, `l` and its cells carry the call site of `(m)` as their source location, for the rest of the process. Same on every release since the stamp existed (pre-dates #274). Observable only through debug metadata: error locations rendered when such a node is evaluated, step-limit or context-cancellation error text, the debugger's macro-expansion inspector, and the profiler.

## Why the value-case fix does not cover it

A list built by `(list ...)` is an `LSExpr`, the same node type the reader produces, so the stamp cannot distinguish it from a list the macro body built with quasiquote a moment ago. Sealing marks reader output; runtime lists are never sealed.

## Options

- Have `list`, `append`, `concat` and friends mark their results (an "owned by runtime" bit) so the stamp skips them the way it skips sealed nodes. Costs one flag write per constructed list.
- Stamp a deep copy of the unsealed part of every expansion instead of stamping in place. Correct by construction, but allocates per expansion node.
- Accept the residual and document it (current state; the warning comment above `stampMacroExpansion` names this issue).

Found while fixing the value case (see the fix PR that references this issue). History of the stamp's aliasing bugs: #274, #370, #406, #431, #517.

Issue #551, opened 2026-08-26
LVal.Copy shares an LBytes value's backing storage with its copy


sam-at-luther:
**Closing: not a bug. I filed this without checking whether the behaviour was intended, and it is — documented, deliberate, and pinned by a test whose stated purpose is to pin it.**

Leaving the original report below for the record, with the correction first.

## Why it is not a bug

`(*LVal).Copy`'s own doc comment states the contract:

> …with the receiver, so it is **not a tool for transferring values between Runtimes**; the in-kernel detach (`lisp/detach.go`) covers that.

`Copy` shares storage within a runtime by design. It copies the spine and repairs only where sharing would break a type's own mutation semantics. `LSortMap` gets an arm not for hermeticity but because `assoc!`/`dissoc!` on the copy would restructure the *original's* map — and even there it deliberately **shares the value pointers**, which its comment says outright.

`LBytes` and `LArray` are reference-semantics types under that design, and both are pinned:

- `TestCopyAliasesArrayBacking` — *"expected Copy to share the array's Cells backing (reference semantics); it did not"*
- `TestCopyAliasesBytesAndMapValues` — *"expected Copy to share LBytes backing; it did not"*

Both carry the same rationale: they are **controls**, so the `TestDetach*Disjoint` assertions prove a real difference rather than a property `Copy` already had. `TestDetachBytesAndMapDisjoint` is the paired assertion, and it passes — `Detach` *does* give disjoint bytes, which is the hermetic path this issue was reaching for.

So the `LSortMap`-is-handled-but-`LBytes`-is-not asymmetry that prompted this report has an explanation: the sorted-map arm protects the container's *identity*, not its contents. For bytes and arrays the backing **is** the container, and sharing it is the point.

Verified the correction the same way as the original claim: implementing the "fix" turns `TestCopyAliasesBytesAndMapValues` red, on the assertion that exists to say this behaviour is intended.

## What was right in the report

The `append!` observation stands as a fact — `builtinAppendMutate` → `appendMutateBytes` assigns *through* the `*[]byte`, so a `Copy` taken beforehand follows the original's growth, not just its writes. That is a sharper consequence of reference semantics than element aliasing, and it isn't spelled out anywhere. If anything is worth doing here it is one sentence in `Copy`'s doc comment noting that for `LBytes` the sharing extends to *replacement* of the slice header, not only to writes through it. That is a docs change, not a behaviour change, and not worth an issue on its own.

[...]

Issue #546, opened 2026-08-26
Native value contracts (elps side): complete the clone protocol, typed handle accessors, runtime affinity


sam-at-luther:
Split out of #383 as the implementable elps-side work. #383 stays open as the umbrella design note; this ticket is what a contributor can pick up and build. The substrate-side half (a static check over substrate's own builtins) is deliberately **not** in scope here — see "Out of scope" below.

## Background: why this matters

`LVal.Native` is `interface{}` (`lisp/lisp.go:329`), the one storage channel invisible to every protection layer elps has: the seal cannot cover it, the ownership checker cannot attribute it, `Detach` rejects it, and `copy` shares it by reference. Native stays exported — it is the sanctioned home for embedder handles — so the fix is not to remove it but to let natives **opt in** to the machinery.

This is not speculative. The bug class has already fired downstream. Substrate serves every transaction by forking a per-phylum template, and forks share native payloads by reference; a B+-tree handle opened at load scope therefore became every transaction's handle. Substrate's own code records the failure (`internal/substrate/shiro/libcc/ccbptree/ccbptree.go`):

> Without the declaration a load-scope open published a template with the probe sealed inside it, and every transaction forking that template failed on the tree's first use.

It is currently held closed by a *convention* — builtin authors call `shirorpc.ProbeContextCapture` — which substrate documents as a hole rather than a guarantee (`preheat/preheat.go:662`): *"a new one that forgets to is a hole in this build's verdict and in the pre-deploy checker's alike."* Items B and C below give elps the type-level tools that convention is standing in for.

Audit status as of elps `6a6be2b` / substrate `cb43cf8`: there is **no live corruption vector** — four of substrate's five native payload types are value types, and the one mutable pointer (`*bptree.BPlusTree`) cannot be built without a context read that trips the probe. This work is about closing the class by construction, not about a fire.

---

## Item A — `copy` and `Detach` must consult `NativeCloner` (do this first)

**What's wrong.** #519 introduced the clone protocol (`lisp/fork.go:66-80`):

```go
type NativeCloner interface {
    CloneNative() interface{}
}
```

`Fork` honors it (`forker.native`, `lisp/fork.go:477-489`: replacer hook → `NativeCloner` → share-by-reference). **`copy` and `Detach` do not.** A native that has explicitly declared itself cloneable is still shared by one path and rejected by another. That is an incoherence on main today, and the cheapest item here.

[...]

PR #599, opened 2026-09-04
Guard the value-graph aliasing and location-bleed class across every walker

### LINKED ISSUE #598: The alias oracle only covers Template.NewVM; copy, detach, (*LVal).Copy and the macro stamper are undriven
## Gap

`elpstest/forkcheck.go` exists to catch exactly the bug class of #576: a walker that rebuilds a value and de-aliases two headers that shared one payload. Its ALIASING oracle is the right instrument — it compares "same object" for every pair of reachable mutable payloads, not just the ones a transaction happens to exercise.

But `RunForkCheck` only ever drives `env.Fork(...)`. It never drives the other walker over the same value graph: `detacher` (`lisp/detach.go`), which backs the lisp `copy` builtin and cross-runtime detach.

So when #576 was fixed in `forker` and guarded by the harness, the identical defect stayed live in `detacher` for the whole time the guard was green. It surfaced later as #585 and was fixed in PR #595 with targeted tests, not a class guard. The instance is guarded; the class is not.

## Proposal

Make the alias/parity oracles walker-agnostic and run them over every value-copying walker, so a new walker (or a new payload kind in an existing one) fails CI rather than shipping.

Sketch:

- Extract the reachable-payload identity map the harness already builds into a helper that takes a "produce a copy of this value" function.
- Instantiate it three ways: `Fork`, the `copy` builtin, and `Detach`.
- Assert the same property for each: for every pair of reachable mutable payloads, "same object" holds in the copy exactly when it holds in the source.
- Include the payload kinds all three walkers memoise today: `*MapData`, `*[]byte`, and pointer `NativeCloner`.

A cheaper first step, if the full generalisation is too invasive: add a static drift guard that fails when `forker` memoises a payload kind that `detacher` does not, keyed on the memo field names in the two structs. That catches the next divergence at CI time without reworking the harness.

## Files

| File | Role |
|---|---|
| `elpstest/forkcheck.go` | the harness; oracles currently bound to `Fork` |
| `lisp/fork.go` | `forker.mapData`, `forker.byteSlice`, `forker.native` memos |
| `lisp/detach.go` | `detacher.maps`, `detacher.bytes`, `detacher.natives` memos (added in #595) |
| `lisp/copy.go` | the `copy` builtin, thin wrapper over `detacher` |

## Acceptance

- The alias oracle runs over `copy`/`Detach` as well as `Fork`.
- Reverting the #595 memos makes the harness fail, not just the targeted tests in `lisp/detach_alias_test.go`.
- Reverting the #576 memo still fails, as it does today.
- A payload kind memoised in one walker and not the other fails CI.

Found while reviewing PR #595 (fixes #585).

sam-at-luther:
## Summary

Five aliasing bugs in a row came out of the same class: a primitive rebuilds a
value graph and two names for one piece of storage come apart, or the rebuild
writes into storage it does not own. #576 (Fork de-aliased two headers on one
map), #579 (a libschema validator losing its credential across a fork), #585 (the identical
defect in `copy`/detach), #582/#583 (the macro stamper writing into shared
structure), and #578's review finding F1 (a call frame reading its defining
environment's LIVE location register instead of a definition-time snapshot).

The guard that existed, `elpstest/forkcheck.go`, only ever drove `Fork`. That is
#598: when #576 was fixed and guarded, the same defect stayed live in the detach
walker for the whole time the guard was green.

This makes the oracle walker-agnostic and adds the channels the old one could
not see. The product guarantee underneath all of it is TRANSACTION ISOLATION —
an embedder loads a program once into a template and runs every transaction on a
fresh fork — so the guard asserts that guarantee directly as well as the
mechanisms that serve it.

- `elpstest/fingerprint.go` — a canonical token stream in which SHARING IS PART
  OF THE ENCODING, plus `Hash()` and a `Diff()` reporting the first divergence
  with the path that reached it.
- `elpstest/aliasguard.go` — mutation probes, per-walker contracts, the walker
  registry, and a `Witness` type that carries evidence rather than a boolean.
- `elpstest/aliasguard_location.go` — the location channel, compared by
  consequence.
- `elpstest/aliasguard_isolation.go` — transaction isolation over N forks, in
  every direction a write can travel, sequentially and concurrently.
- `lisp/walkers.go` + `lisp/walkers_drift_test.go` — the memo registry in
  production code and the drift guard, in four halves.
- `elpstest/aliasguard_broken_test.go` — ten deliberately broken reference
  walkers, the cap-truncation controls, and the claims-under-test.
- `elpstest/aliasguard_fuzz_test.go` — one fuzz target looping the walker
  registry over generated graphs.

Everything the guard offers is exported from `elpstest`, so substrate and other
embedders can point it at their own values.

**Is substrate's real topology tested?** Not by this PR, and #603 is the answer
to that question rather than a gap in this one: it takes the shape an embedder
actually deploys — one preheated template, N forks, one warm `LoadCache` — and
runs it through an isolation oracle, which nothing did before. Its findings are
summarised under "what is NOT built here".

## How the fingerprint encodes sharing

An ordinal table keyed on PAYLOAD POINTER. First arrival at a payload emits
`KIND#n{contents…}` and assigns ordinal n; every later arrival at the same
pointer emits `KIND#n` alone.

- Two headers over one map emit the same ordinal twice.
- Two headers over two equal maps emit different ordinals.

So two graphs that are `equal?` and differ only in how they share fingerprint
differently — exactly the distinction #576 turned on, and one no value
comparison can make. The table also makes the walk cycle-safe and keeps a
diamond-shaped graph linear.

Canonical order everywhere: package names, symbol names, an environment's
bindings and a package's exports are sorted before they are walked; a Go map is
never ranged directly. That last property is what keeps the encoding
deterministic and is pinned by `TestFingerprintIsDeterministic` — a raw map
range introduced into the walk fails it.

Where Go is opaque, identity is substituted for content: a function is `fid` +
package name (a Go func pointer is not comparable), and a native payload is its
Go type name plus an identity ordinal.

**"Has a payload" is keyed on the PAYLOAD, not on the header's type.** It was
keyed on `v.Type == LNative`, which is wrong twice over: `LVal.Native` is shared
storage — `LBytes` keeps its `*[]byte` there, `LSortMap` its `*MapData`, `LFun`
its `*funData` — and an embedder can annotate an ordinary node with a payload of
its own. The measured case is #603's: a Reader annotating a **sealed** `LSExpr`
reaches every fork by reference, because Fork shares a sealed value outright
*before* the native policy runs, so its `NativeCloner` is never consulted — and
nothing here saw it. On a graph holding an aliased map, aliased bytes, a native
and an annotated sealed node the cross-fork census went **1 → 4** entries, and
probe sites stayed at **2 → 2**.

Two things about that fix are worth stating because both were found by measuring
rather than by reasoning, and both are the opposite of the obvious change:

- **The probe-site walk is deliberately NOT widened.** That arm decides where the
  guard can *write* a sentinel, and a payload it cannot write into is not a probe
  site whatever header carries it. Widening it makes an annotated `LSExpr` — and
  every cell view once #602 lands — enter the deliberately-opaque native arm, so
  the walk **stops** and the real probe sites in its cells are lost.
- **`isPointerPayload(v.Native)` alone floods the census: 1 → 142.** Every builtin
  is an `LFun` holding a `*funData`, and packages are shared between forks by
  design, so a census of embedder payloads became a census of the standard
  library. `kernelOwnedPayload` bounds it — `LFun`, `LError`'s `*CallStack`, and
  cell-view links — each with its reason.

The quote flag is encoded: `'(1)` and `(1)` used to fingerprint identically, so
the whole class was invisible. `LVal.meta` and `LVal.spliced` are still not
encoded and now have rows in the exclusion block saying so — neither has an
accessor — tracked in #600.

**What the encoding cannot see**, and why the mutation probes are not
decoration: identity ordinals are assigned per walk, so equal contents
fingerprint equally regardless of sharing, and two distinct `*[]byte` headers
over ONE backing array get two distinct ordinals — the encoding reports "not
shared" while the memory is. Both gaps are covered by the probe sweep, and both
are committed negative controls.

The environment walk stops at the global/root boundary and emits a sentinel;
without it every probe would fingerprint the whole standard library.

Excluded from the value channel, each with a reason in the code: `evalCtx`,
`LEnv.ID` and the `Runtime` pointer; a `*CallStack`'s identity; the seal bit and
captured environments, both behind options, because `copy` clears the seal and
shares closures by documented contract. `LEnv.loc` is not excluded — it has its
own channel. Macro-expansion metadata is behind an option too, for a reason
peculiar enough to have its own row in the contract table below.

## Why mutation probes beat pointer comparison

Pointer identity is an inference about semantics. The semantics are "does a write
through A show up at B", so the guard writes a sentinel into one reachable
payload and records WHICH SITES SEE IT. That set is the alias equivalence class,
measured rather than inferred, and a walker preserves aliasing exactly when the
class it produces is the class it was given. Every write is undone and the
fingerprint asserted back to its pre-mutation value.

It is not academic: Fork mints a fresh `*funData` per function header, which
nothing can observe, and a map implementation that copied on read would preserve
pointers while breaking the semantics. Measuring the write is blind to both.

Site kinds are sorted-map entry, bytes element, and (for walkers whose closures
are in scope) a captured binding in a closure's environment. Every fork is taken
one hop deeper as well, as defence in depth against a class a single hop cannot
exhibit — see the review-history section for why the `#579` justification this
line used to carry was false.

The sweep is O(n²) in site count, so it is capped — **loudly**, with an
`AliasCheck.MaxProbeSites` knob, the boundary pinned on both sides, and
`math.MaxInt` handled as a genuine "no cap" rather than as a trap.

### The isolation directions

The participants are a template and its n forks — n+1 environments, but only
**two roles** — so a write that leaks between environments crosses from one role
to the other or between two occupants of the fork role. That is exactly **three
single-hop directions**, and `CheckTransactions` asserts all three. The fourth
row below is not a fourth direction: it is the same hop as the row above it,
observed later.

| Direction | Property |
|---|---|
| fork → other fork | "a transaction on one fork is invisible to every other fork" (swept i×j after every transaction) |
| fork → its template | "the template is unchanged by a transaction on a fork", plus a concurrent arm that doubles as the `-race` gate |
| fork → template → later fork | "a fork taken after other forks were mutated is pristine" — the fork→template hop again, observed after the fact, **not** a composition of the rows above |
| **template → existing fork** | **"a transaction on the template is invisible to every existing fork"** |

The template → existing fork direction was missing until a reader asked which
directions were actually asserted. The asymmetry
is structural: if a fork shares a payload with its template — exactly #576/#585 —
then a write **through the template** lands in the fork. Fork→template and
template→fork are the two ends of one shared pointer, and only one end was
tested. "The embedder treats the template as immutable after load" is precisely
the kind of supposition this guard exists so nobody has to make.

It reuses the check's existing `Tx` set, so every caller gained the property
without a call-site change, and a transaction set that never moves the template
is reported rather than passing for free. Its negative control is a fork that
forks properly and then re-points one binding at the template's own value:
detected, naming which fork moved and the diverging path; removing the property
turns that control red, with `go vet` clean on the weakened tree so the red is a
real failure and not a build error.

**What it catches and what it does not**, measured rather than assumed:
reverting the #576 map memo fails "a fresh fork is indistinguishable from its
template" and the pristine-successor property, and leaves this one **green**.
That is not a gap. This direction catches OVER-sharing; #576 is DE-aliasing, and
a fork that copies too eagerly is *more* isolated from its template, so a
template write cannot reach it. Opposite ends of one axis, caught by different
properties — which is the argument for asserting every direction rather than
assuming one implies the others.

### A channel no same-vs-same comparison can watch

Macro-expansion metadata is the one channel the fingerprint comparison is
structurally unable to see, and saying why is more useful than the fix.

Fork **deliberately drops** it (`lisp/fork.go`). Property 2 asks whether a fresh
fork is indistinguishable from its template — so a token for this field in the
default fingerprint would fail on *correct* code, and leaving it out makes the
field invisible. **A field a walker is supposed to drop is invisible to a
same-vs-same comparison by construction.** That is why #600's gap 2 had no
coverage at all: deleting `cp.macroExpansion = nil` left `go test ./...` entirely
green.

It needs a direct property, and it measures the harm rather than the contract
line:

> no macro-expansion metadata on a fresh fork reaches a template value

`macroExpansionContext.Args` holds the unevaluated call-site arguments as
`[]*LVal` — the ORIGINAL nodes — and `(*LVal).MacroExpansion` hands them out, so
a fork that kept its template's metadata hands a consumer pointers into the
template's tree. The witness names both ends:

    user:expansion/1/0 carries expansion metadata for lisp:defun
      whose arg 1 is the template's user:expansion/1/2/0

**Sealed args are not a leak, and args are not always sealed** — measured, not
assumed. Sealed parse-tree nodes are immutable and shared with every fork
outright by the same kernel policy, so a pointer to one conveys nothing: a quoted
literal call records 12 args, all 12 sealed. A call form built at RUNTIME records
15, of which **5 are not sealed** — mutable, template-owned nodes. The property
is about those.

Two facts fell out of building the control, both worth knowing:

- **A fork inherits no `Debugger` either** (`lisp/fork.go`: "Profiler and
  Debugger do not travel"), so a fork cannot mint metadata unless an embedder
  attaches one to the fork. That is the only configuration in which a fork may
  legitimately hold any, and it is what the positive control models. Without it
  the property would be satisfied by a walker that strips metadata everywhere,
  including from where it belongs.
- **Metadata is only reachable in narrow shapes.** It needs a Debugger ATTACHED
  (dormant is enough — the kernel gates on `Runtime.Debugger != nil`, not on
  `IsEnabled`), a BUILTIN macro (only synthetic nodes are stamped), and
  `macroexpand-1` to RETAIN the stamped tree. An ordinary macro call evaluates
  its expansion and discards it. Both anti-vacuity conditions are asserted, so
  the test fails loudly if any of that changes.

## Why locations are compared by consequence

A fork deliberately starts with an EMPTY location register (#440), so a template
and its fork legitimately hold different locations; comparing registers directly
would report the drop Fork exists to do. The register is observable through one
surface — error rendering — because a step limit, a nesting limit and a context
cancellation all raise through `ErrorConditionf`, which stamps `env.loc` into the
rendered text, into `Source()` and into the call-stack frames.

So the guard stamps a unique sentinel location onto one reachable environment and
requires the rendering of an error raised elsewhere to be byte-identical. The
forcing device is a countdown context rather than a step budget, because a
countdown carries no accumulated runtime state and replays exactly.

Three properties: no cross-environment bleed; a fork carries no evaluator
location; and definition-site snapshots stay frozen (#578 F1), asserted up front
so a probe that stopped tripping at a body entry fails loudly instead of passing
by reaching nothing. A fourth applies the same across forks.

## The four walker contracts

| Walker | Closures | Backing | macroExpansion | Memoises | What the oracle asserts |
|---|---|---|---|---|---|
| `Fork` | in scope | see below | **drops** (`fork.go:475`) | `*MapData`, `*[]byte`, native ptr, `*LVal`, `*LEnv` | full isolation, captured bindings are probe sites, two hops |
| `copy` | refused (shared) | rebuilt | **copies** — #604 changes this | `*MapData`, `*[]byte`, native ptr, `*LVal` | same values and sharing; closures are shared, so what they captured is out of scope |
| `Detach` | refused (rejected) | rebuilt | **drops** (`detach.go:178`) | same as `copy` | the refusal still happens, with unchanged error text |
| macro stamp | refused | preserved | writes it | `*LVal` | the expansion mutates nothing reachable outside its own output |

Adding a fifth walker is one row in `elpstest.Walkers()`, and it gets fuzz
coverage automatically because the fuzz target loops the registry.

**The macroExpansion column is three different behaviours, encoded rather than
averaged** — `TestMacroExpansionBehaviourPerWalker` pins each. `FingerprintOptions`
gains an opt-in `MacroExpansion` token set by **no** walker's defaults, precisely
because averaging them would make it wrong for two of the three; it records
argument *identities* rather than presence, because the harm this channel carries
is a pointer.

**Copy: copies** — measured on this tree by
`TestMacroExpansionBehaviourPerWalker`, and its recorded args can point at
MUTABLE source nodes (3 carried across on the fixture, 2 sealed and **1 not**).
That is the same shape as the Fork leak above, and it matters most on the
`TextLoader` path, whose whole purpose is to hand each evaluation a private tree.

**#604 (stacked) changes Copy to DROP the metadata**, as Fork and detach do, with
its own control (`TestCopyDropsMacroExpansionMetadata`, carrying the same
anti-vacuity condition: the source must hold metadata whose recorded args include
an unsealed node). Remap was considered and rejected — it needs a per-walk
context memo, which is a new memo-shaped field the drift guard above would then
require registering, plus a post-pass, for debugger-only data.

This PR's assertion stays the **measurement on this tree**, deliberately, so
#604's flip from "copies" to "drops" is a visible diff in the test rather than a
silent agreement.

The macro stamper's contract deserves one sentence, because the brief asked for a
different one. It was asked to guarantee "its output must not alias unsealed
source structure"; the code does the opposite, deliberately — a stamped copy
SHARES everything behind a pointer and privatises only its own struct fields. The
guarantee that IS load-bearing, and the one this PR asserts, is the ownership
rule: **the stamp writes only to nodes it allocated**, which is what revert-proof
row 5 exercises.

Two documented exceptions are encoded as contract rather than as failures: a
shared closure keeps reading the containers its defining scope holds, and the
Cells backing array — which needs its own paragraph, because its status is about
to change.

**The Cells backing array.** `copy` and detach do not preserve slot aliasing and
are not being changed. **Fork does not preserve it either today, and that
contradicts `docs/fork.md:118`**, which promises for mutable data under Fork
*"hermetically copied, aliasing and cycles preserved"*. Observable from pure
ELPS, with no Go internals:

```
(set 'l (list 30 10 20))
(set 'tail (cdr l))      ; cdr reslices, sharing the elements
(stable-sort < l)        ; sorts in place, as a transaction would
tail
```

    cold environment / template  ->  '(20 30)
    fork                         ->  '(10 20)

`lisp/fork.go` allocates a fresh backing array per header; element pointers are
memoised, so the elements stay shared, but the array SLOTS do not. **The
direction is Fork copying TOO MUCH** — isolation is intact, over-copying cannot
leak a transaction anywhere. It is a FIDELITY gap, the same shape as #576 and
#585 but for cells rather than maps and byte buffers.

**This stops being an exception and becomes an asserted contract once #602 sits
underneath this PR.** #602 makes a view record its root, so Fork preserves slot
aliasing; the guard then walks the link as a REFERENCE and can assert *"a fork's
view shares its slots with the fork's own root exactly as the template's view
shared them with the template's root"*. The specification for that commit —
which two function bodies change, the exact semantics of a stale link, and the
property above — is written out at `isCellViewLink` in `elpstest/fingerprint.go`
(`eea2084`), so the #602 commit is a fill-in rather than a design. Until then the
both-trees predicate here is **inert by construction**: no constructor in `lisp/`
writes a `*lisp.LVal` into `Native` yet, so it is never true, and this PR cannot
call #602's API because it is #602's base.

## The drift guard, four halves

- REGISTRY: every walker that rebuilds payload storage must memoise the same
  payload kinds. #585 exactly.
- SOURCE SCAN: every memo-shaped struct field in `lisp/` belongs to a registered
  walker or to the shrink-only exemption list, and every field the registry names
  must still exist.
- PAYLOAD SCAN: every `case *T:` arm of the two walkers' `v.Native` type switches
  is a registered memo kind or an exempted one.
- STRUCT-COPY SCAN: every `*x = *y` and `x := *y` on an `LVal` in `lisp/` is
  either inside a registered walker or on a shrink-only allowlist with a reason.

The three original halves all start from a walker and check what it declares. The
fourth starts from the SOURCE and asks where a payload pointer can cross from one
header to another at all — `*cp = *v` carries every pointer field and rebuilds
nothing. Measured before deciding: **18 sites**, 7 inside registered walkers, 11
not. Type-checked via `go/packages`, so `*cp = *v` on an `*LEnv` or a
`*token.Location` is not confused with the same spelling on an `*LVal`, and keyed
by enclosing FUNCTION plus a site count rather than by file — `lisp.go` holds the
registered `(*LVal).Copy` alongside four unregistered copies, so a file-level rule
would exempt all of them.

Four of the eleven are not hazards that happen to be safe: `Quote`, `Splice`,
`shallowUnquote` and `FunRef` are where the **intentional** aliasing lives.
`(quasiquote (unquote v))` *is* `Quote`, and the two-names-one-payload graph it
produces is what every registered walker exists to PRESERVE — a walker that
de-aliased them would be the defect. It found something on its first run: the
fixture in the pin added by the commit before it, which is the scan working.

**`(*LVal).Copy` is now registered, and reported as defective.** It is a fifth
value-rebuilding walker that both original halves missed — the field scan looks
for memo-shaped struct fields and `Copy` is a memo-less recursive method, and the
payload scan reads only `fork.go` and `detach.go`. Measured, it **rebuilds**
`*MapData` storage per header without memoising it (the #576/#585 defect shape,
live, in a walker neither fix reached) and **shares** an `LBytes` `*[]byte`
(#551, live). Registering it with `Rebuilds: false` would have been the
comfortable move — true of the bytes payload, and beside the point. A
registration describes the walker, not the subset of its behaviour that keeps the
guard green. So it registers with `Rebuilds: true` and no memos, and a new
shrink-only `knownDefectiveWalkers` allowlist carries both defects with the test
that measures each. Those tests assert the defects **still exist**, so the day
either is fixed its pin goes red and the row must be deleted — a row cannot
outlive what it records. **The fix is #604, stacked behind this PR.**

The exemption lists are shrink-only, every row carries its reason, and a dead row
fails the guard.

## Permanent negative controls: the guard on the guard

The revert-proof exercise below is a one-time check. It proves the guard works
today and guards nothing tomorrow, so the negative controls are committed —
entirely inside the harness, never in `Walkers()`:

| Control | Models | Detected by |
|---|---|---|
| copier with no `*MapData` memo | #576 / #585 | probe sites disagree at `map entry "k"` |
| copier with no `*[]byte` memo | #576, second payload kind | probe sites disagree at `bytes[0]` |
| copier with no native memo | #576, third payload kind | the fingerprint's identity ordinal (no probe can look inside) |
| fork that carries the template's location | #440 / #578 F1 | "a fork starts with an empty evaluator location register" |
| fork that hands every fork one stateful native | a `CloneNative` that clones once | "declares NativeCloner and is shared anyway" |
| stamper that writes into its source | #582 | "expansion mutates nothing reachable outside its own output" |
| copier that shares the source's BUFFER | a walker passing an opaque payload straight through | the mutation-probe sweep at `bytes[0]` — the fingerprints agree, so only a write can see it |
| copier that shares only the DUPLICATE buffers, past the probe cap | a leak the truncated sweep cannot reach | a partial-coverage witness naming `MaxProbeSites`; raising the cap then finds it at `bytes[0]` |
| copier that INTERNS equal-content buffers onto one array | over-aliasing a plausible copy-path optimisation would introduce | the alias-class comparison — the fingerprint cannot see it, and this is its only end-to-end coverage |
| fork that shares a payload with its template | #576/#585 seen from the other end of the pointer | "a transaction on the template is invisible to every existing fork" |
| **census keyed back on `v.Type == LNative`** | **#600 gap 1, both shapes** | **the `*[]byte` and `*MapData` go unseen, and the sealed-node annotation is "invisible to the cross-fork census"** |
| **a census with no `kernelOwnedPayload` bound** | **a key that reports everything** | **"141 entries for a graph with four payloads in it"** |
| **the cell-view exclusion removed** | **treating #602's root link as a payload** | **a plain two-element list gains `annot#1(*lisp.LVal)` and every later ordinal shifts** |
| **a walker that strips macro-expansion metadata everywhere** | **a fix that satisfies the property by deleting the feature** | **the fork's OWN expansion metadata must not be reported** |

A positive control keeps them honest: the same reference copier with every memo
present must PASS, and a correct fork over the template-sharing program reports
nothing, so each control demonstrates direction rather than noise. The census
carries its own negative half — an `LString` must reach it under no key — so a
census that simply reported everything would fail.

## Revert proof

Each row reverts one real historical fix in production code and shows which
property goes red. **This is a committed gate, not a one-time exercise**: `make
mutation-proof` (`scripts/mutation-proof.sh`) runs all nine on every PR and
requires each to be caught by a needle measured for uniqueness and stability.

| # | Reverted | Result |
|---|---|---|
| 1 | `*MapData` memo in `forker` (#576) | RED — `TestAliasGuardOverEveryWalker`, `TestForkCheck_SortedMapAliasAcrossHeaders`, `TestForkFingerprintsIdenticallyToItsTemplate`, and `TestTransactionIsolation` at "a fresh fork is indistinguishable from its template" and "pristine successor" |
| 2 | the three memos in `detacher` (#595) | RED on `copy` AND `Detach`, not on `Fork`. Witnesses name `map entry "self"`, `bytes[0]` and `native#…` |
| 3 | delete ONE memo from `detacher` only | RED in the STATIC drift guard — no fuzzing required |
| 4 | #578 F1 | RED on the definition-site property (`want program.lisp:3:23, got 3:14`) AND on the bleed sweep |
| 5 | the macro stamp writes in place (#582) | RED: witness `user:a` acquiring `at(stamp-call.lisp:1:1)` |
| 6 | nothing (control) | GREEN |
| 7 | `forker.pkg` shares `funNames` | RED: `want funname(_fun#,peek), got funname(_fun#,peek-alias)` |
| 8 | `6ef3da5` (#579) | RED on `TestForkCheck_SchemaValidatorCredential` — see below |
| 9 | **`cp.macroExpansion = nil` (#600 gap 2)** | **RED on "no macro-expansion metadata on a fresh fork reaches a template value", with two witnesses naming the leaked paths** |

Three of the attributions did not land where they were expected, and the honest
result is recorded rather than engineered around:

- Reverting #576 fails TRANSACTION ISOLATION, but at "a fresh fork is
  indistinguishable from its template" and "pristine successor", NOT at FORK
  INDEPENDENCE — #576 is an intra-fork de-aliasing, so a property about what one
  fork's transaction does to another cannot see it.
- Reintroducing #578 F1 fails the definition-site property and the
  cross-ENVIRONMENT bleed sweep, not the cross-FORK one: F1 leaks within one
  environment tree, never between two.
- Reverting #576 also leaves the new template→fork direction green, for the
  reason given in that section: it catches over-sharing, and #576 is
  de-aliasing.

**Row 9's needle rides as a MUST-NOT on all eight other rows**, so one gate run
measures both halves: the row trips by its own needle, and no other mutation
emits it. That is how uniqueness is established here — by CI, in one run, rather
than by driving nine mutations by hand. Its own must-NOT is property 1, because
deleting that line changes no value the fingerprint encodes, which is the gap
restated.

**Eight of the nine rows pin a property string this PR adds. The 579 row does
not, and that is recorded rather than glossed.** Reverting `6ef3da5` emits **no
property string at all**: measured 5/5, it reddens exactly
`TestForkCheck_SchemaValidatorCredential` and nothing else. That test comes from
`477ea95` and is **not in this PR's diff**. So the guard here is blind to #579 —
defensibly, since #579 is a capability revoked by HEADER IDENTITY rather than a
leak, so alias/isolation channels are the wrong instrument and cold-vs-fork
parity is the right one.

**That caveat is closed by #601, which is stacked on top of this PR**, not by
anything here. #601 folds cold-vs-fork parity into this same harness, so the
oracle that always could see #579 emits a property string like every other
channel — *"a transaction on a fork raises exactly when it raises on a cold load
of the same program"* — measured unique across all the mutations. **The property
row for #579 lands there, at the top of the chain, where the property exists.**
It was briefly committed here and reverted: this branch is #601's base, so the
string does not exist in this tree until #601 merges into it, and the gate would
have been red by construction on a PR asking to be merged. The manifest, rule 3,
the Makefile and the gate's success line all state the current position rather
than the intended one.

## What is NOT built here, because it already exists

- **Load-cache sealing.** Enforced rather than asserted, covered by five existing
  test files and two fuzz targets, with the design in `docs/sealed-ast.md`.
- **Runtime affinity.** `lisp.RuntimeBound` (#546) already checks declared
  payloads at use, fork and detach time. The guard adds the two things affinity
  checking cannot do: pointer-identity sharing across forks, which needs no
  opt-in, and `TestEveryReachableNativeDeclaresItsSharing`.
- **The embedder's real `LoadCache` topology.** That is #603's, not this PR's —
  one preheated template, N forks, one warm cache, run through an isolation
  oracle. One of its findings belongs here as a documented contract, because this
  is where `LoadCache` sharing is explained: **concurrent first-touch is not
  single-flight.** `FuzzLoadCacheMultiEnv`'s `stores == 1` invariant holds
  SEQUENTIALLY; two environments that miss the same key at the same moment both
  parse and both store. That is not a correctness bug — the entries are equal and
  sealing makes them immutable — but it is a duplicated parse under concurrency,
  and anyone reading `stores == 1` as a single-flight guarantee would be wrong.

## Fuzz budget

`scripts/fuzz.sh --list` reports 41 targets with this one added; `make
fuzz-budget-check` passes with **eleven shards** — the tenth had room for zero
more, so this PR adds the eleventh rather than leaving the next target to
discover the cliff. It is ONE target looping the walker registry internally, so a
fifth walker is fuzzed the day it registers.

**The generator cannot reach every shape, and that is written down rather than
assumed.** Neither the probe cap nor the environment cap is reachable by fuzzing,
so the controls for both are deterministic and committed —
`TestTheFuzzGeneratorCannotReachEitherCap` drives the real generator and fails if
that stops being true.

**The generator's static predictions are now pinned to real evaluation.**
`generateAliasGraph` decides, without evaluating anything, what each binding will
hold; a wrong prediction emits an operation the value does not accept, the
program raises, and the fuzz target reads the raise as "a program the walker
legitimately cannot process" and SKIPS it. Silent lost coverage. It mispredicted
a quasiquote of an already-quoted binding — `lisp.Quote` copies an unquoted value
and keeps its type, but WRAPS an already-quoted one in an `LQuote` — and `(list
…)` returns a quoted list, which the generator also had wrong. Measured effect
over the 8 committed seeds plus 2000 generated inputs: **53 of 2007 skipped
before, 38 after**, walker-layer skips 2 → 0. (None of the 8 committed seeds
skipped in either arm, so this cost coverage on generated inputs, not on the
corpus.) `TestTheGeneratorsPredictionsMatchTheRuntime` drives 600 scripts, loads
every template and transaction, and compares every recorded kind and quote flag
against what the value turns out to be — it goes red with 31 type mismatches on
the pre-fix generator, and refuses to pass vacuously.

**The fuzzer found something on its first run, and it was the oracle, not elps.**
A generated graph of closures made `copy` look like it de-aliased, because the
fingerprint walked into a SHARED closure's captured environment — `copy`'s
documented behaviour. Fixed by `FingerprintOptions.SkipCapturedEnvironments`.

## Review history

Nine adversarial review rounds, including a coverage gap a reader found by asking
which isolation directions were actually asserted. Each round re-ran the revert
proofs and reproduced every row; each also found a **false sentence in
newly-added prose**, and in most cases that sentence was load-bearing for a
decision. Full detail is in the commit messages; the short version:

| Round | What it falsified | Consequence |
|---|---|---|
| 1 | the guard itself was weakenable without any test noticing | a silent location-sweep cliff at 24 environments; the drift guard switchable off in one edit; the contract table advertising a rule the code does not enforce; **the headline mutation-probe sweep had no negative control at all** |
| 2 | *"the fingerprint covers the whole graph"* — my justification for leaving `maxProbeSites` alone | a 104-site graph with the defect past the cap returned **zero witnesses** for a live isolation leak. Also: `sameIndexSet` was **not** redundant — it is the only coverage of over-aliasing at the backing-array level, and my comment told the next maintainer it was safe to delete |
| 3 | *"every caller asks in order to STOP"* — my justification for the new truncated flag | the probe fix **reintroduced the off-by-one the same commit fixed next door**: six ordinary shapes, including a bare 96-entry sorted map, failed the guard on correct code |
| 4 | that an "unlimited" cap was safe | `MaxProbeSites: math.MaxInt` **disarmed the sweep entirely** — the walk collected zero sites and reported no truncation, so a leaking walker came back green |
| — | that three isolation directions were all of them | the **template → existing fork** direction was unasserted; see the isolation section above |
| 6 | that the new direction swept every LIVE fork | the pristine-successor fork was **not** in the set property 5 swept, against a Baseline already promising "every live fork" — a walker over-sharing only there produced **zero witnesses**. Also: a stale count of the properties was still live in the **exported** API doc, past both the drift guard's scan window and its casing |
| 7 | that the mutation gate was safe to require | it was **flaky at ~1 run in 6**, on a required PR check — worse than no gate, because it trains maintainers to re-run then delete it. The flaky row and a **mislabelled** row were the same row: the "#579" mutation actually reverted a second #576-class protection, under the wrong issue number |
| 8 | that a claim fixed in one place stays fixed | **three provenance mislabellings in one chain**, and the fix for one of them introduced a fourth — see below |
| 9 | that a false *reason* is harmless when the conclusion is right | the `*CallStack` exemption's stated reason was **false** and measurably so; and three fixes that looked obvious were each wrong in a way only measurement showed — see below |

### Round 9: three corrections that measurement made, and reasoning would not

- **A false reason under a true conclusion.** The `*CallStack` exemption said "no
  constructor can alias one across two headers: `SetCallStack` is called once".
  `(*LVal).Copy` is `*cp = *v`, which carries `Native`, so it aliases one across
  two headers — `TestCopyAliasesCallStackAcrossHeaders` measures it. The
  conclusion survives for a better reason; the stated one did not, and a reader
  auditing the row would have been misled by it.
- **The obvious fix for #600 gap 1 was wrong in two places.** Replacing the
  predicate "in all three walks", as the issue proposed, would have stopped the
  probe walk at annotated nodes (losing coverage while appearing to extend it)
  and flooded the census from 1 to 142 entries with the standard library. Both
  found by measuring the before and after, not by reading the code.
- **A registration that describes only the convenient half is a lie.**
  `(*LVal).Copy` with `Rebuilds: false` would have been true of its bytes
  behaviour and silent about the map defect that the registry check exists to
  report.

### The best evidence in this PR is what happened to its own claims

A guard against claims that outlive their truth is only as good as its own
prose. Round 8 traced one such claim commit-to-commit, and it is the sharpest
illustration available of why this PR exists:

| where | what it said |
|---|---|
| `6ef3da5` `libschema/fork_test.go` | "a fix that only survived one hop **would fail** here" — conditional, guarding against an inadequate *candidate* fix |
| `477ea95` `elpstest/forkcheck.go:39` | "a fix that survived one hop and not two **has happened** (issue #579)" |

A precaution became a historical claim, and the issue number came along from
the file it was read in. This PR then copied that sentence into
`aliasguard.go`. **No such bug exists**: every commit body in the repo
containing "hop" yields six distinct sentences and none describes one, and
`#579` measurably fails at the **first** fork — reverse-applying its fix
`6ef3da5` reddens `TestForkPreservesValidatorCredential`, the one-hop test.

The comments now state the real justification (defence in depth against a class
a single hop cannot exhibit) and cite the one genuine instance **by commit and
test name, in the past tense**: `d26953a` records that on a shared libtesting
suite the fork-of-fork arm was *once* the only arm that noticed
(`TestForkCheck_TestingSuitePerFork`). Not by issue number — the issue numbers
in this area have a 0-for-2 record.

**Three related mislabellings, one pattern.** The mutation manifest (round 7),
the 579 row's catcher (round 8), and this two-hop rationale — a claim corrected
in one place and left standing everywhere it was copied to. And the first fix
for the third one *restated* the archaeology in both files, creating a fresh
instance of the class inside the commit removing an instance of the class; the
justification is now canonical in one place and referenced from the other. The
first proposed citation was also conditional-mood ("would have passed"), which
would have elevated a second hypothetical into history — caught in review.

The direction matrix was corrected the same way earlier: it claimed four
directions and "the complete set for two participants", both false. This
document carried that stale copy for three commits after the code was fixed.

Every one is now a committed negative control, verified red under its weakening
and green when reverted. Round 3 introduced the rule the guard now follows, and
round 4 tightened it:

> **A comment asserting a property of the code is either backed by a test that
> fails when the assertion is false, or it does not go in** — and the named test
> must be able to *falsify the named sentence*, not merely be consistent with it.

Claims that failed that bar were removed rather than kept: the location cost
figures I could not reproduce in round 3 (they reproduce fine once the
measurement harness was fixed, and are restored with the machine named), and the
determinism test's description, which now records exactly what it does and does
not demonstrate.

Other fixes across the rounds, each with a control: `DefaultMaxEnvironments`
24 → 128 and `DefaultMaxProbeSites` 96 → 256 (a cap an ordinary program trips is
a failure that is not a bug); the location `truncated` flag's own off-by-one;
`ReachableEnvironmentsN` and its clamps at both ends of the range;
`WalkerMemos()` deep-copying as its doc promised; the `internal/walkraw` slots
made write-once; `quoteKey` no longer double-quoting.

## Test Plan

**CI is the verification, and that is a deliberate change of practice rather
than a shortcut.** Three concurrent local suites took the shared build machine to
load 32 on 4 cores and starved every agent on it for an hour; one of my own arms
returned a failure that was purely load (2198s and RC=1 for `go test ./...`,
which is 39s and green on a quiet box). A suite result measured under that
contention is not evidence. So the standing rule is now: locally only `go build`,
`go vet`, `gofmt` on touched files, and single targeted `go test -run` runs;
everything else runs on the PR's CI, which already covers every arm — Lint &
Test including `elpsvet` and the mutation gate, `-race`, `elpscheck`, eleven fuzz
shards, and the benchmark gate.

Every commit in this PR is pushed individually rather than batched, so a red arm
narrows to one commit.

What each change was verified with locally before pushing, stated exactly:

- `go build ./...`, `go vet ./elpstest/ ./lisp/`, `gofmt -l` on the touched files,
  `golangci-lint run ./elpstest/...` (0 issues, v2.6.2 — CI's pin)
- one targeted `-run` per new test
- for the new revert-proof row: `git apply --check` on the patch and `go build`
  on the mutated tree — rules 1 and 2 of the harness — then revert. Rule 3, that
  the row is caught by its own needle and by no other row's, is CI's to report,
  which is what the must-NOT column is for.

Earlier heads were verified by a full local sweep against a committed, frozen
tree: CI green on runs 1206 (`cf3cd19`), 1207 (`7bde0a5`), 1208 (`ac5a2e9`), 1209
(`5cefce4`), 1210 (`74e4ac8`), and a fourteen-arm local sweep at `75a2478` with
every arm exit 0.

**The mutation gate's stability, stated as a bound rather than as
determinism.** 55 consecutive end-to-end runs of `scripts/mutation-proof.sh`
at the same tree, 0 failures — 40 from one process and 15 from a second,
independent one, because a single sample is what missed the flakiness the
first time. What that licenses:

- It does **not** prove the gate cannot flake. By the rule of three the 95%
  upper bound on the failure rate is about 3/55 ≈ **5%**.
- It **does** exclude the regime being fixed. The pre-fix gate failed about 1
  run in 6 (2 failures in 12 measured runs, ~17%); at that rate the chance of
  55 consecutive passes is 0.83^55 ≈ **3e-5**. So the old behaviour is ruled
  out at overwhelming odds; a residual few-percent rate is not.

That distinction is the whole lesson of round 7 and is stated here rather than
rounded up.

**One line of this PR reaches outside its own additions.**
`elpstest/forkcheck.go:39` is on `main`, from `477ea95`, and is otherwise
untouched here — but it carried the false `#579` citation traced above.
Leaving a known-false claim standing while shipping a PR whose argument is that
false claims survive review would be indefensible, so it is corrected here and
called out rather than deferred.

Process corrections, since the Test Plan is the thing a reviewer trusts and I got
it wrong more than once:

- An early push claimed both golangci-lint invocations were clean when I had not
  completed a lint run against that tree — the run was starved by unrelated
  machine load, I killed it, and reused the previous wording. CI caught exactly
  what I had asserted without checking. Every sweep since ran to completion
  against a **committed, frozen** tree before this list was updated, and that
  discipline caught two lint issues (`predeclared`, twice) before they reached CI.
- A `gofmt -w lisp/` swept five unrelated files into the diff, twice. Reverted
  both times; the delta is guard and guard-test files only. A whole-repo `gofmt`
  arm also reported 13 files as a finding when all 13 are unformatted on `main`
  too and none is in this diff — the arm was scoped wrong, and is now scoped to
  the branch's own files.
- One sweep was run while the tree was still being edited, and its failures were
  my own interference — hence the freeze-then-sweep rule. It happened a second
  time while measuring mutations: four arms were re-run clean afterwards, and
  mutation work moved to an isolated worktree so it cannot recur.
- A weakening once "passed" only because the weakened tree did not compile. Every
  red control is now confirmed with `go vet` clean on the weakened tree, and
  `make mutation-proof` enforces the same rule mechanically.
- A measurement harness reported **more successes than it ran**: a reviewer's
  determinism loop printed "20 pass" for 15 runs, because its own `uniq -c`
  recap lines matched the same grep. Caught by recounting with an anchored
  pattern before the figure was used. A harness that miscounts its successes
  *upward* is the same shape as a needle that passes for the wrong reason.
- A needle measured **40/40 driving its mutation in isolation** still failed
  once in 35 end-to-end runs. Isolated measurement cannot certify a needle —
  witness output varies with test ordering and parallelism under load, and only
  the full script reproduces that. This is the reasoning error that produced the
  flaky gate, and it is now recorded in the manifest notes.
- Interrupting the gate sweep mid-run left the selftest's deliberate injection
  in the tree, because its cleanup trap never fired; the next `go test` failed
  with "simulated unrelated failure", which reads like a real regression.
  Restored, and the following sweep confirmed the trap fires on a clean exit.
- A test written to prove the quote flag was encoded was **vacuous**: it compared
  `QExpr` against `Quote(QExpr(...))`, and `QExpr` already sets the flag, so the
  fingerprints differed whether or not the emit existed — deleting the emit left
  it green. Caught by falsifying it before trusting it, and rebuilt on a subject
  where the flag is the only difference.
- A revert-proof row was committed to the **wrong tree**. Its property string is
  produced by #601, which is stacked *on top of* this branch, so the gate here
  would have been red by construction until the whole chain merged. Reverted, and
  the row moves to the top of the chain. Rule 3 rejecting it was the rule working.

## Suite timing

`go test ./elpstest/`, best of five, both arms measured back to back on the same
machine (4 cores) — `origin/main` at a32ada7 in one worktree against this branch
in another:

- before: 0.385s
- after: 1.120s

+0.74s as measured at `fdf5774`; later rounds add further controls (the probe-cap
and unlimited-cap tests sweep 104-site graphs, which is quadratic work), and the
fourth drift-guard half type-checks package `lisp` through `go/packages`, which
costs about 1.3s in `./lisp/`. The cost is the guard's deterministic tests, which
build a fully loaded standard-library environment per walker per case, plus the
fuzz seed corpus `go test` replays on every run. The package still runs in about
a second and its cases are `t.Parallel()`.

The mutation gate adds 39s to CI (34s for eight mutations, 5s for the harness
selftest, measured on ubuntu-24.04-arm) — expect roughly an eighth more with the
ninth row; CI reports its own timing per run. That is why it is on the PR gate
rather than a nightly.

Closes #598


sam-at-luther:

VM construction now has independent cold/template parity, alias/isolation and location controls. That does not implement this PR's broader all-walker guard for Lisp copy, detach, Go LVal.Copy and macro stamping, or its registry/drift infrastructure. [#598](https://github.com/luthersystems/elps/issues/598) and the remaining portions of [#600](https://github.com/luthersystems/elps/issues/600) still need that distinct coverage. Adapt the remaining guard to the released APIs; the old Fork implementation and its private cell-view representation should not be restored.

sam-at-luther:
Closing in favour of #642, which carries the parts of this PR that still apply on main after #623.

Subsumed by the template work: the fourth isolation direction (template to existing fork) is structurally moot, since a template is an immutable construction plan (`lisp/template.go`, "Afterwards its mutable values may change without changing the plan"); the macro-expansion channel is closed for VMs (`template_plan.go` nils it, `template.go` rejects it on sealed nodes); the eleventh fuzz shard already landed; and 5 of the 9 mutation patches target the deleted `lisp/fork.go`. `aa0dbe4`'s principle, keying "has a payload" on the payload not the header type, is what `templateInventory.val` does in production.

Carried into #642 as four commits: the walker memo registry and drift guard (`b3843a9`+`c278340`, with rows for `templateInventory`, `templateCompiler` and the new `copier`), the LVal struct-copy scan (`ad0886b`, which on main immediately named three template-compiler sites, now audited), the Copy-path sort benchmarks (`a6f32bb`), and the `forkcheck.go` doc fix removing the false #579 citation (`7acdba6`+`74e4ac8`).

Not ported: `fingerprint.go`, `aliasguard*.go` and the mutation-proof gate. About 60% of the diff is wired to `env.Fork`; porting it to `NewTemplate`/`NewVM` means re-authoring the isolation directions and the mutation patches against `template*.go`. That remainder is tracked on #598, re-scoped.

---

PR #601, opened 2026-09-04
Fuzz the fork parity property: n forks against n cold loads (FuzzForkParity)

sam-at-luther:
Stacked on #602 (base: `claude/elps-perf-optimization-jcgfd2-fork-cell-views`, at 4255801), which is stacked on #599. Draft. Five commits: b4f65c6 (oracle + fuzz target), ba3bc39 (schedule pins), ef0831e (load/fork failures are witnesses), 39c0e48 (parity folded into the alias guard) — replayed onto #602 as 3e27015, c22bda1, 7ea6210, f3538f1 — and c684e7b (the pinned failure closed; see "The known gap is closed" below).

## What parity is

In the owner's words:

> For two VMs that get instantiated from the same source code, ELPS code should work identically to two forks from the same template. Fork was meant to be a performance optimisation to speed that up. All this machinery is to guarantee and check when that's not the case.

As a check: for a program P and per-environment transaction sequences T_1..T_n, running T_i on fork i of one template(P) must give byte-identical per-transaction results and byte-identical post-run reachable state to running T_i on cold environment i that loaded P itself. Isolation and fidelity are consequences of that, not separate goals.

## What existed and why it was not enough

- `RunForkCheck` (elpstest/forkcheck.go) has a cold arm, but only over hand-written programs, one transaction per fork. Not fuzzed.
- `FuzzAliasGuard` generates programs with controlled aliasing and forks them, but compares structure (fingerprints), never execution results. Its header says "a random per-fork transaction sequence"; measured, `generateAliasGraph` emits one transaction per fork (its own repro says so) and `CheckTransactions` runs `Tx[i]` on fork i once.
- `FuzzLoadCacheMultiEnv` / `FuzzSharedProgramMultiEnv` are differential over n environments, but the n are independent loads, never forks.
- The alias guard's headline oracle (`CheckTransactions`) knew nothing of parity at all, which is why #599 carries "the new guard is blind to #579" and the mutation-proof 579 row pins a test name rather than a property.

## What this adds

- `elpstest/parity.go`: `ParityCheck` / `CheckParity` / `RunParityCheck`. Template + n forks vs n cold loads; every transaction's result compared via the existing `renderResult`, every environment's post-run state via `FingerprintEnv` under `templateOpts`. Two schedules (sequential with lazily taken forks; round-robin over eager forks); 1 or 2 hops. A raising transaction is a result to compare. The only harness error once sequences exist is a template that does not load: a cold environment that fails to load, and a fork that cannot be taken, are witnesses and the run continues over the other environments. The result comparison has two properties: `ParityPropertyRaises` (exactly one arm raised) and `ParityPropertyReturns` (values differ, or both raised differently) - split because #579's signature is an error where a fresh VM gives a value, while #576 and template-share give different values, and a needle both emit distinguishes neither. libschema's process-wide `_validation_fun_<n>` gensym is normalised in the parity comparison (same class as the per-Runtime `_fun<envID>` the fingerprint already normalises; pinned by the folded 579 test below).
- `elpstest/aliasguard_parity.go` (39c0e48): **parity folded into the alias guard** as a channel inside `CheckTransactions`, computed by `CheckParity` over the same `Tx` set with the check's substituted walker, one hop, sequential/lazy (the sweep already takes forks eagerly). Every existing caller gains it with no call-site change. It runs BEFORE the sweep: a transaction that raises on a fork and not on a cold load is a parity finding, and the sweep would only abort on it as a harness error, so `CheckTransactions` returns what it has. A parity template that fails to load is the load-asymmetry witness (the same program already loaded once). Fork is the only walker parity applies to: copy/Detach/macro-stamp rebuild a value inside one environment and have no cold counterpart; their analogue is `CheckWalker`'s mutation-probe property (stated as a definition, not a tested claim). Seams in existing files: 11 lines in `aliasguard_isolation.go` (287-297) and three build-count expectations in `aliasguard_broken_test.go` (parity builds 1 + len(Tx) environments). `aliasguard.go`, `fingerprint.go`, `scripts/mutation-proof.sh` and the Makefile are untouched.
- `elpstest/parity_fuzz_test.go`: `FuzzForkParity`. Wraps FuzzAliasGuard's generator (`generateAliasGraphFrom`), extended with numeric sequences, views (`cdr`, `rest`, `slice`), in-place mutations and a transaction sequence per environment. Seeds committed and pinned; `TestForkParity_KnownShapes` runs the whole corpus on every PR. No skips.
- `elpstest/aliasguard_parity_test.go`: `TestTransactionIsolation_SchemaValidatorCredential` runs the #579 program through the guard (minus the always-raising transaction the sweep forbids; `ExpectNoSharedNatives` deliberately unset with a documented exemption: `*validatorTag` is a zero-size stateless credential shared process-wide by design per 6ef3da5's own comment). Controls listed below.
- `elpstest/parity_test.go`, `elpstest/parity_sharing_test.go` (`!elpscheck`): controls and schedule pins.
- `elpstest/parity_view_link_test.go` (c684e7b): the view-link visibility check the restack note asked for, run rather than read — see below.

## The known gap is closed (c684e7b)

Program `(set 'p0 (list 30 10 20)) (set 'w0 (cdr p0))`, transaction `(stable-sort < p0) w0`. Cold: `'(20 30)`. Fork, before #602: `'(10 20)`. That was the positive control at b4f65c6, gated behind a narrow skip while the fix was not under this branch.

On the tree rebased onto #602, `TestForkParity_ViewSortGapStillOpen` went red exactly as its comment said it must the day the gap closed:

```
--- FAIL: TestForkParity_ViewSortGapStillOpen (0.00s)
    parity_fuzz_test.go:434: the pinned view+sort parity gap no longer reproduces (result witness: false, state witness at user:w0: false).
        If Fork now preserves backing-array sharing, that is the fix landing: delete this test, the
        viewSortGapShape skip in FuzzForkParity and TestForkParity_SkipIsNarrow, and keep viewSortGapSeed
        as an ordinary seed.  If Fork did not change, the parity oracle has been weakened.
```

So, per the recorded restack notes: the stable-sort skip in `FuzzForkParity`, `viewSortGapShape`, `TestForkParity_SkipIsNarrow` and `TestForkParity_ViewSortGapStillOpen` are deleted; `viewSortGapSeed` is a plain seed whose comment names #602, the fix commit b9153c3 and this restack; `TestParityGapSeedIsWhatItsCommentSays` is kept. The seed now passes under `TestForkParity_KnownShapes` and `FuzzForkParity` with no skip: it is the permanent negative control for #602 living in this branch. Restore the per-header cells array in `forker.val` and it, `TestForkParity_ViewLinkIsInvisibleToTheOracle` and #602's own `TestForkPreservesCellSlotAliasing` all go red (`go vet` clean on the weakened tree).

The restack note also asked whether #602's view link (`Native` = the root `*LVal`, `Int` = the offset, on `LSExpr` headers only) is visible to `FingerprintEnv` and `renderResult` between a cold environment and a fork. Confirmed by running: `TestForkParity_ViewLinkIsInvisibleToTheOracle` loads a program with a view of every producing shape at load scope (`cdr`, a two-hop `cdr`, `slice 'list`, `rest` of a vector, `slice 'vector`, `(append 'vector v)` with no values) cold and into a forked template, checks the premise (every view in both arms carries a live link — `IsCellView`, `CellView` ok — and the fork's links name the fork's own roots), and requires identical fingerprints under the parity options (6897 tokens, equal) and no witness over two schedules that sort through every root, observe every view and take a view inside a transaction (sequential one-hop; interleaved two-hop). The fingerprint keys headers on the `*LVal` and emits an `LSExpr`'s cells, never its `Native` or `Int`, and Fork remaps the link onto the fork's own root.

## 579 proof (39c0e48's tree; measured at 9bdfb2e, identical compared code)

`scripts/mutations/579-libschema-validator-credential.patch` applied in a scratch worktree, `go build ./...` OK, `go test ./elpstest/`:

```
--- FAIL: TestForkCheck_SchemaValidatorCredential (0.24s)
--- FAIL: TestTransactionIsolation_SchemaValidatorCredential (0.30s)
    Fork: a transaction on a fork raises exactly when it raises on a cold load of the same program
          environment 0, transaction 0: (s:validate T 3)
            cold: list ()
            fork: error: env0-tx0.lisp:1:1: bad-arguments: Value is not a schema constraint: #<builtin>. ...
    Fork: a transaction on a fork raises exactly when it raises on a cold load of the same program
          environment 1, transaction 0: (s:validate anon 3)
    Fork: a fork's reachable state after its transactions is the cold load's
```

Property counts: raises 2, state 1, returns/loads/forkable 0. The previous behaviour (only `TestForkCheck_SchemaValidatorCredential` red, zero property strings) no longer holds.

Needle uniqueness, each of the other seven patches applied in turn (build OK for all):

| patch | needle `…raises exactly when it raises…` |
|---|---|
| 397-fork-shares-funnames | absent (0) |
| 440-fork-carries-loc | absent (0) |
| 576-fork-map-memo | absent (0) |
| 578-f1-live-defining-loc | absent (0) |
| 582-macro-stamp-in-place | absent (0) |
| 585-detach-memos | absent (0) |
| template-share | absent (0) |

Intended manifest row (follow-on, not in this PR): `579-libschema-validator-credential;a transaction on a fork raises exactly when it raises on a cold load of the same program|TEST:TestForkCheck_SchemaValidatorCredential;`.

## Skip-disabled sweep (ef0831e, throwaway, before #602)

| class | ran | passed | failed | skipped via error |
|---|---|---|---|---|
| sort | 4,536 | 1,733 | 2,803 | 10 |
| no-sort | 1,451 | 1,451 | 0 | 3 |

All 13 skips are a template that does not load (FuzzAliasGuard's generator recording the wrong kind for a quasiquote of a quasiquote); load-asymmetry and fork-refusal witnesses 0. Sort with no view anywhere: ran 33 / failed 0. Every reachable failure was the view+sort class — the class #602 closes and c684e7b's 120 s run below no longer skips.

## Negative controls

- `TestForkParity_DetectsADealiasingFork`, `TestForkParity_DetectsASharingFork` (`!elpscheck`), `TestForkParity_DetectsAForkRefusal` (production `Fork` refuses via `checkQuiescent`), `TestForkParity_DetectsAnAsymmetricLoad`, `TestForkParity_DetectsARaiseAsymmetry`.
- Fold controls: `TestTransactionCheckCarriesParity` (a de-aliasing fork through `CheckTransactions` reports the value property), `TestTransactionCheckSurvivesAForkThatRaises`, `TestTransactionCheckReportsAParityLoadAsymmetry`, `TestParityBuildsOneTemplatePlusOneColdPerTransaction`.
- Weakened-tree battery (each mutation in a scratch worktree, `go vet` clean, reverted): result comparison, state comparison, cold-load witness, fork-refusal witness, continue-past-dead-fork, lazy forks, Hops, view generation, base kinds, the channel seam, the raise/return split, the early return on a raise, the gensym normalisation, the template-error mapping - each reddens exactly its control. (The fuzz-skip and skip-predicate rows are gone with the skip.) Plus, at c684e7b: the per-header cells array restored in `forker.val` reddens the seed under `TestForkParity_KnownShapes`, `TestForkParity_ViewLinkIsInvisibleToTheOracle` and #602's `TestForkPreservesCellSlotAliasing`.

## Fuzz budget

Unchanged, still red, deliberately untouched: `make fuzz-budget-check` at c684e7b reports 5 × 10 min + 10 min = 60 min vs timeout-minutes 60 — headroom 0 min, less than the one-FUZZTIME (10 min) margin a sweep must keep spare (41 targets / 10 shards). Matrix untouched; with the 11th shard the alias-guard branch adds: largest shard 4, headroom 10, room for 3 more.

## Throughput of FuzzAliasGuard with the parity channel

Fixed execution count (`-fuzztime 300x`, arms alternated twice, box load 5.5-8.9 with other agents' suites running; the two 120 s arms at load 15-18 were inside the noise floor and are not used):

| round | ef0831e | 39c0e48 |
|---|---|---|
| 1 | 11.3 s | 14.3 s |
| 2 | 11.5 s | 12.1 s |

+13 % per execution on average (+5 % to +27 %), under the one-third threshold, so parity is unconditional in the target; no sampling.

## 120 s fuzz runs of FuzzForkParity

b4f65c6: 47,058 execs, PASS. ef0831e: 28,649 execs, PASS (concurrent sweep on the box). c684e7b, skip gone: 23,960 execs, PASS, no findings (107 new interesting inputs, corpus 427; box 1-min load 17 falling to 10 during the run, with two other agents' suites winding down — the exec rate is depressed by that, the verdict is not).

## Test plan (c684e7b, rebased onto 4255801)

- rebase of 39c0e48 onto 4255801: clean, four commits replayed, no conflicts
- `go test ./elpstest/ -run TestForkParity_ViewSortGapStillOpen` before step 3: red, output above
- `go test ./...` ok; `go test -tags elpscheck ./...` ok; `go test -race ./elpstest/...` ok
- `make static-checks`: first pass 0 issues; the elpscheck-tagged pass timed out at box load ~30 (three agents' suites) and was re-run alone as the Makefile runs it (`golangci-lint run --build-tags elpscheck ./lisp/...`, longer `--timeout`): 0 issues. `gofmt -l` on touched files empty
- `make elpsvet`: both passes exit 0
- `make fuzz-budget-check`: red as reported above, untouched
- weakened-tree checks as above
- 120 s `FuzzForkParity` as above

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -199,21 +199,23 @@ static-checks: check-golangci-version check-golangci-config
 # mutation-proof: revert each REAL historical fix in production code and
 # require it to be caught, by a needle measured for uniqueness and stability.
 #
-# For EIGHT of the nine rows that needle is a property string emitted by the
-# guard this PR adds. "By name" means the SPECIFIC property, not "some test
-# failed": needles shared across mutations assert nothing about the bug they
-# are filed under, and a needle that is only ~84% stable makes a required gate
-# flaky, which is worse than no gate.
+# For all NINE rows that needle is a property string emitted by the guard.
+# "By name" means the SPECIFIC property, not "some test failed": needles
+# shared across mutations assert nothing about the bug they are filed under,
+# and a needle that is only ~84% stable makes a required gate flaky, which is
+# worse than no gate.
 #
-# THE 579 ROW IS AN EXCEPTION, AND IT IS MEASURED, NOT CONCEDED.
-# Reverting that fix emits no property string at all -- it reddens exactly one
-# pre-existing test, from the earlier forkcheck oracle (477ea95), which is not
-# in this PR's diff. So that row asserts "#579 stays fixed" rather than "the
-# new guard catches #579". The manifest notes in scripts/mutation-proof.sh
-# record why, and this comment says so here rather than leaving the sentence
-# above to overstate what all nine rows demonstrate. (It closes at the top of
-# the stack: #601 folds cold-vs-fork parity into this same harness, and the
-# property row for #579 lands there, where the property exists.)
+# The 579 row was an exception until cold-vs-fork PARITY became a channel of
+# this same harness (PR #601). Before that, reverting 6ef3da5 emitted no
+# property string at all -- it reddened exactly one pre-existing test from
+# the earlier forkcheck oracle (477ea95) -- because #579 is a credential
+# revoked by HEADER IDENTITY, which none of the guard's original three
+# channels observe. The caveat is now CLOSED rather than reworded: the parity
+# channel emits a property string for it, carried as a must-NOT on every
+# other row so its uniqueness is re-measured by the gate itself, and the row
+# retains the TEST: needle as a second signal. The manifest notes in
+# scripts/mutation-proof.sh carry the history and why the row lives at the
+# top of the stack (the property exists from #601 upward, nowhere below).
 #
 # The ten broken reference walkers in elpstest/aliasguard_broken_test.go model
 # those bugs with hand-written imitations. This reverts the actual fixes. The
--- a/elpstest/aliasguard_broken_test.go
+++ b/elpstest/aliasguard_broken_test.go
@@ -1555,12 +1555,12 @@ func TestTheConcurrentArmIsSkippedOnRequest(t *testing.T) {
 	if err != nil {
 		t.Fatalf("harness error: %v", err)
 	}
-	if builds != 1 {
-		t.Errorf("SkipConcurrentArm built %d environments, want 1.\n"+
+	if want := 1 + parityEnvBuilds(concurrencyProbeTx); builds != want {
+		t.Errorf("SkipConcurrentArm built %d environments, want %d.\n"+
 			"Two means the CONCURRENT ARM RAN anyway. A walker that declares this shares a payload,\n"+
 			"so driving two of its forks in parallel mutates one *MapData from two goroutines -- a\n"+
 			"data race by construction, reported against this guard's own tests and taking every\n"+
-			"other parallel test down with it.", builds)
+			"other parallel test down with it.", builds, want)
 	}
 }
 
@@ -1577,12 +1577,12 @@ func TestTheConcurrentArmStillRunsForARealWalker(t *testing.T) {
 	if err != nil {
 		t.Fatalf("harness error: %v", err)
 	}
-	if builds != 2 {
-		t.Errorf("the real Fork walker built %d environments, want 2.\n"+
+	if want := 2 + parityEnvBuilds(concurrencyProbeTx); builds != want {
+		t.Errorf("the real Fork walker built %d environments, want %d.\n"+
 			"One means the CONCURRENT ARM DID NOT RUN. That arm is the -race gate for template\n"+
 			"mutation under interleaving, and it is the coverage the substituted-fork skip exists to\n"+
 			"protect -- a skip that also swallows real walkers has removed the property instead of\n"+
-			"narrowing it.", builds)
+			"narrowing it.", builds, want)
 	}
 }
 
@@ -1612,11 +1612,11 @@ func TestTheConcurrentArmStillRunsForASubstitutedWalker(t *testing.T) {
 	if err != nil {
 		t.Fatalf("harness error: %v", err)
 	}
-	if builds != 2 {
-		t.Errorf("a BENIGN substituted walker built %d environments, want 2.\n"+
+	if want := 2 + parityEnvBuilds(concurrencyProbeTx); builds != want {
+		t.Errorf("a BENIGN substituted walker built %d environments, want %d.\n"+
 			"One means the CONCURRENT ARM DID NOT RUN merely because Fork was non-nil. That is the\n"+
 			"old axis: substitution is not a declaration that the walker shares, and an embedder\n"+
 			"wrapping Fork for instrumentation must not silently lose the -race gate. Key the skip\n"+
-			"on SkipConcurrentArm.", builds)
+			"on SkipConcurrentArm.", builds, want)
 	}
 }
--- a/elpstest/aliasguard_cellview_test.go
+++ b/elpstest/aliasguard_cellview_test.go
@@ -205,31 +205,37 @@ func TestReachableWalkFollowsALiveRoot(t *testing.T) {
 }
 
 // dealiasingOn returns a Fork substitute that de-aliases the views on
-// exactly the calls `when` selects (1-based call number), and is the real
-// Fork otherwise.  CheckTransactions takes len(Tx) forks before running the
-// transactions and then ONE more for the pristine successor, so calls
-// 1..len(Tx) are the fresh forks and call len(Tx)+1 is the successor.
-func dealiasingOn(when func(call int) bool) func(*lisp.LEnv) (*lisp.LEnv, error) {
-	call := 0
-	return func(env *lisp.LEnv) (*lisp.LEnv, error) {
-		call++
-		if when(call) {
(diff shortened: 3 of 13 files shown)


sam-at-luther:

The replacement FuzzForkParity compares multiple independent cold loads with template instances, checks values/state/effects, and exercises lazy, interleaved and concurrent schedules. Historical seed programs and equivalent negative controls remain; one schedule per input plus complementary seed tests preserves coverage without the old watchdog cost. See [the runner and controls](https://github.com/luthersystems/elps/blob/v1.61.0/elpstest/template_fuzz_test.go).

Issue #373, opened 2026-08-11
Document Go-style append/slice capacity sharing; consider capacity-clamped slice; document concat as the copy idiom


sam-at-luther:
Decision (maintainer): keep Go semantics — append writes into shared capacity like Go's append does — and document the behavior and its pitfalls rather than adding copy-on-append.

Rationale: detecting shared backing cheaply is not possible without reference counting, and copy-on-every-append turns amortized O(1) vector building into O(n^2). The one dangerous instance of the class (program literals shared across environments) is mechanically fixed by the seal/copy-on-write design (seal propagates through slice; append copies on sealed input), so what remains is ordinary within-environment value aliasing, i.e. language semantics, consistent with the language's deliberate mirroring of Go slices.

Work items:
1. Document on the append/slice docstrings and language docs: slice shares backing including retained capacity, like Go; appending to a slice result can write into the source's backing.
2. Document `concat` as the safe-copy idiom: `(concat 'bytes b)` / `(concat 'vector v)` / `(concat 'list xs)` allocate fresh exactly-sized backing (safe on the empty case since #334). Go-side deep-copy for transfer cases is `Detach()`.
3. Evaluate: capacity-clamped slice results (Go's three-index `s[a:b:b]` idiom, built in) — reads stay zero-copy, but append on a slice result can never write into the source's retained tail; behavior changes only in the footgun cases; index writes on the shared prefix still alias as expected. If benchmarks show no regression, this shrinks the documented pitfall substantially.

The mutation sites remain annotated `//elps:mutates` (branch claude/exp-vet-alias) with the caveat disclosed.

sam-at-luther:
**Raising severity: this reproduces on pristine `main` with kernel builtins alone, and it is what #392 actually was.**

Independently verified on `663c0f3`, with `libelpspath` genuinely absent from the tree (`ls lisp/lisplib/ | grep -c elpspath` → `0`), using only `slice` and `append!`:

```lisp
(set 'v (vector 10 20 30 40))
(set 'view (slice 'vector v 0 2))
(append! view 999)
(debug-print (list 'source v 'view view))

'('source (vector 10 20 999 40) 'view (vector 10 20 999))
```

The source vector's third element is rewritten by an append to a slice *view* of it. No library involved, no seal involved — `slice` hands out a view retaining capacity into the source's backing array, and `append!` writes through it.

## This changes what #392 was

[...]

PR #602, opened 2026-09-04
Preserve cell-view slot aliasing across Fork

sam-at-luther:
Stacked on #599 (base: `claude/elps-perf-optimization-jcgfd2-alias-guard`, at 74e4ac8).

## The property

In the repo owner's words: "For two VMs that get instantiated from the same source code, ELPS code should work identically to two forks from the same template. Fork was meant to be a performance optimisation."

## The bug

`cdr` (and `rest`, `slice`) return a view: a fresh header whose `Cells` is a three-index reslice of the source list's backing array, sharing its slots. A destructive in-place operation through either header is visible through the other. That is language behaviour, and `docs/fork.md` promised a fork reproduces it ("aliasing and cycles preserved").

It did not. `forker.val` allocated one array per header, so two headers over one template array came out of the fork over two arrays. The elements stayed shared (the value memo); the slots did not.

```lisp
;; loaded into the template
(set 'l (list 30 10 20))
(set 'tail (cdr l))
;; then run as a transaction
(stable-sort < l)
tail
```

Cold environment: `'(20 30)`. Template: `'(20 30)`. Fork: `'(10 20)`.

## Why the address-sort approach was rejected

The first fix built and measured (option A, not in this PR) recovered the relationship at fork time: record every copied header, sort by start address, merge overlapping ranges. It is correct, but the sort is paid per mutable cell-bearing value on every fork, whether or not anything aliases. Measured on this box, base 74e4ac8 vs that implementation, interleaved n=12, repo gate:

- `ForkSortedMap` sec/op +33.89% (p=0.000) against the 15% gate; B/op +10.25% against the 5% gate. That template holds 6141 mutable cell-bearing values and zero views.
- `EnvConstruction/mode=fork` +2.15% / +4.83%.
- Attribution, measured: ~3/4 of the added time was the sort, ~1/4 the record slice.

So the goal stayed and the method changed.

## The design

Record the view-to-parent relationship where the view is made; resolve it in O(1) per header at fork. No sort, no record slice, no `unsafe`.

- A view header carries its root (the first header in its ancestry that owns its `Cells`) in `Native` and its element offset into the root's `Cells` in `Int`. Both fields are otherwise unused on an `LSExpr`, so `sizeof(LVal)` stays at 112 bytes, before and after (`TestLValSizeUnchanged`, `TestCellViewLinkNeedsNoField`). A dedicated `*LVal` field would have made it 120 bytes, +7% on every value in the system, which would have tripped the 5% allocation gate broadly.
- Root, not immediate parent: a view of a view resolves in one hop, offsets compose by addition, and an iterated `(set 'x (cdr x))` chain retains one header instead of every intermediate. `(cdr (cdr l))` links to `l` at offset 2.
- Producers that link (every caller of `clampCap` that hands a shared reslice back to lisp, plus the one `append` shape): `cdr`; `rest` (list and vector); `slice` for the `'list` window and for the `'vector` wrapper over it; `(append 'vector seq)` with no values, the one input the #373 clamp does not reallocate. `Quote`/`Splice`/`shallowUnquote` header copies propagate an existing link through the struct copy; they do not create one for a root, because that shape is not reachable for mutable list data from lisp and linking would retain transient quasiquote results.
- `forker.val` copies the root, hands the view the matching window of the copy clamped to its length, and remaps the link onto the fork's root. A root's cell slice is now published before it is filled, so a view reached through the root's own cells finds an array to window.
- The link is advisory. The re-point happens only after checking, on the template, that the view's element 0 still is `root.Cells[off]` (slot identity). A header whose `Cells` was reassigned since (`append!` reallocating a vector, elpspath's in-place set, embedder Go code writing the exported field) has already parted from its root in the template, so the fork copies it privately, which is what the template holds.

## Isolation resolutions

- Length ranges, not capacity: spare-capacity aliasing is deliberately not preserved. A view is clamped to its length on both sides, so an append through it reallocates rather than writing a neighbouring slot of the root. Isolation wins (#373). Stated in `docs/fork.md`.
- The `LFun` shared-cells fast path is untouched; the `len==0` branch is untouched.
- Copy and detach allocate fresh storage and drop the link (`TestCopyAndDetachDropCellViewLink`); `TestCopyDoesNotPreserveBackingArraySharing` still passes unchanged.
- A view of a sealed list is not linked: Fork shares sealed values outright.
- A view made inside a fork links to the fork's own root, never template memory (`TestForkCellViewMadeInTransaction`, including a fork of the fork).
- A view whose root is unreachable from the fork's roots is still copied correctly, through the link (`TestForkCellViewUnreachableRoot`).
- The walker registry's payload scan caught the new `*LVal` arm in detach; it is exempted with the reason (it names another header, which both rebuilding walkers already memoise as `PayloadValue`).

## Permanent control

`TestForkPreservesCellSlotAliasing` (elpstest) pins `'(20 30)` on the fork. Reintroducing the per-header fresh array turns it red; verified on a weakened tree, as were the other new assertions (each with the thing it checks deleted, `go vet` clean, named test red).

## Benchmarks

Method: `scripts/bench-run-arms.sh` (verbatim, package set narrowed to the three packages holding every Fork benchmark — `lisp`, `lisp/lisplib`, `lisp/lisplib/libjson` — so twelve interleaved rounds fit inside a quiet window on a box shared with two other agents' suites), then `benchstat` and `cmd/benchgate` with the shipped waiver file. Base 74e4ac8 vs b6adfac, interleaved, n=12 each, 22:49–22:58 on an idle 4-vCPU Xeon @ 2.10GHz; a sampler logged foreign CPU consumers every 60s for the whole window and found none (two earlier attempts were aborted when other agents' suites started mid-run; their partial output is discarded). `BenchmarkForkViews` was copied into the base tree as an uncommitted test-only file so its rows pair.

Gate verdict: **PASS** — `benchgate: interpreted 38 delta row(s) + 391 no-change row(s); 17 significant move(s) in the bad direction; 0 at or above the gate (timing 15%, allocation 5%)`.

```
pkg: github.com/luthersystems/elps/lisp                  base (74e4ac8)   pr (b6adfac)
ForkManyFunctions-4      sec/op      304.4µ ±  7%     311.1µ ±  8%        ~ (p=1.000 n=12)
ForkSortedMap-4          sec/op      3.115m ±  9%     3.113m ±  5%        ~ (p=0.932 n=12)
ForkManyFunctions-4      B/op       369.5Ki ±  0%    369.5Ki ±  0%        ~ (p=0.403 n=12)
ForkSortedMap-4          B/op       2.468Mi ±  0%    2.468Mi ±  0%        ~ (p=0.478 n=12)
ForkManyFunctions-4      allocs/op   1.617k ±  0%     1.617k ±  0%        ~ (all samples equal)
ForkSortedMap-4          allocs/op   20.84k ±  0%     20.84k ±  0%        ~ (p=0.268 n=12)
geomean (sec/op)                     34.04µ           34.38µ         +1.00%

pkg: github.com/luthersystems/elps/lisp/lisplib
EnvConstruction/mode=fork-4     sec/op    187.8µ ± 7%    194.9µ ± 6%        ~ (p=0.514 n=12)
EnvConstruction/mode=fullload-4 sec/op    321.8µ ± 3%    324.7µ ± 5%        ~ (p=0.977 n=12)
ForkViews/arm=lists-4           sec/op    275.6µ ± 6%    263.2µ ± 3%        ~ (p=0.219 n=12)
ForkViews/arm=views-4           sec/op    361.5µ ± 6%    356.0µ ± 4%        ~ (p=0.630 n=12)
EnvConstruction/mode=fork-4     B/op     290.4Ki ± 0%   290.4Ki ± 0%        ~ (p=0.471 n=12)
ForkViews/arm=lists-4           B/op     335.9Ki ± 0%   335.9Ki ± 0%        ~ (p=0.880 n=12)
ForkViews/arm=views-4           B/op     389.1Ki ± 0%   382.1Ki ± 0%   -1.81% (p=0.000 n=12)
EnvConstruction/mode=fork-4     allocs/op 1.235k ± 0%    1.235k ± 0%        ~ (all samples equal)
ForkViews/arm=lists-4           allocs/op 1.791k ± 0%    1.791k ± 0%        ~ (all samples equal)
ForkViews/arm=views-4           allocs/op 2.391k ± 0%    2.091k ± 0%  -12.55% (p=0.000 n=12)
geomean (sec/op)                          222.4µ         220.4µ        -0.89%
geomean (B/op)                           267.5Ki        266.7Ki        -0.30%
geomean (allocs/op)                       1.755k         1.717k        -2.21%

pkg: github.com/luthersystems/elps/lisp/lisplib/libjson
ForkDecodedMaps-4        sec/op      825.1µ ±  7%     832.3µ ±  7%        ~ (p=0.755 n=12)
ForkDecodedMaps-4        B/op       819.9Ki ±  0%    820.0Ki ±  0%        ~ (p=0.521 n=12)
ForkDecodedMaps-4        allocs/op   5.006k ±  0%     5.006k ±  0%        ~ (all samples equal)
```

Reading it: on templates without views the change is unmeasurable — every existing fork row is `~` with byte-identical B/op and allocs/op (the whole per-value cost is one `Native` type check in `forker.val`). On the shape the fix exists for, `ForkViews/arm=views`, the fork is cheaper: the 300 per-view arrays are gone (−300 allocs/op, −7 KiB/op), replaced by windows onto the root copies; time is within noise.

The seventeen significant below-gate moves are all sec/op on ns-scale, zero-allocation benchmarks outside Fork (`Equal/list` 26ns → 28ns, `PackageGetFun` 10.1ns → 10.8ns, and the like), with `Equal/string` and `EqualDistinct/int` moving −10% in the same run; allocs/op and B/op are unchanged on every one of them. Re-measured on their own (same arms, interleaved, n=12, quiet window): the `Alias*`, `Equal/list` and `EqualDistinct/list` moves did not reproduce (`~`). `Equal/deep` (+9.90%, p=0.000) and `EqualDistinct/deep` (+10.53%, p=0.004) did, for the third time across three runs, alongside `Equal/string` −11.74% and `EqualDistinct/int` −13.63% in the same run; `PackageGetFunAliased` +6.84% (p=0.013) likewise. `equal` and package lookup never reach any changed code, and allocations are identical, so these are instruction-layout shifts from the inserted code (cell views add ~15 lines to `builtins.go` and ~45 to `fork.go`, both of which precede `lisp.go` in link order) — recorded here rather than explained away, and below the gate on every measurement.

Corroboration: the first six rounds of an aborted full `./...` run (n=6, before another agent's suite started) gave the same picture — every fork row `~`, `benchgate` exit 0, zero rows at or above the gate.

Rejected alternative, for the record (measured earlier on the same box, same method, n=12, base 74e4ac8): the address-sort approach gave `ForkSortedMap` +33.89% sec/op and +10.25% B/op — a gate failure on a template with no views.

## Follow-up on review (4255801)

- `make elpsvet` was red on b9153c3 (CI job 101200856221): four findings, `linkCellsView` writing `view.Native`/`view.Int` and `clearCellsView` writing `v.Native`/`v.Int` on values the helper did not construct. Fixed structurally, with no `//elps:mutates` annotation anywhere: `newCellsView` builds the view header field by field (no struct copy) and sets the link on the value it is building; `cdr`/`rest`/`slice` return it directly; the two vector-producing sites build their holder with it and wrap it with `vectorFromHolder`; `Copy` and `detach` clear the link inline on the copy they are building. Both helpers are deleted, so "only a constructor writes the link" is now enforced by elpsvet rather than promised by a comment.
- Exported accessors, the only way code outside `lisp` reads the convention: `IsCellView()` (cheap, unvalidated — does this header carry a link; keeps a view's `Native` out of a walker's payload arm even when stale) and `CellView() (root, off, ok)` (validated — `ok` only when element 0 still is `root.Cells[off]` and the window fits). `forker.val` now calls `CellView()` instead of inlining that check, so what counts as a live link is one rule in one place. `TestCellViewExportedAccessors` pins both stale shapes and that Fork copies such a view privately through the same call; weakened (check dropped inside `CellView`) it goes red on both that test and `TestForkCellViewStaleLinkCopiesPrivately`, vet clean.
- The convention is stated once, on `cellsView` in `lisp/lisp.go`; the `Native`/`Int` field docs, `forker.val`, `detach`, the walker registry's exemption row and `docs/fork.md` all point there rather than restating it.
- No re-benchmark: the change is structural — the same field stores moved into the constructor, the same slot-identity comparisons moved into `CellView`, and `vectorFromHolder` builds exactly the two-cell shape `Vector` does — so the instructions Fork executes per value are unchanged.

## Test plan

On 4255801 (and on b6adfac for the benchmark rows):

- `make elpsvet` — both passes clean: `go run ./cmd/elpsvet -test=false ./...` exit 0, `GOFLAGS=-tags=elpscheck go run ./cmd/elpsvet -test=false ./...` exit 0 (on b9153c3 the first pass exited 3 with the four findings above)
- `go test ./...` — pass
- `go test -tags elpscheck ./...` — pass
- `go test -race ./lisp/... ./elpstest/...` — pass
- `make static-checks` — 0 issues (both builds); gofmt clean on every touched file
- Weakened-tree sweep (each new assertion deleted in turn, `go vet` clean, named test red): 11/11 red
- Benchmark: `scripts/bench-run-arms.sh` (base 74e4ac8 vs b6adfac, interleaved, `BENCH_COUNT=12`) then `benchstat` and `cmd/benchgate` with the shipped waiver file; foreign-process and load samples recorded every 60s during the run

## Fuzz campaign finding (c2682c6)

A 10-minute CI fuzz campaign on the top of this stack (#603 at 590300e; run 33941632540, shard 6, job 101240091335) found `FuzzSequenceOps/7192dc3ce712e22c`:

```
step 1: append! mutated SEALED storage in place (fingerprint 7d096b0c1bb0704d -> b9c9d6a5c3bbc47a)
  [0] array: (vector ())
  [1] list: ()
```

**Bisect**, one targeted run per tree (`go test -run 'FuzzSequenceOps/7192dc3ce712e22c' ./lisp`, release and `-tags=elpscheck`, seed placed under `lisp/testdata/fuzz/FuzzSequenceOps/`):

| tree | result |
|---|---|
| main a32ada7 | pass |
| #599 febe667 | pass |
| #602 8650ccc | **FAIL** |
| #601 5ce4fe5 | FAIL (identical) |
| #603 590300e | FAIL (identical) |

So the defect is this branch's. It is not a write through a view into a root, and the shared `()` singleton is untouched. Traced: step 0 is `(append 'vector '())` on a **sealed empty** list, which this branch routed through the new `len(vals)==0` branch → `newCellsView`, and `newCellsView` copies `sealed: parent.sealed` onto the header it builds — right for `cdr`/`rest`/`slice 'list`, whose result *is* the sealed storage, but here that header became the new vector's **data holder**, which was born sealed. `lisp/seal.go`'s contract is that a vector's holder never is (`append!` writes it without consulting the flag — that is why it needs no guard — and `InheritSeal` refuses to mark an array for the same reason). Step 1's `append!` then grew a "sealed" holder, which is the oracle's report. Before this branch the carve-out path was `Array(nil, clampCap(cells))`: a fresh, unsealed holder over a zero-capacity window.

Measured on the buggy tree, the user-visible consequence is worse than the oracle's message — every later view of such a vector inherited the stale flag and the guarded sites refused plain runtime data:

```lisp
(set 'v (append 'vector '()))       ; => (vector)
(append! v 3 1 2)                   ; => (vector 3 1 2)
(slice 'vector v 0 2)               ; => modify-literal-error
(stable-sort < (slice 'list v 0 3)) ; => modify-literal-error
```

Same for `(slice 'vector lit 0 0)`, `(slice 'vector '(1 2 3) 1 1)` and `(append 'vector (rest '(1)))`.

**Fix** (`lisp/lisp.go`, `lisp/builtins.go`): one helper, `vectorView(parent, off, cells)`, is now the only path from a cell window to a vector. An empty window is not a view — no slot to share, no link to record — so it returns `Vector(nil)` exactly as before views existed; only a non-empty window (which both callers have already established is unsealed, since they refuse a sealed non-empty input first) goes through `newCellsView` + `vectorFromHolder`. Non-empty behaviour is unchanged: views still preserve slot aliasing across Fork, a sealed non-empty input still raises `modify-literal-error`, nothing copies on write.

**Tests**: the crasher is committed as a permanent seed at `lisp/testdata/fuzz/FuzzSequenceOps/7192dc3ce712e22c` (the convention `FuzzSharedTreeEval` uses). `TestSealedWriteEmptyCarveOut` now also asks the vector's data holder (`Cells[1]`) whether it is sealed — it only asked the vector header before, which is never sealed, and that is how the class stayed green. New `TestSealedEmptyCarveOutVectorIsMutable` drives all seven carve-out forms through `append!`, `slice 'vector`, `stable-sort` through a `slice 'list` view (the slot aliasing a runtime vector keeps), the no-values `append` chain, and re-reads the literal. Both fail on 8650ccc and pass on c2682c6, release and `elpscheck`; `sequence_fuzz_test.go` is untouched.

Local gate on c2682c6: `go build ./...`, `go vet ./lisp/... ./elpstest/...`, gofmt on touched files, `golangci-lint run ./lisp/...` (0 issues), elpsvet on `./lisp/...` in both builds, and targeted runs of the cell-view / seal / alias-guard tests in `lisp`, `elpstest`, `lisp/lisplib` (both builds) — all green. #601 and #603 were rebased onto c2682c6 (plain rebase, `--force-with-lease` from 5ce4fe5 and 590300e): #601 → 056129e, #603 → b076a2f; the seed and both tests pass on b076a2f in both builds.


sam-at-luther:

Publication-time physical backing reconstruction preserves cell-slot aliases, overlapping views and capacities without advisory cell-view links. The historical sort/view and empty-vector behaviors remain covered by [backing-span tests](https://github.com/luthersystems/elps/blob/v1.61.0/lisp/template_span_test.go) and [legacy regressions](https://github.com/luthersystems/elps/blob/v1.61.0/lisp/template_legacy_regression_test.go).

Issue #600, opened 2026-09-04
Alias-guard channels still unwatched in copy/detach/(*LVal).Copy and the macro stamper


sam-at-luther:
A coverage audit was run against the alias/location guard (#598, PR #599) at `ac5a2e9`. It asked a different question from the six adversarial review rounds: those asked whether the guard is *correct*, this asked whether it is *complete* — are there channels through which a fork can leak that the guard does not watch at all?

Method: swept `git log main` (~250 commits), the issue list and merged PRs back to #196; named the channel each historical fix concerned; cross-referenced against what `fingerprint.go` encodes and explicitly excludes, the probe site kinds in `aliasguard.go`, the location channel, the isolation properties, and the memo registry in `lisp/walkers.go`.

Five uncovered channels, three demonstrated with running code.

## 1. A native payload on a non-`LNative` header is invisible to every native check

The kernel keys off `v.Native != nil` (`lisp/fork.go:507` — the arm for a payload the kernel has no copy strategy for, riding on a non-LNative type). The guard keys off `v.Type == LNative`: `fingerprint.go:399`, `aliasguard_isolation.go:734`, `aliasguard.go:964`.

**Demonstrated.** An `*auditCounter` on an `LSExpr` header came out of `env.Fork()` as the **same pointer** in template and fork, while `SharedNativePayloads` reported 0, `NativeDeclarations` did not list the type, and `CheckTransactions{ExpectNoSharedNatives: true}` returned zero witnesses. At the fingerprint level: two `LSExpr` headers with *different* payloads fingerprint equal, and one payload under two headers fingerprints identically to two payloads — the #576 distinction is unavailable for this shape.

Not hypothetical: `newTextLoaderWalk` sanctions it explicitly ("Native is the ONLY exported per-node slot an embedder's Reader has for annotation").

Fix is about three lines: replace the `v.Type == LNative` predicate with `isPointerPayload(v.Native)` in all three walks.

Note for context: substrate already guards its own side of this class statically — `cmd/substratevet`'s `nativepayload` rule, from substrate#448 — by checking native *construction spellings* rather than header type. The two approaches are complementary; this issue is the elps-side gap.

### Re-measured at `74e4ac8` with a paired control

A control at PR #599's head confirms it, with the **real** `(*lisp.LEnv).Fork` rather than an injected walker — the kernel's `default:` arm shares the payload on its own:

| header carrying the same shared `*hiddenNative` | witnesses |
|---|---|
[...]

PR #603, opened 2026-09-04
Run the deployed load-cache topology through the isolation oracles (#600 gap 6)

sam-at-luther:
Stacked on #601 (parity-fuzz, `056129e`), which is on #602 (`c2682c6`), which is on #599 (`febe667`). Head `b076a2f` (rebase of `65de8aa` onto the fixed chain, no content change): two new files, plus three comment-only edits to existing files that cited the pinned-gap test by its former name (see finding 1) — no code line changed in them.

## The property

In the owner's words: "For two VMs that get instantiated from the same source code, ELPS code should work identically to two forks from the same template. Fork was meant to be a performance optimisation." Parity with cold VMs is the spec; isolation and fidelity are consequences.

## The hole

substrate's deployed shape is ONE preheated template, N forks (one per transaction) and ONE warm `lisp.LoadCache` shared by all of them — the cache travels across `Fork` (`lisp/fork.go`), and `docs/fork.md` says that is safe because every entry is immutable and sealed. On 74e4ac8 every load-cache test (`FuzzLoadCacheMultiEnv`, `FuzzLoadCacheHostileReader`, the custody and sharing tests) runs the cache over INDEPENDENT environments, and no `TransactionCheck`/`AliasCheck` installs a cache at all. The topology an embedder actually ships had never been run through an isolation or parity oracle.

## What the tests assert

`elpstest/loadcache_topology_test.go`

- `TestLoadCacheTopology_CheckParity` — the topology through `elpstest.CheckParity` (the headline oracle, here via #601). One template whose six forks share its cache against six cold environments, each on a fresh cache of its own; `NewEnv` hands out and records a new cache per call, so the warmth is asserted by role afterwards: exactly one cache served hits (the template's, 7 = five on `shared.lisp` + two on `late.lisp`), every cold cache served none, every cache stored `shared.lisp` once. Per-step results compared, including an unhandled `(stable-sort < shared-lit)` that must RAISE identically on both arms. Both schedules × both hop counts. **No witness on any arm** (`loads=56 hits=7 stores=49` on each).
- `TestLoadCacheTopology_Isolation` — `CheckTransactions` with one cache shared by the template, six forks, the concurrent arm's second template, and the parity channel's template and cold environments. Every property holds, sequentially and concurrently (the concurrent arm is the `-race` gate). The cache is shown to be doing its job: exactly 17 stores — 11 as on 74e4ac8 (10 distinct sources + property 5's `tmpl-tx0.lisp`) plus the parity channel's six `env<i>-tx0.lisp` — and every other load a hit. Measured: `loads=80 hits=63 stores=17`.
- `TestLoadCacheTopology_Parity` — `RunForkCheck` with a fresh cache per `NewEnv`: fork and fork-of-fork arms HIT what the template and earlier forks minted while each cold arm MISSES in its own cache.
- `TestLoadCacheTopology_SharedEntryIdentity` — the documented design asserted explicitly. Every fork loads `shared.lisp` after forking (the template never binds it, asserted), so a node two forks hold in common can only have come through the cache. Sequential arm: one store, five hits, all six forks hold the SAME sealed `*LVal`; `(stable-sort < shared-lit)` is refused identically on every fork (`'('refused 'modify-literal-error)`, compared fork-vs-fork and fork-vs-cold); every fork's fingerprint (`Seal + PackageMetadata`) equals a cold environment's on its own fresh cache — a shared sealed entry is indistinguishable from a private one. Concurrent arm asserts the contract rather than the sequential counts (finding 2).
- `TestLoadCacheTopology_NativeAnnotationIsReported` — hostile-reader angle, now the closed-gap negative control (finding 1).

`elpstest/loadcache_topology_unsealed_test.go` (`//go:build !elpscheck`, because the ownership checker panics on an unsealed `*LVal` reaching two Runtimes before any property can report — same reason as `aliasguard_templatefork_test.go`).

## Controls and red-on-weakening evidence

Every weakening was applied to the frozen tree, run with `go vet ./elpstest/` clean on the weakened tree, then reverted. Raw output is quoted.

**(a) Cache entries NOT sealed** — `TestLoadCacheTopology_UnsealedSharedEntryIsAWitness`. `CachedSource` is opaque (only `newCachedSource` mints one, through an admission that seals or copies-and-seals), so the real hook cannot produce this shape; the double is a `Runtime.Reader` that parses once, `Copy()`s (clears the seal), and hands the same unsealed slice to every load, no `LoadCache` installed — substrate's pre-#368 parse cache. One reader per template (`NewEnv`), shared with its forks because `Reader` travels across `Fork`. Fork 0 loads and reads; fork 1 loads and `stable-sort`s in place. BOTH `CheckTransactions` channels report it:
```
Fork: a fork's reachable state after its transactions is the cold load's
  leaked payload reachable at: user:shared-lit/0
  environment 0
Fork: a transaction on one fork is invisible to every other fork
  leaked payload reachable at: user:shared-lit/0
  transaction 1 moved fork 0
```
Same Reader re-sealing its parse: zero witnesses. The only difference between the arms is the seal bit. Weakened by swapping the transaction order (observer loads second): red, `0 witnesses`.

Rebase note: the first revision used one process-wide reader and produced NO witness once #601 was underneath — the parity channel (`f3538f1`) runs first and builds its cold environments through the same `NewEnv`, so its cold tx1 sorted the single shared parse before the sweep's fork 0 ever loaded it. Scoping the reader per template restored the shape the control needs, and the parity channel now reports it too.

**(b) Per-env COPY instead of the shared value** — `TestLoadCacheTopology_PrivateCopyControl`. A `LoadCache` cannot copy an opaque entry, so the nearest buildable shape is a private cache per fork. Every parity assertion passes, `CheckTransactions` with the private-cache fork walker reports nothing, and the identity assertion is the ONLY thing that notices — stated in the test as the point: over-copying is not a leak, and the pointer-identity assertion is the one assertion in this file that distinguishes sharing from copying. Its inverse (sequential identity arm over private caches): `the forks hold 6 distinct nodes under shared-lit, want 0` — red.

**(c) A fork with a PRIVATE cache diverging from the template's** — `TestLoadCacheTopology_MixedWarmth`. Shared cache pre-warmed by a throw-away env (template hits everything); odd forks carry a fresh private cache (every load a miss), even forks hit. All forks match their cold reference; `CheckTransactions` with the mixed fork walker reports nothing. Weakened by dropping the pre-warm: `stored 4 new entries over a fully warm cache` — red.

Other assertions, each red on weakening: sealed check (`shared-lit is not sealed` ×6 when the read is replaced by a `Copy()`); refusal (`got list '(1 2 3), want list '('refused 'modify-literal-error)` ×6 when the sort targets `(copy shared-lit)`); fork-vs-cold parity (`reachable state differs from the cold run ... at user:counter want int(2) got int(1)` when the cold env runs the neighbouring tx); store counting (`loads=36 hits=0 stores=36` when `Store` drops entries); hit count (`served 10 hits` when every fork loads `late.lisp`); template-never-binds-shared-lit (`the template binds shared-lit ('(3 1 2))` when the program loads it); cache refusal of the annotated parse (`the cache stored the annotated parse (1 new entries)` when the Reader stops annotating).

## Findings

**No leak.** With the real cache installed, no fork observed another fork's or the template's state, sequentially or concurrently, through hit or miss paths, under `CheckTransactions`, `RunForkCheck` or `CheckParity`.

**1. Native annotation on a sealed header — was #600 gap 1, now closed and pinned as a control.** A Reader that annotates a quoted-list `LSExpr` node through `LVal.Native` ("the ONLY exported per-node slot an embedder's Reader has for annotation", `newTextLoaderWalk`):
- The cache REFUSES the parse (`newProgramForCache`'s strict walk), so "cached once and forked N times" is impossible through this hook — the load runs uncached (asserted: no store).
- The annotation still reaches every fork BY REFERENCE, through Fork's sealed-node sharing; its `NativeCloner` is not consulted because the sealed node is shared wholesale (asserted: same `*LVal`, same payload pointer, 4 forks).
- `CheckTransactions` with `ExpectNoSharedNatives` now REPORTS it: one witness per environment pair (template + 4 forks = 10, asserted exactly), each at `user:lit`, `*elpstest_test.annotation ... [declares NativeCloner and is shared anyway]`, and no witness of any other kind. History: on 74e4ac8 `reachableNatives` keyed on the `LNative` header and saw nothing, so this test was born as the pinned opposite (`...GapStillOpen`, required zero native witnesses, measured red when the oracle was widened by hand). Commit `aa0dbe4` on #599 keyed "has a payload" on the payload; once it sat under this branch the pinned form went red with exactly this witness, as designed, and the arm was converted to `TestLoadCacheTopology_NativeAnnotationIsReported`. The rename is this PR's change, so this PR owns every reference to the name: the three comments beneath it that cited the old name (`fingerprint.go`, `aliasguard_payloadkey_test.go`, `aliasguard_isolation.go`) are updated here, comment-only, to the new name and the past-tense description.

**2. Concurrent first-touch is not single-flight (contract, not a bug).** When six forks first-touch `shared.lisp` at once, two or more can miss together, each parse and each `Store` (last write wins) — the `LoadCache` contract promises exactly this ("elps makes no assumption that a Stored entry is later Loadable"). Measured on 74e4ac8 over 60 runs of the concurrent arm: 46 stored every entry once; `shared.lisp` stored twice in 9 runs and three times in 2; `late.lisp` twice in 5 and three times in 2. Parity held on every run. The concurrent arm asserts the exact contract: distinct `shared-lit` nodes across forks == stores of `shared.lisp`, `loads == hits + stores`, `stores == 10 + double misses`. Worth knowing for `FuzzLoadCacheMultiEnv`'s `stores == 1` assertion, which holds only because that target runs its environments sequentially.

## Test plan

Per the standing rule, the delivery suites (`go test ./...`, `-tags elpscheck`, `-race ./elpstest/...`, `make static-checks`, `make elpsvet`) run on this PR's CI, not locally. Run locally on `65de8aa` (clean tree, `set -o pipefail`):

```
go build ./...                                                          ok
go vet ./elpstest/                                                      ok
gofmt -l <the five touched files>                                       (empty)
go test ./elpstest/ -run 'TestLoadCacheTopology_NativeAnnotationIsReported|TestCensusSeesAnAnnotationOnASealedNode' -count=1   PASS
```

Earlier heads, for the record: `886e3cf` (on c684e7b) passed every `TestLoadCacheTopology*` test locally; `4ef6a37` (on 74e4ac8) passed all five delivery gates locally (`go test ./...` 64s, `-tags elpscheck` 141s, `-race ./elpstest/...` 24s, `make static-checks` 31s, `make elpsvet` 6s).

**Fuzz campaigns on the chain.** `workflow_dispatch` runs of `fuzz.yml` with `fuzztime=10m` per target (11 shards, every registered target):
- First campaign, runs [33941632540](https://github.com/luthersystems/elps/actions/runs/33941632540) (this branch @ `590300e`) and [33941631011](https://github.com/luthersystems/elps/actions/runs/33941631011) (#604 @ `6f80732`): two crashers. `FuzzSequenceOps/7192dc3ce712e22c` bisected to #602 (an empty vector carve-out built a sealed holder), fixed there in `c2682c6` with the seed committed; `FuzzLSPSession/36ec6d2dd72dfd1f` reproduces on `main` (pre-existing, filed as #607).
- Second campaign, run [33950054036](https://github.com/luthersystems/elps/actions/runs/33950054036) on `b076a2f` (this head, on the fixed chain): clean on every target except `FuzzLSPSession/ae756f8085a54f2b`, the same pre-existing #607 class (reproduces on `main`, seed recorded on the issue). No crasher in the stack's own code.

### DIFF
--- a/elpstest/aliasguard_isolation.go
+++ b/elpstest/aliasguard_isolation.go
@@ -788,11 +788,12 @@ func reachableNatives(env *lisp.LEnv) map[any]string {
 		// LBytes holds a *[]byte there, LSortMap a *MapData, and an embedder
 		// can annotate an ordinary node. Keying on `v.Type == LNative` missed
 		// all of it -- including the case measured in #603
-		// (TestLoadCacheTopology_NativeAnnotationGapStillOpen), where a
-		// Reader's annotation on a SEALED node reaches every fork by
-		// reference because a sealed value is shared outright, before the
-		// native policy runs, so its NativeCloner is never consulted. Nothing
-		// here saw that, and this census is the surface that should have.
+		// (TestLoadCacheTopology_NativeAnnotationIsReported, which pinned
+		// the gap as open until this change closed it), where a Reader's
+		// annotation on a SEALED node reaches every fork by reference
+		// because a sealed value is shared outright, before the native
+		// policy runs, so its NativeCloner is never consulted. Nothing here
+		// saw that, and this census is the surface that should have.
 		//
 		// A cell-view link is excluded: it is a reference to a root, not a
 		// payload (isCellViewLink, elpstest/fingerprint.go).
--- a/elpstest/aliasguard_payloadkey_test.go
+++ b/elpstest/aliasguard_payloadkey_test.go
@@ -104,8 +104,9 @@ func TestCensusSeesPayloadsOnNonNativeHeaders(t *testing.T) {
 // so an embedder annotation on one reaches every fork by reference and its
 // NativeCloner is never consulted. Nothing in this package could see that:
 // the annotation sits on an LSExpr, and the census only looked under
-// LNative. It is now visible, which is what makes
-// TestLoadCacheTopology_NativeAnnotationGapStillOpen (#603) able to close.
+// LNative. It is now visible, which is what let #603's pinned-gap test
+// close into TestLoadCacheTopology_NativeAnnotationIsReported, the control
+// that now requires the census to report exactly that annotation.
 func TestCensusSeesAnAnnotationOnASealedNode(t *testing.T) {
 	t.Parallel()
 	env := payloadKeyEnv(t)
--- a/elpstest/fingerprint.go
+++ b/elpstest/fingerprint.go
@@ -471,11 +471,11 @@ func isCellViewLink(v *lisp.LVal) bool {
 // shared storage -- LBytes holds a *[]byte there, LSortMap a *MapData, LFun a
 // *funData -- and an EMBEDDER can put a payload on an ordinary node too. A
 // Reader that annotates an LSExpr is the measured case (#603,
-// TestLoadCacheTopology_NativeAnnotationGapStillOpen): the LoadCache refuses
-// to cache such a parse, but Fork still shares the annotation with every fork
-// by reference, because a SEALED node is shared outright before the native
-// policy runs -- so its NativeCloner is never consulted, and nothing here saw
-// it.
+// TestLoadCacheTopology_NativeAnnotationIsReported, which pinned the gap as
+// open until this closed it): the LoadCache refuses to cache such a parse,
+// but Fork still shares the annotation with every fork by reference, because
+// a SEALED node is shared outright before the native policy runs -- so its
+// NativeCloner is never consulted, and nothing here saw it.
 //
 // The types whose arms own Native are excluded because those arms already
 // encode its identity (map#, bytes#, native#) and doing it twice would double
--- a/elpstest/loadcache_topology_test.go
+++ b/elpstest/loadcache_topology_test.go
@@ -1,0 +1,977 @@
+// Copyright © 2026 The ELPS authors
+
+// The deployed topology, run through the isolation and parity oracles.
+//
+// substrate's production shape is ONE preheated template, N forks (one per
+// transaction), and ONE warm lisp.LoadCache shared by all of them: the cache
+// travels across Fork (lisp/fork.go, the LoadCache line of the forked
+// Runtime), and docs/fork.md states the sharing is safe because every entry
+// is immutable and sealed throughout.  The mechanism has been verified by
+// hand, and every load-cache test on commit 74e4ac8 — FuzzLoadCacheMultiEnv,
+// FuzzLoadCacheHostileReader, the custody and sharing tests — exercises it
+// over INDEPENDENT environments.  No TransactionCheck installed a cache, so
+// the shape an embedder actually ships had never been run through an
+// isolation oracle.  These tests close that (issue #600, gap 6).
+//
+// The property, in the owner's words: for two VMs instantiated from the
+// same source, ELPS code should work identically to two forks from the same
+// template.  Fork is a performance optimisation.  Parity with cold VMs is
+// the spec; isolation and fidelity follow from it.
+//
+// What is asserted, and by which test:
+//
+//   - TestLoadCacheTopology_Isolation: elpstest.CheckTransactions over a
+//     template whose cache is shared by the template, its forks and the
+//     concurrent arm's second template.  Every isolation property, with the
+//     cache installed, sequentially and concurrently (the concurrent arm is
+//     the -race gate).
+//
+//   - TestLoadCacheTopology_Parity: elpstest.RunForkCheck, whose cold arm
+//     builds a fresh environment per transaction.  With a fresh cache per
+//     NewEnv call, the fork arm HITS the entries the template and earlier
+//     forks minted while the cold arm MISSES in its own cache — parity
+//     across cache warmth, on the oracle that already exists on this base.
+//
+//   - TestLoadCacheTopology_SharedEntryIdentity: the documented design,
+//     asserted explicitly.  Two forks that hit one key hold the SAME *LVal;
+//     it is sealed; the in-place mutation is refused identically on every
+//     fork; and each fork's fingerprint equals a cold environment's that
+//     used its own fresh cache — from lisp's point of view a shared sealed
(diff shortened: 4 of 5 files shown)


sam-at-luther:
Note on this PR's red fuzz shard, since the session that was driving this stack is no longer running.

Failing check: Fuzz shard 11/11 on `b076a2f` (run 33950054036), target `FuzzLSPSession`, crasher `ae756f8085a54f2b`. It is not this PR's: this stack touches `lisp/` and `elpstest/`, the crasher is in `lsp/`, and it reproduces on plain `main` (a32ada7) at the 30s watchdog. It is a second instance of the class #607 describes, found independently of `36ec6d2dd72dfd1f`.

Measured against the fix in #609 (`fix/lsp-hang-fuzz-36ec6d2d`): `ae756f8085a54f2b` FAILS on `main` at 30.00s and PASSES on that branch in 0.446s with no change to the fix, so the size gate closes this instance too. I have committed it there as a permanent seed alongside the first one. Once #609 merges into `main` and this stack rebases, the shard goes green here.

No re-run: the failure is deterministic and reproduced locally on `main`.

---

sam-at-luther:

The replacement exercises real shared/private/mixed LoadCache topologies, late loads, exact hit controls and one/two publication hops in [TestTemplateLoadCacheTopology and TestTemplateLoadCacheForkCheck](https://github.com/luthersystems/elps/blob/v1.61.0/elpstest/template_loadcache_test.go). This resolves the deployed-cache portion of [#600](https://github.com/luthersystems/elps/issues/600), not that issue's unrelated all-walker coverage gaps.

PR #604, opened 2026-09-05
Memoise payloads per map, bytes and native in (*LVal).Copy

sam-at-luther:
Stacked on #599 (base `claude/elps-perf-optimization-jcgfd2-alias-guard`, rebased onto its head febe667). Draft. Seven commits: b18126c (the fix), 02b6293 (the `copy`-builtin pin), ee6b054 (the known-defect rows closed and the pins inverted), 47e6c80 (macro-expansion metadata dropped, as Fork and detach do), 0e5c65f (#599's per-walker measurement flipped for Copy), 3592b6b (sort comparator arguments passed by reference — the benchmark-gate remedy, below), 50cdce0 (the exact allocation pin moved under the precedent's build constraint).

## The defect (#600 gap 5)

`(*LVal).Copy` (lisp/lisp.go) was a fifth value-rebuilding walker in neither registry, and it rebuilt per *header*:

- **De-aliased maps.** Two headers over one `*MapData` — what `(quasiquote (unquote a))` produces — came out as two maps (`copyMapData` allocates fresh per call, no memo). A write through one name was invisible through the other on the copy. The #576/#585 class, fixed in the fork walker and the detacher (742598b) and live here the whole time both guards were green.
- **Shared bytes.** `*cp = *v` kept the `*[]byte`, so copy and source wrote one buffer (#551).
- **Shared map values.** `copyMapData`'s contract shares value pointers, so a container held as a map value was reachable from the copy as the source's own header.
- **Not cycle-safe.** Copy recursed without bound on a self-containing value (lisp/package_admit.go classified cycles before copying for that reason).
- **Copied macro-expansion metadata** whose shared context points at the source tree's nodes (below).

Copy sits on lisp.TextLoader's isolation boundary (`loader.go:119`, "hand each evaluation a PRIVATE tree"), so each of these is one load writing another's data.

### Repros (cold environment as the reference), before → after

Map — program `(set 'a (sorted-map "k" 1)) (set 'b (quasiquote (unquote a))) (set 'pair (list a b))`, then `pair` replaced by `pair.Copy()`, probe `(assoc! (first pair) "y" 7) (list (get (second pair) "y") (get (first pair) "y"))`:
- cold: `'(7 7)` · old Copy: `'(() 7)` · now: `'(7 7)`, source untouched. (`TestCopyMemoisesMapPayloadAcrossHeaders`)

Bytes — program `(set 'a (to-bytes "abc")) (set 'b (quasiquote (unquote a))) (set 'pair (list a b))`, probe `(append! (first pair) 7) (list (length (second pair)) (length (first pair)))`:
- cold: `'(4 4)` · old Copy: `'(4 4)` *for the wrong reason* — the source buffer moved to length 4 · now: `'(4 4)` with the source still at 3. (`TestCopyMemoisesBytesPayloadAcrossHeaders`)

## The fix

`lisp/copier.go`: the detacher's memo pattern applied verbatim — one memo per payload kind (`*MapData`, `*[]byte`, a NativeCloner payload held by pointer), each seeded before its contents are walked so a map reaching itself through a second header closes onto its own copy, plus the header memo so a subtree reachable twice is copied once and a cycle terminates. A NativeCloner is cloned exactly as the detacher clones it in copy mode (once per pointer payload, no runtime-affinity check). Deliberately unchanged: the seal is cleared on every node; closures keep their environment by reference; a list's cells backing array is not preserved (`TestCopyDoesNotPreserveBackingArraySharing`, untouched and passing); an LError's `*CallStack` is shared (immutable by construction: `CallStack.Copy` allocates exact-length Frames at every capture site and nothing writes a captured stack; `TestCopyAliasesCallStackAcrossHeaders` still passes); a native that is not a NativeCloner is shared. *(An LArray's backing was in this list as "shared (`TestCopyAliasesArrayBacking`)" until the fuzzer showed why it could not be — see "FuzzAliasGuard finding" at the end: arrays are now walked like lists.)*

**Macro-expansion metadata: dropped (choice (a)).** `TestMacroExpansionBehaviourPerWalker` on #599 measured the walkers apart: Fork and detach drop the record, Copy copied it. The record's *shared* per-expansion context (`macroExpansionContext.Args`) records the macro call's unevaluated arguments as pointers to the source tree's nodes — on that fixture 1 of 3 unsealed — so a copy that kept it was a private tree with a back-pointer into the tree it was copied from. Debugger-only exposure (the record exists only under an attached `Runtime.Debugger`), but exactly this walker's class. Why (a) and not remap (b): `Args` lives on the context that `macroExpansionInfo.Copy` deliberately kept shared across every node of one expansion, so remapping means a per-walk context memo (a new memo-shaped field needing a registry row) plus a pass after the walk — well over ten lines for a record the other two walkers already drop. What a debugger user loses, by reading the `MacroExpansion()` readers (`lisp/x/debugger/inspector.go`, `engine.go`, `dapserver/translate.go`, `dapserver/handler.go` — all read the *paused expression's* record): the record is stamped *during* evaluation and the load paths (TextLoader, Program) copy a tree *before* evaluating it, so nothing is lost there; a tree that was expanded, retained (`macroexpand-1`) and then copied loses the debugger's macro attribution on the copy — as it already does on every fork. The stepper's known #466 hazard (two nodes carrying one MacroID after a copy) now has no path, since a copy carries no ID. `macroExpansionInfo.Copy` had no other caller and is removed; its ID comment is corrected. The header memo does not survive the walk: the copier is a stack local of `Copy` and no closure captures it.

**#599's per-walker measurement, flipped here (0e5c65f).** CI tests this PR merged with its base, so `TestMacroExpansionBehaviourPerWalker`'s Copy arm — written to assert COPIES and report the seal state without a verdict — went red on the drop, as designed. Its Copy arm now asserts DROPS, citing `TestCopyDropsMacroExpansionMetadata`; the Fork and detach arms are untouched; the seal-state report moves onto the *source* and becomes the anti-vacuity for the Copy arm (the fixture must still record an unsealed node — it records 3 args, 1 not sealed). The walker-contract table in `FingerprintOptions.MacroExpansion`'s comment now reads "(*LVal).Copy DROPS it (#604)". #599's own row stays "copies — #604 changes it"; this flip is the diff that makes both true at once.

**Registry.** The walk is registered as `copier` in `lisp/walkers.go`, with the same memo set as the fork and detach walkers, so the drift guard holds it to that set (`TestRebuildingWalkersMemoiseTheSamePayloadKinds` now compares it as a clean walker). The base's `(*LVal).Copy` registry row (Rebuilds, no memos) is removed: it described the walk this PR replaces, and with no allowlist row an unmemoised Rebuilds row is rightly a hard problem. The LVal struct-copy scan resolves walkers by enclosing function (`(*copier).copy`), and `(*LVal).Copy` itself no longer contains a struct copy, so nothing keyed on the old name.

**Known-defect allowlist — the intended lifecycle.** The base carries `knownDefectiveWalkers` with two rows for Copy (the `*MapData` and `*[]byte` defects), each pinned by a test asserting the defect is *still live*. Both pins went red on this fix — the designed signal. This PR deletes both rows (the list is now empty; its comment records why) and inverts the pins into permanent **negative controls under the same names** — `TestCopyDeAliasesMapPayloadAcrossHeaders` (two headers over one map copy to two headers over one map; a write through one is seen through the other and not in the source) and `TestCopySharesBytesPayloadAcrossHeaders` (a copy owns its bytes; two headers over one buffer copy to two headers over one copied buffer) — with failure messages that say what a regression means. The bytes control's second-header fixture gets the same struct-copy exemption row the map fixture already had. That is what a shrink-only allowlist row is for: pinned while live, deleted the day the pin goes red, the pin kept as the control.

**Contract change, accepted in direction by the coordinator:** a sorted map's *values* are now walked like a list's cells. Measured reason: the alias guard's probe writes a byte through the copy's map entry and asserts the source did not move; with value-pointer sharing the copy's entry *is* the source's LBytes header, so no memo could make `CheckWalker` green. `copyMapData` itself (assoc/dissoc's path) keeps sharing values and is unchanged; its pins in `copy_sortedmap_test.go` pass unchanged. `TestCopyAliasesBytesAndMapValues`, which pinned the defect as the detach tests' control, becomes `TestCopyRebuildsBytesAndMapValues`.

**Cycle safety is a behaviour change:** Copy used to overflow the stack on a self-containing value; it now terminates with the cycle closing onto the copy. Pinned by `TestCopyTerminatesOnACycle` (`cyclic.Cells = append(cyclic.Cells, cyclic)`; `cp.Cells[1] == cp`) and, across the memo spill, `TestCopyMemoSpillsPastTheInlineArray` (62-element list with a shared subtree at both ends and a self-reference).

**Cost.** Copy is a per-call primitive on small values (insert-sorted copies leaf ints inside a binary search; `assert` copies its test per evaluation; lambda creation copies formals), and `TestVectorBuiltinAllocations` pins those counts as equalities — an earlier draft with a heap map per walk failed it at 49 vs 31. So the header memo is a 16-entry inline array spilling to a map only past it, the copier is stack-resident (`go build -gcflags=-m`: no escape; map entries are walked by loop rather than through `sortedmap.clone`'s callback, which would capture it), and a leaf costs its header alone. The allocation gate is unchanged; `TestCopyLeafAllocatesLikeAStructCopy` (1 alloc) and `TestCopySmallWalkDoesNotAllocateAMemo` (5 for a 3-element list) pin Copy's own boundary.

## Every Go caller of `(*LVal).Copy` (all packages, other receivers' Copy excluded)

| site | copies | wants rebuilt map values? | path |
|---|---|---|---|
| lisp/loader.go:119 (`TextLoader`) | a parsed expression per evaluation | yes — "a PRIVATE tree" is the purpose; parse trees hold no maps, so the cost is the header memo per node | **load** |
| lisp/program.go:286 | a parsed expression before sealing | yes; no maps in parse trees | load |
| lisp/package_admit.go:184 | an admitted package value (any data) | yes — the comment says "Copy severs every alias the caller retained"; a map's values were the one alias it did not sever | admit (once) |
| lisp/op.go:187 (`assert`) | the test expression per evaluation | n/a (expression) | request |
| lisp/op.go:841 (`handler-bind`) | an error value's cells before invoking the handler | yes — the handler gets a private copy; error data holding maps is now fully private | error path |
| lisp/builtins.go:1635–1640 (`lvalByFun.Less`, stable-sort) | *(before 3592b6b)* the two **elements** compared, per comparison | semantic: yes (the comparator cannot reach the list's own storage); **cost**: a list of maps now pays a walk of each map's values per comparison instead of a structural clone sharing them — scalar values are one header either way | **request** — `BenchmarkStableSortMaps` |
| lisp/builtins.go:1714–1724 (`insert-sorted`) | *(before 3592b6b)* the item and the probed element, per probe | same as above | **request** — `BenchmarkInsertSortedMaps` |
| lisp/builtins.go:2038 (`make-sequence`) | each number | n/a (leaf) | request |
| lisp/builtins.go:460, :1175; lisp/defformals.go:94, :98; lisp/env.go:853; lisp/lisplib/libschema/libschema.go:588 | a formals list (symbols) | n/a (leaves) | registration / lambda creation |
| lisp/lisp.go:739 (`Array`), :1285 (`ArrayDims`) | an array's dims list (ints) | n/a | request |
| elpstest/lisptest.go:522 | benchmark-runner expressions | n/a | test harness |
| cmd/elpsvet/testdata/... | analyzer fixtures | n/a | not production |

No caller relied on shared map-value headers for O(1) behaviour; the only cost-bearing sites were the two sort comparators, which 3592b6b removes from this list. `lisp/detach.go:351` is `CallStack.Copy` and excluded.

### Why the sort comparators copied their elements, and whether a shallower copy would do

Both copies date from the initial import (0281c7f, 2018-11-22), with only the two-line comment "Functions are always copied when being invoked. But the arguments are not copied in general." and no issue or bug behind them. The mechanism explains the guarantee: unlike `map` and `foldl`, which hand each element to the user's function uncopied through `env.FunCall(f, fargs)`, `lvalByFun.Less` and `insert-sorted`'s probe build a fresh S-expression `(fun a b)` around the *raw element nodes* and `env.Eval` it. Under the old `Copy` that guarantee was already "cells private, map structure private, map *values* shared", so a comparator writing through a map-valued element reached the source; the deeper walk closes that, at the cost this PR's benchmark rows measure. Would a shallower copy preserve the guarantee? A header-only copy would not protect a payload a comparator writes — so no, not as a substitute for the walk. The right follow-on is the one `map`/`foldl` already take: hand the comparator the elements themselves and copy nothing, which matches the rest of the library's contract (a user function sees the element, not a private copy) and removes the per-comparison cost outright. Taken in 3592b6b, below.

## The lisp `copy` builtin is unchanged

By reading: `builtinCopy → (*LVal).deepCopy → detacher{shareOpaque: true}` (lisp/copy.go:84–112). It never routes through `(*LVal).Copy`. `TestCopyBuiltinIsUnchangedByTheCopier` pins the lisp-observable result of `copy` on a map holding a bytes value and a nested list, and on two headers over one bytes value; the identical test file passes on 74e4ac8 and on this head (run with `-run` on both). `TestCopyDoesNotPreserveBackingArraySharing` and every pin in `copy_sortedmap_test.go` are untouched and pass.

## Controls

Weakening battery (each measured once on the fix commit's code, `go vet` clean, reverted):
- delete the `*MapData` memo → `TestCopyMemoisesMapPayloadAcrossHeaders` + `TestCopyMeetsTheAliasGuard` red (and, on this head, `TestCopyDeAliasesMapPayloadAcrossHeaders`)
- delete the `*[]byte` memo → `TestCopyMemoisesBytesPayloadAcrossHeaders` + `TestCopyMeetsTheAliasGuard` red (and `TestCopySharesBytesPayloadAcrossHeaders`)
- delete the native memo → `TestCopyClonesANativeClonerOncePerPayload` red
- delete the header memo → `TestCopyMeetsTheAliasGuard` red. **Measured once, 36 minutes, not to be repeated:** without the header memo the cyclic fixture recurses until Go's 1 GB stack limit — which is also the measurement that Copy is now cycle-safe where it was not.
- stop walking map values → `TestCopyRebuildsAContainerHeldAsAMapValue` + `TestCopyMeetsTheAliasGuard` red

Macro-expansion control: `TestCopyDropsMacroExpansionMetadata` (lisp/copier_test.go) — #599's fixture (a call form built at runtime so its nodes are unsealed, expanded by a builtin macro under a dormant debugger, retained by `macroexpand-1`); anti-vacuity first (the source must carry metadata recording an unsealed node of its own tree), then no node reachable from the copy reports `MacroExpansion()`, and — the walker-independent form — no node reachable through the copy's metadata is in the source's reachable set. Before 47e6c80 the second assertion failed on the unsealed args. The three #466 pins in `copy_macroexp_test.go` now assert the drop: `TestCopyDoesNotAliasMacroExpansionInfo` and `TestCopyKeepsSharingTheMacroExpansionContext` keep their names (the original's record and shared context are untouched by a copy), `TestCopyDuplicatesTheMacroExpansionID` becomes `TestCopyCarriesNoMacroExpansionID` (a name asserting the retired decision would mislead; nothing outside its file cited it).

`TestCopyMeetsTheAliasGuard` drives Copy through the alias guard's `CheckWalker` with a test-local `elpstest.Walker`.

## Benchmark

See the **Benchmark Comparison** job on this PR (benchgate, 15 %/5 %, PR vs base). `lisp/copier_bench_test.go` (`BenchmarkStableSortMaps`, `BenchmarkInsertSortedMaps` — the sort-of-maps shape with the deeper walk) is on the base, so the gate diffs those rows against a base row. On 0e5c65f the gate was red on exactly those rows (numbers in the section below); 3592b6b is the remedy, and the gate on the current head is the verdict. Existing Copy-path benchmarks also in the diff: `BenchmarkTextLoaderLoad`, `BenchmarkLoadIntoEnv`, `BenchmarkAliasConcatCopy`, `BenchmarkAliasStableSort`. On 0e5c65f the loader rows moved below the gate but not by nothing (`LoadIntoEnv/library-78KiB/reader-cache-copy` B/op +8.6 %, sec/op +14.5 %; `TextLoaderLoad` B/op +8.9 %) — that is the header memo on a large program tree, the price of cycle safety and aliasing preservation on the load path; recorded here so it is not rediscovered. No local benchmark was run (repo rule).

## Test plan (local, per the rule; everything else is CI)

- `go build ./...`, `go vet ./lisp/` / `./elpstest/`, gofmt on touched files — clean on 0e5c65f
- targeted runs with the exit status checked: on 0e5c65f `TestMacroExpansionBehaviourPerWalker` (elpstest) and `TestCopyDropsMacroExpansionMetadata` (lisp) — ok; on 47e6c80's tree `go test -run 'TestCopy|TestVectorBuiltinAllocations|TestEveryLValStructCopyIsInAWalkerOrAllowlisted|TestWalkerRegistryMatchesTheSource|TestRebuildingWalkersMemoiseTheSamePayloadKinds' ./lisp/` — ok; on ee6b054's tree the two inverted pins plus the seven registry/scan checks and the CallStack pin — ok
- `make elpsvet` and `golangci-lint run ./lisp/` — clean on the fix commit's tree
- `TestCopyBuiltinIsUnchangedByTheCopier` on this branch and on 74e4ac8 — ok
- the weakening battery above

## Comparator arguments are now passed by reference (3592b6b)

The benchmark gate on 0e5c65f was red on exactly the two request-path callers the table above flags — StableSortMaps sec/op +76 %, B/op +101 %, allocs/op +85 %; InsertSortedMaps sec/op +27 %, B/op +13 %, allocs/op +45 %; AliasStableSort sec/op +17 %; AliasStableSortSliceView +15 % below gate, same cause. The "follow-on" of the section above is taken here, as one commit: `lvalByFun.Less` (both branches) and `insert-sorted`'s search probe no longer `Copy` their arguments per comparison; they pass the element headers. The call stays an evaluated S-expression rather than a `FunCall`, so an element that is an unquoted symbol is evaluated exactly as before — only the copy is gone. `search-sorted` never copied (it passes a fresh index). An earlier draft of this body claimed evaluation "writes on the nodes it evaluates"; that does not hold for this path: `eval` returns a quoted node as-is and looks a symbol up; the writes it referred to land on macro *expansions* and a call's fresh argument list, not on an argument node, which is why `map`/`foldl` have always been safe passing elements through `FunCall`.

**The semantic decision, stated plainly.** A stable-sort predicate or key function, and an insert-sorted predicate, now receive the list's own elements — the contract `map`, `foldl`, `select` and every other higher-order builtin already have. The only programs that behave differently are comparators (or key functions) that **mutate** their argument (`assoc!`, `append!`, `aset` on it) or **retain** it past the call: the write or the reference now reaches the element in the list — **or, if that element is a sealed program literal (`(list '(2 1) '(1 0))`, or a node reached from a sealed tree through `first`/`rest`), the write raises `modify-literal-error` and the whole `stable-sort`/`insert-sorted` call aborts.** That second consequence is a behaviour change of its own (review round 1, below): the old per-comparison Copy cleared the seal, so a mutating predicate wrote a throwaway copy and the sort completed; measured febe667 → head, `(set 'xs (list '(2 1) '(1 0))) (stable-sort (lambda (a b) (stable-sort < a) (< (first a) (first b))) xs)` went from `'('(1 0) '(2 1))` to `modify-literal-error`, and likewise for the key-function path and for `insert-sorted` with a sealed item. It is an error rather than a silent write for the #378 reason — a silent copy masks code that believed it owned the value — and the remedy is the one #378 gives, `(copy ...)` the literal. The copy was never a complete isolation either — `Copy` shared map values by pointer and shared bytes until this PR — so a comparator writing through a map-valued element already reached the source. Precedent: `builtinSortStable`'s sealed-literal branch replaced a silent copy-on-write with `modify-literal-error` after a census found zero non-test programs on that path (issue #378). Documented in the `stable-sort` section of `docs/func.md`.

**Census (call heads of `stable-sort`, `insert-sorted`, `search-sorted`; `sort` was searched too but there is no such builtin — the only `sort` in builtins.go is the Go import — so its column is trivially zero and dropped), classified by comparator / key function — (i) builtin or pure lambda over its arguments, (ii) mutates an argument, (iii) stores an argument outside the call.** Counts and shapes only; the customer corpora were read in place.

| corpus | call sites | (i) | (ii) | (iii) |
|---|---|---|---|---|
| this repo, `.lisp` files | 0 | – | – | – |
| this repo, lisp strings in Go (tests, fuzz seeds, benchmarks, docs) | every comparator is a builtin (`<`, `>`, `'>`) and every key function a builtin (`identity`, `-`, `first`) or a pure lambda returning a map field | all but one | 1 — `TestSortedMapCopyViaSort`, the test that asserted the copy (updated, below) | 0 |
| customer phylum, pre-production snapshot | 9 `stable-sort` (+1 test-file call through a helper wrapping one of them); 0 `insert-sorted`, 0 `search-sorted` | 9 | 0 | 0 |
| customer phylum, production snapshot | 9 `stable-sort`; 0 / 0 | 9 | 0 | 0 |

Shapes of the nine: one `string>` over a key list; one `<` with a key function that reads a global lookup table by the element; seven pure lambdas comparing one or two string/int fields of each argument (one dispatches to a comparison function passed in by the caller). A word-match on `sort` reports 45 lines per snapshot, but the extra hits are identifiers containing the word (a bank sort-code field, `sort-func` bindings) and comments, not calls. Class (ii) and (iii) are zero in both customer snapshots, so no real program changes behaviour.

**Tests.** `TestSortComparatorArgumentsAreTheElements` (lisp/sort_test.go) pins the new contract for all four sites — stable-sort predicate, stable-sort key function, insert-sorted predicate, insert-sorted key function — by writing through the argument and asserting the element in the list carries the write; every assertion fails on 0e5c65f. `TestSortedMapCopyViaSort` (lisp/map_test.go) asserted the behaviour this commit changes and is **updated, not deleted**: it is now `TestSortedMapMutationViaSortReachesTheElements`, asserting the write reaches both elements (both arguments are marked because sort.Stable's argument order is an implementation detail). `TestStableSortAllocationsDoNotDependOnMapWidth` (lisp/sort_alloc_test.go) sorts eight one-entry maps and eight nine-entry maps with a Go-implemented `<`-on-a-field predicate and asserts the per-call allocation counts are equal — under the per-comparison copy they differed by hundreds — with an anti-vacuity that copying the wide map costs more than the narrow one; it runs in every build. The exact count (43) is pinned separately in `TestStableSortAllocationCount` (lisp/sort_alloc_pin_test.go, 50cdce0) under the `!race && !elpscheck` constraint `TestVectorBuiltinAllocations` uses: CI's test-elpscheck step measured 52 on 3592b6b, because the checked build adds ownership bookkeeping to every eval — the same on both widths, so the property holds there and only the exact count moves. `TestVectorBuiltinAllocations`'s `insert-sorted` row moves 31 → 25 (three probes, two leaf copies each); the table comment records why. The `copier.go` cost comment no longer cites insert-sorted as a Copy caller.

**Local checks on 3592b6b** (nothing else; CI and the benchmark gate are the verifier): `gofmt -l` on the touched files; `go build ./...`; `go vet ./lisp/... ./elpstest/...`; `golangci-lint run ./lisp/...` (v2.6.2, 0 issues); with `set -o pipefail`, `go test -run 'TestStableSortAllocationsDoNotDependOnMapWidth|TestVectorBuiltinAllocations' ./lisp`, `go test -run 'TestSortComparatorArgumentsAreTheElements|TestSortedMapMutationViaSortReachesTheElements|TestSort$' ./lisp`, `go test -run 'TestSliceViewSharesElements|TestSliceCapacityAliasing|TestStableSortLiteralPristine|TestCopySortsInPlaceWithoutCopyOnWrite|TestSealedWriteRaisesCatchableCondition|TestUnsealedWritesKeepInPlaceSemantics|TestSealedWriteEmptyCarveOut|TestSingletonCoWContainerOps|TestSubstrate378ClassKilled|TestValueMutationDoesNotCorruptCachedAST|TestHigherOrderCallbacksAreAdversarial|TestHigherOrderSharingIsReachable|TestSequenceOpsAreDocumented|TestMacroexpandRestDoesNotLaunderSealedBacking|TestLoadCacheWriteThroughSharedTreeIsRefused|TestCopyMeetsTheAliasGuard|TestArray$' ./lisp`, and `go test -run 'TestRunBenchmarkGivesEachIterationAFreshAST|TestQuotedLiteralMutationIsRefused|TestForkCheck_SortedMapAliasAcrossHeaders|TestGuardDetectsDealiasedSortedMap|TestASortedMapEntryIsAProbeSite' ./elpstest` — all ok. On 50cdce0: gofmt, `go build ./...`, `go vet ./lisp/`, `golangci-lint run ./lisp/...` (0 issues), and the two allocation tests targeted in both the release build and `GOFLAGS=-tags=elpscheck` — ok.

## Review round 1 (comparator commits) — c1c929d

Adversarial review of 3592b6b + 50cdce0. Everything structural was verified true (eval writes nothing to an argument node; no elpscheck ownership trip across forks; unquoted/macro elements evaluate identically to febe667; insert-sorted already inserted the original item; the new tests fail on febe667 and pass on head; census 0/0 on the production snapshot). Findings and fixes, one commit (c1c929d, rebased onto 48a0cab's memo size hint, plain push):

1. **MEDIUM — unstated second behaviour change: a mutating predicate on a sealed literal element now raises.** Stated in the semantic-decision paragraph above and in `docs/func.md`'s `stable-sort` section (with the #378 policy as the reason it is an error and not a silent write). `TestSortComparatorArgumentsAreTheElements` gains three cases — the stable-sort predicate path (the review's repro), the key-function path, and `insert-sorted` with a sealed item — each naming the condition through `handler-bind` (so the assertion does not depend on an error column) and asserting the literals are as written afterwards. The suite fails on febe667 (the sealed cases sort silently there, the write absorbed by the copy) and passes on head; the commit message says so.
2. **LOW — three `docs/func.md` examples sorted quoted literals and raise since #378.** `(stable-sort > test)` and the key-function example now sort `(list ...)` values; the "stays rewritten for the life of the process" bullet now shows the refusal (`modify-literal-error`) the `(probe)` example actually raises, with `(copy lit)` / `(concat 'list lit)` as the remedy.
3. **LOW — `lvalByFun.Less` comment overstated the old copy's leakiness** ("never isolated the list anyway"). Reworded: the copy gave the predicate a private list or map structure — which is why the new test's first cases fail on febe667 — and only map values and bytes were shared; matches the "never a complete isolation" above.
4. **NIT — `TestStableSortAllocationCount`'s doc comment said AllocsPerRun's first call sorts the list;** the test sorts explicitly before measuring, and the sentence now says so.
5. **Info — the census `sort` row was trivially zero** (no such builtin); footnoted in the census header and dropped from the rows.

Local checks on c1c929d: `gofmt -l` on the touched files; `go build ./...`; `go vet ./lisp/... ./elpstest/...`; `golangci-lint run ./lisp/...` (0 issues); with `set -o pipefail`, `go test -run 'TestSortComparatorArgumentsAreTheElements|TestStableSortAllocationCount|TestStableSortAllocationsDoNotDependOnMapWidth' ./lisp` in the release build and `go test -tags=elpscheck -run 'TestSortComparatorArgumentsAreTheElements|TestStableSortAllocationsDoNotDependOnMapWidth' ./lisp` — ok, both before and after the rebase onto 48a0cab; the same `sort_test.go` dropped into a febe667 tree and run targeted — FAIL, as claimed.

## Copy is now a registered walker (e8dca7d)

Two commits on top of c1c929d, closing the gap the review of this PR's controls found: `(*LVal).Copy` was registered in `lisp/walkers.go`'s *memo* registry but not in `elpstest.Walkers()`, so the class-level oracle never drove it — `FuzzAliasGuard` (every generated alias graph, 11 CI shards), `TestAliasGuardOverEveryWalker`, `TestNoLiveWalkerOverAliasesEqualBuffers` and every other loop over the registry ran Fork, `copy`, Detach and the stamper and skipped Copy. Copy was covered by one fixed fixture through a test-local `Walker` — a second description of its contract that nothing held to the first.

**38ee5b8 — the fifth row.** `Walkers()` gains `LVal.Copy`: `WalkerCopy` over `(*LVal).Copy` (an LError result is a harness error, as the `copy` row treats an evaluation error), `ClosuresRefused` with no `Refusal` text (Copy shares closures by reference, as `copy` does — the same arm, and `checkCopy` skips the refusal probe for it exactly as it does for `copy`), `BackingRebuilt`, memo set `lisp.WalkerMemoKinds("copier")`. `lisp/copier_test.go`'s `copierWalker()` now selects that row by name (panics if the row is gone), so there is one description of the walker; `TestCopyMeetsTheAliasGuard` stays as the fixed-fixture pin and `copierHintedWalker` still derives from it. The registry doc names five walkers; the probe-cost table records it was measured before the fifth row.

**What the registry-driven checks found when Copy joined: nothing — on the committed seeds.** Run targeted with the row in place — `TestAliasGuardOverEveryWalker`, `TestAliasGuardNativePayloadAcrossWalkers`, `TestAliasGuardClosureState`, `TestAliasGuardSelfReferentialMap`, `TestEveryRegisteredWalkerDeclaresItsMemos`, the `FuzzAliasGuard` seed corpus (8 seeds, `-run`, no `-fuzz`), `TestNoLiveWalkerOverAliasesEqualBuffers`, `TestATruncatedForkSweepReportsOneFinding`, `TestMacroExpansionBehaviourPerWalker` (its Copy arm asserts DROPS directly on `src.Copy()`, not through the registry, so the new row neither double-covers nor contradicts it), `TestStampWalkerRunsItsMetadataBranch`, `TestTheDirectionMatrixMatchesTheProperties` (elpstest); `TestCopyMeetsTheAliasGuard`, `TestCopyWithHintMeetsTheAliasGuard`, `TestWalkerRegistryMatchesTheSource`, `TestRebuildingWalkersMemoiseTheSamePayloadKinds`, `TestEveryLValStructCopyIsInAWalkerOrAllowlisted` (lisp) — all ok, zero witnesses. No oracle was changed. *CI's fuzz campaign then found what the seeds did not — the next section.*

**e8dca7d — two mutation-proof rows against `lisp/copier.go`,** each caught by a NAMED property and possible only because Copy is now a row:

| row | mutation | must trip | must NOT trip |
|---|---|---|---|
| `604-copier-map-memo` | delete the `*MapData` memo lookup (two headers over one map copy to two maps) | `LVal.Copy: the copy has the same mutable payloads as the source` | the bytes row's needle; `Detach: …same mutable payloads…` (585); 576's and 600's needles |
| `604-copier-bytes-shared` | the copy keeps the source's `*[]byte` | `LVal.Copy: a write on one side is invisible on the other` | the map row's needle; 576's and 600's needles |

The bytes row pins a property no other row pins: sharing the source's buffer changes neither the fingerprint (per-graph ordinals) nor the site count, so only the probe write moving the *other* side sees it. The 600 row carries both new needles as must-NOT, per the manifest convention. Measured on each mutated tree (`git apply --check`, `go build ./...`, one targeted run over the fuzz seeds + `TestAliasGuardOverEveryWalker` + `TestAliasGuardSelfReferentialMap` + `TestNoLiveWalkerOverAliasesEqualBuffers`): map-memo emits its needle 4×, the bytes needle 0×; bytes emits its needle 10×, the map needle 0×; neither emits any other row's needle. `make mutation-proof` was not run locally (CI's Lint & Test runs it and the selftest); the Makefile comment's counts follow (eleven rows, ten property strings, one test).

Local checks on both commits: `gofmt -l` on the touched files; `go build ./...`; `go vet ./lisp/... ./elpstest/...`; `golangci-lint run ./elpstest/... ./lisp/...` (0 issues); `bash -n scripts/mutation-proof.sh`; the targeted runs above with `set -o pipefail`.

## FuzzAliasGuard finding (e1910d8)

Registering Copy paid off on the first CI run: fuzz shard 3 on e8dca7d found a crasher in 6 seconds (run 33939856295, `FuzzAliasGuard/f2a71578e30741e2`, artifact `fuzz-crashers-33939856295-shard3`). Fork, `copy` and Detach all passed the same input; Copy was the odd walker.

**The crasher** (the guard's repro, template only — no transactions):

```
(set 'v0 (sorted-map "k0" 1 "k1" 1))
(set 'v1 (vector v0 v0))
(assoc! v0 "self" v1)
(set 'probe (list v0 v1))
```

**Witnesses, verbatim (before — reproduced locally on e8dca7d from the reconstructed seed):**

```
LVal.Copy: the copy has the same values and the same sharing as the source
  fingerprint diverges (sharing is part of the encoding, so this fires on a de-aliased payload as well as on a changed value)
  at value/0/"self"/1/0 (token 31 of 38/56)
    want: hdr#1
    got:  hdr#9(sorted-map)
LVal.Copy: the copy has the same mutable payloads as the source
  sites that saw the write in the source: {probe -> cell 0 -> map entry "k0", probe -> cell 0 -> map entry "k1", probe -> cell 0 -> map entry "self"}
  sites that saw the write in the copy:   {probe -> cell 0 -> map entry "k0", probe -> cell 0 -> map entry "k1", probe -> cell 0 -> map entry "self", probe -> cell 0 -> map entry "self" -> cell 1 -> cell 0 -> map entry "k0", probe -> cell 0 -> map entry "self" -> cell 1 -> cell 0 -> map entry "k1", probe -> cell 0 -> map entry "self" -> cell 1 -> cell 0 -> map entry "self"}
  leaked payload reachable at: probe -> cell 0 -> map entry "self" -> cell 1 -> cell 0 -> map entry "k0"  (a payload the copy split: the source reaches it under one name)
  value: source has 3 probe sites, the copy has 6
```

**After (e1910d8):** `FuzzAliasGuard/seed#6` passes; zero witnesses on every seed and every registry-driven check.

**Diagnosis, confirmed.** The copier's `case LArray: return cp` kept the struct-copied `Cells` for an array outright ("reference semantics"), so the copy's vector held the SOURCE's dims and data-list headers and, through them, every element header: the copy reached the source's map through the vector — not hermetic (a write through the copy lands in the source) — and the fingerprint saw a header the copy's walk never memoised (`hdr#9`), while the registry row said `BackingRebuilt`. **Fork has no LArray arm** (`lisp/fork.go`): an array's two headers go through its default cells walk and the header memo. **The fix mirrors that exactly:** the arm is deleted, so the dims header and the data-list header go through `c.copy` and the header memo — two vector headers over one data list (`(quasiquote (unquote v))`) copy to two headers over ONE copied list, aliasing preserved, hermetic. `go build -gcflags=-m` unchanged in kind: nothing new escapes.

**Inverted pin.** `TestCopyAliasesArrayBacking` (the detach tests' control, asserting the sharing) becomes `TestCopyRebuildsArrayBacking`: disjoint `Cells` and data list, a write through the copy invisible in the source, and a `(quasiquote (unquote v))` pair of vector headers still sharing one copied list. Same lifecycle as the map/bytes pins. Every claim that Copy shares array backing is corrected: `copier.go` (the "deliberately unchanged" list, the seal comment, the default arm; a new "What changed once the walker was in the registry" section), `lisp.go`'s `Copy` doc, `detach.go`'s header, `copy.go`'s `deepCopy` doc, `parser/rdparser/fuzz_test.go`'s walk comment, `detach_test.go`.

**Seed.** No `elpstest/testdata/fuzz/` corpus is committed for this target, so the crasher is an `f.Add` seed in `aliasGuardSeeds`, reconstructed from the printed repro: `{1, 0, 3, 0, 0, 0, 1, 0, 0}` (nvars 2 → map, vector of v0 v0; cycle v0 → v1; trip 1; no transactions). It reproduces the two witnesses above verbatim on e8dca7d and passes on e1910d8, and runs on every CI run as `FuzzAliasGuard/seed#6`.

**Mutation-proof row `604-copier-array-shared`** (patch restores the arm). Measured on the mutated tree over the same targeted set as the other copier rows: it reddens exactly ONE test in `./elpstest` — `FuzzAliasGuard/seed#6` — emitting each of the two `LVal.Copy` property strings once; `TestAliasGuardOverEveryWalker`, `TestAliasGuardSelfReferentialMap`, `TestNoLiveWalkerOverAliasesEqualBuffers` stay green (no vectors in those graphs). The property that fires FIRST is the fingerprint one, `LVal.Copy: the copy has the same values and the same sharing as the source` (comparePair checks the fingerprint before the site count). It is not the map-memo row's needle, but the map-memo mutation emits it too (4× there), so it is not unique on its own; the manifest header's precedent for a shared string (585: pin what distinguishes the row's signature) applies, and the distinguishing signature here is the seed — neither the map nor the bytes mutation touches `seed#6` (0/0). The row therefore requires both: `LVal.Copy: the copy has the same values and the same sharing as the source|TEST:FuzzAliasGuard/seed#6`; must-NOT: the bytes row's needle, 585's, 576's, 600's. The 600 and bytes rows gain the fingerprint string as must-NOT (0 measured on each). `seed#6` is positional; a seed inserted before it renumbers it and the row goes NOT PROVEN loudly — the rule-1 failure mode, recorded in the notes. Counts follow: twelve rows, eleven property strings, one test alone (579), one property+test pair.

**Cost.** No Copy caller's observable behaviour changes beyond hermeticity: parse trees carry no LArray (vectors are built at runtime by `vector`), so the load paths (TextLoader, Program) copy nothing new; `handler-bind`'s error copy and `package_admit`'s admitted value now get private vectors, which is the contract. `TestVectorBuiltinAllocations` is unchanged in both builds (no row moved). CI's benchmark gate is the measurement for anything the fixed walk costs.

**Local checks on e1910d8** (`set -o pipefail`; release build and `-tags=elpscheck`): `go test -run 'TestCopyRebuildsArrayBacking|TestDetach|TestCopyRebuildsBytesAndMapValues|TestCopyMeetsTheAliasGuard|TestCopyWithHintMeetsTheAliasGuard|TestWalkerRegistryMatchesTheSource|TestRebuildingWalkersMemoiseTheSamePayloadKinds|TestEveryLValStructCopyIsInAWalkerOrAllowlisted|TestVectorBuiltinAllocations|TestCopyLeafAllocatesLikeAStructCopy|TestCopySmallWalkDoesNotAllocateAMemo|TestCopyMemoSpillsPastTheInlineArray|TestCopyDeAliasesMapPayloadAcrossHeaders|TestCopySharesBytesPayloadAcrossHeaders|TestCopyTerminatesOnACycle|TestArray' ./lisp` — ok (the `!elpscheck` allocation pins are absent from the checked build by their constraint); `go test -run 'FuzzAliasGuard|TestFuzzSeedsCoverTheHistoricalShapes|TestAliasGuardOverEveryWalker|TestAliasGuardNativePayloadAcrossWalkers|TestAliasGuardClosureState|TestAliasGuardSelfReferentialMap|TestNoLiveWalkerOverAliasesEqualBuffers|TestEveryRegisteredWalkerDeclaresItsMemos|TestMacroExpansionBehaviourPerWalker|TestTheGeneratorsPredictionsMatchTheRuntime' ./elpstest` — ok; `go build ./...`; `go vet ./lisp/... ./elpstest/... ./parser/...`; `gofmt -l` on the touched files; `golangci-lint run ./lisp/... ./elpstest/...` (0 issues); `bash -n scripts/mutation-proof.sh`; `git apply --check` + `go build ./...` on each of the three copier mutations. No suites, no `-race`, no fuzz campaign, no benchmark, no `make mutation-proof`.


sam-at-luther:
Benchmark gate status on the current head (6f80732), and what it needs.

Everything else is green: build and test in both builds, all 11 fuzz shards (the FuzzAliasGuard crasher on e8dca7d is fixed at e1910d8 and seeded), and the mutation-proof gate with the three copier rows.

The gate is red on three load-path rows only:

| row | delta | gate |
|---|---|---|
| LoadIntoEnv/library-78KiB/reader-cache-copy sec/op | +15.3% | 15% |
| LoadIntoEnv/library-78KiB/reader-cache-copy B/op | +8.6% | 5% |
| TextLoaderLoad B/op | +8.9% | 5% |

Package geomean for lisp: sec/op +0.15%, B/op -0.88%. The sort rows this PR was first red on are no longer regressions (3592b6b).

Cause: the per-header memo that makes `(*LVal).Copy` cycle-safe and alias-preserving (the same class as the fork walker) costs one map entry per copied node. On a parse tree that is roughly 16 to 20 bytes per node against about 150 bytes for the node itself. It is a per-load cost, not a per-transaction one; forks do not pay it. The size hint in 48a0cab removed nothing, which is how the map-growth diagnosis was ruled out (the rows are byte-identical before and after it). The reader-cache-copy arm calls the public `Copy` from an embedder-shaped Reader, so no internal hint can reach it.

Options, for the reviewer to pick:

1. Accept the load-path cost with a reviewed, dated waiver for these rows (`cmd/benchgate/waivers.go`, `-waivers`). Recommended. Nothing changes in the code.
2. Exempt leaf headers from the memo. Cuts the entries by about 70%, but the alias guard's fingerprint gives every header an ordinal, so the oracle would need a `Copy`-only exemption. Not recommended: it weakens the oracle to buy a load-time number.
3. Verify tree-ness once at TextLoader admission and use a memo-free walk for verified trees. Sound, but it only reaches TextLoaderLoad; the embedder arm still needs option 1.

Not re-running the job: the three rows have reproduced identically on four consecutive runs (50cdce0, e8dca7d, c1c929d, 6f80732).

---

sam-at-luther:

The released template plan replaces VM cloning, not Go LVal.Copy. [LVal.Copy](https://github.com/luthersystems/elps/blob/v1.61.0/lisp/lisp.go) still preserves array backing, shares byte payloads and map value pointers, and does not implement this PR's graph-wide payload memoization. Its proposed within-value-copy semantics, alias/cycle controls and performance effects remain separate work. NativeCloner remains relevant to value copy/detach, not template admission.

sam-at-luther:
Closing in favour of #642, which carries this PR's production commits rebased onto v1.61.1.

Not subsumed: all five `(*LVal).Copy` defects this PR fixes were measured live on main after #623 (map de-alias per header, shared `*[]byte`, shared LArray backing, copied macro-expansion metadata, per-comparison copies in the sort comparators). Kept as six isolated commits: memoisation + `copy`-builtin pin, macro-expansion drop, LArray walk, comparator-by-reference (the one behaviour change, kept droppable), allocation pin, load-path memo pre-size.

Dropped: `ee6b054`, `0e5c65f`, `38ee5b8`, `e8dca7d` (registry and mutation-gate hunks on #599 files), `fa3fbb9`/`24377ee` (the LSP fix, which is #609 and lands from there), and `6f80732` (imports `internal/walkraw`, which does not exist on main). The `elpstest.Walker` helpers in the tests were re-homed as package-local probes in `lisp/`.

---

PR #605, opened 2026-09-05
Audit every native payload type with an elpsvet analyzer

sam-at-luther:
## The class of bug

`Fork` copies the LVal spine and shares `LVal.Native` **by reference**. A template is forked once per request, so a mutable native payload that reaches a template is state shared by every fork in the process. The fork/isolation oracle in `elpstest` can only recognise a payload as stateful when its Go type declares `lisp.NativeCloner`; a mutable payload type that does not declare it is a leak channel no isolation test can see from the outside. The failure is a wrong answer with no error, only under forking.

This PR adds `elpsnativepayload`, the fourth `cmd/elpsvet` rule, ported from the substrate repository's nativepayload analyzer and adapted to the module that defines the constructors. It runs in CI through the existing `make elpsvet` target (both passes; the package list is `./...`, so there is no hand-scoped list to drift and no scope-drift grep to maintain).

## The rules

Every construction spelling is checked: `lisp.Native(x)`, `lisp.NativeOf[T](x)` (inferred or explicitly instantiated — the callee is unwrapped from the `*ast.IndexExpr`), a `lisp.Value(x)` the compiler can see falling through to Native (a defined type over `[]byte` does fall through), a keyed literal setting the `lisp.LVal.Native` field, a write to that field by any path (matched on the field object: `ErrorVal`, conversions, promoted fields), and taking the field's address.

A construction is reported unless one of:

1. the payload's static type has a **basic underlying type** (`unsafe.Pointer` excluded) — a soundness decision, not an allowlist one;
2. the type **declares `lisp.NativeCloner`** (checked structurally on the type's own method set, as an interface assertion would — a value of a pointer-receiver cloner is still reported);
3. the type is on the **audited allowlist** in `cmd/elpsvet/nativepayload.go`, each row keyed on the exact type (pointer-ness included) with its reason;
4. a **`//elpsvet:allow-native <justification>`** (at least three words) covers the site: trailing, standalone on the line above, or for a multi-line literal on the opening or the `Native:` line; or the enclosing function's doc. A bare or shorter marker does not suppress; one justification covers every construction on its line.

Unlike the substrate port, an **interface-typed payload (or type parameter) is reported**, not hidden behind a flag: this module is where `interface{}` enters the system, so the sites that pass it through carry an allow naming their contract rather than being silently unknowable.

## Classification of the real sites

| Site | Payload type | Verdict | Reason |
|---|---|---|---|
| `lisp/env.go`, `lisp/lisp.go`, `lisp/fork.go` (LFun) | `*funData` | allowlist | fork re-mints it per fork with the env remapped; detach refuses (Detach) or shares the LFun value whole (`copy`, same walker with `shareOpaque`) — inside one runtime, not a fork leak |
| `lisp.Bytes`, fork/detach `*[]byte` arms | `*[]byte` | allowlist | both walkers allocate a fresh slice and copy |
| `SortedMapFromData`, `copy`, fork/detach map arms | `*MapData` | allowlist | `f.mapData` / `detachMapData` / `copyMapData` rebuild the structure |
| `ErrorCondition[f]`, `SetCallStack`, fork/detach stack arms | `*CallStack` | allowlist | a `Stack.Copy()` snapshot at capture, never pushed again; both walkers deep-copy via `detachCallStack` |
| `lisp/env.go` `ErrorCondition` (`case error`), `lisp.ErrorCondition` | `error` | allowlist | read-only by contract: the kernel only formats it / hands it back; the row is a contract, not a proof |
| `libregexp` | `*regexp.Regexp` | allowlist | documented concurrency-safe after Compile; `Longest` (the one mutator) is never called |
| `libtime.Time` | `time.Time` | allowlist | value type, no mutating API |
| `libtime.Duration` | `time.Duration` | basic tier | defined type over `int64`, no row needed |
| `libschema.validatorMarkerCell` | `*validatorTag` | allowlist | pointer to a zero-size struct; the credential is the type, nothing to write through |
| `libjson.DumpMessageBuiltin` | `*ownMessage` | allowlist | `msg`/`loadable` are written on the one mint line and only ever read afterwards; `MessageBytesBuiltin` converts (not copies) `msg` into an LBytes, so the two alias within one runtime — forks copy bytes, so the fork claim holds |
| `libtesting` suite global | `*TestSuite` | NativeCloner | declares `CloneNative` on `main` (libtesting.go:113) |
| `lisp.Native` / `lisp.NativeOf` / `lisp.Value` fallthrough | `interface{}` / `T` | allow | the constructors themselves; every caller is checked at its own site |
| `fork.go` LNative + default arms, `detach.go` clone store | `interface{}` | allow | the walkers applying the policy (replacer → NativeCloner → share) to a payload already checked where it was minted |
| `env.go` `ErrorCondition` default arm | `interface{}` | allow | error-data cell, same read-only contract as the `error` row; the value stays the caller's |
| `env.Terminal` | `*LEnv` | allow | `LMarkTerminal` trampoline marker consumed by the eval loop that produced it, never bound; detach refuses it |
| `libgolang` struct-field accessor | `interface{}` (reflected) | allow | projection of an existing native payload; aliases at most what its parent already shares |
| `internal/fuzzval.native` | assorted | allow (func doc) | fuzz corpus generator, minted fresh per iteration; the mutable shapes are the point |

Nothing genuinely mutable and unprotected was found on `main`. The one thing worth flagging: an earlier checkout (28747b4, before `TestSuite.CloneNative` landed) would have been reported at the libtesting site — which is exactly the gap this rule exists to make visible.

## How a new payload type fails

`lisp.Native(&myHandle{})` anywhere in the module fails `make elpsvet` with a message naming the type and the three ways out (implement `NativeCloner`, add an allowlist row with a reason, or annotate with a justified allow). Adding an allowlist row also requires adding it to the audited inventory in `nativepayload_test.go`, so the reason is reviewed.

## Test plan

- `cmd/elpsvet/nativepayload_test.go` + `testdata/src/nativepayload`: every spelling, the basic tier, both `NativeCloner` receiver kinds (and a wrong-shaped `CloneNative`), allowlist keyed on exact type, interface-typed reports, and the allow marker in every placement with and without a justification.
- `go run ./cmd/elpsvet -test=false ./...` is clean on this branch; CI runs it via `make elpsvet`.

## Review round 1 (fixed in `fb65f1b` analyzer, `3afb543` docs)

| # | Finding | Fix |
|---|---|---|
| 1 | `isLValType` keyed on the receiver being named `LVal`, so writes through `lisp.ErrorVal` (`type ErrorVal LVal`) — `&lisp.ErrorVal{Native: h}`, `e.Native = h`, `(*lisp.ErrorVal)(v).Native = h` — were invisible | Match the **field object**: `TypesInfo.Selections[sel].Obj()` for writes, `TypesInfo.Uses[key]` for literal keys, accepted when it is the field named `Native` declared in package lisp (`isNativeField`). Fixture: `review.go` `errorValLiteral`, `errorValAssign`, `errorValConversionAssign` |
| 2 | Promoted-field write through embedding (`w.Native = h`) invisible | Closed by 1 (Selections resolves the promoted path to the same object). Fixture: `promotedWrite`, `promotedPointerWrite`, `explicitEmbeddedWrite` |
| 3 | Shared `//elpsvet:allow` marker: `elpsownership` accepts a bare marker, so one unrelated sentence on a package-level native silenced both rules | Own marker `//elpsvet:allow-native`; every annotation, doc and fixture updated. `elpsownership`'s `allowed()` now stops at the marker's word boundary so `allow-native` does not satisfy it either (`TestOwnershipAllowStopsAtWordBoundary`). `docs/sealed-ast.md` now states what each rule actually enforces: ownership — bare marker suppresses, justification is convention; native — three-word justification enforced, one per line covers the line |
| 4 | `p := &v.Native; *p = h` unreported and undocumented | Any `&<LVal>.Native` (UnaryExpr `&` over the Native field selector, parenthesised or promoted) is reported unconditionally — there is no payload type at that site. Multi-value assign stays a documented blind spot; reflection and positional literals added to the "what it does not see" list. Fixture: `addressOfNative*` (4 reported, 1 allowed, 2 negative) |
| 5 | Literal reported at `kv.Value.Pos()`, so a marker on the `&lisp.LVal{` line did not cover a `Native:` two lines down | Report at `lit.Pos()`; a marker on the opening line, the `Native:` line (key or value), or standalone above the opening line all suppress; a marker on some other field's line does not. Fixture: `multiLineLiteral*` (5 cases) |
| 6 | A one-character justification (`//elpsvet:allow .`) suppressed | Justification must be **at least three words** (`nativeAllowMinWords`), documented in the analyzer header, CLAUDE.md and sealed-ast.md; `TestJustifiedNativeAllow` pins `.`, one and two words as non-suppressing, three as suppressing. One justification covering every construction on its line is stated as deliberate |
| 7 | `*funData` row said detach "refuses LFun"; `copy` runs the same walker with `shareOpaque` and shares the value whole | Row reworded: fork re-mints (fork.go LFun arm); detach refuses (Detach) or shares the LFun value whole (copy) — inside one runtime, not a fork leak |
| 8 | `.github/workflows/elps.yml` comment said "three rules" | Now four, with the fourth named |
| 9 | (note) `MessageBytesBuiltin` converts `m.msg` rather than copying | Recorded in the `*ownMessage` row's reason so the next reader does not re-derive it |

**Evidence for 1/2/4/5:** the new `review.go` fixture was run against the unchanged `b38ea1f` analyzer before the fix. Result: `no diagnostic was reported` at review.go:19, 23, 28 (ErrorVal literal/assign/conversion), 34, 38 (promoted writes), 48, 53, 57, 61 (address-of), and 79, 108 (multi-line literal wanted on the opening line), plus `unexpected diagnostic` at :81, :88, :95, :103, :110 (the old analyzer reporting the multi-line literals on the `Native:` line — including the ones whose marker sits on the opening line). On `fb65f1b` the whole package test passes.

Local verification on `3afb543`: `gofmt -l` on touched files, `go build ./...`, `go vet ./cmd/elpsvet/...`, `go test -run . ./cmd/elpsvet`, `golangci-lint run ./cmd/elpsvet/...` (0 issues), one `go run ./cmd/elpsvet -test=false ./...` (exit 0, no findings).

### DIFF
--- a/.github/workflows/elps.yml
+++ b/.github/workflows/elps.yml
@@ -111,10 +111,11 @@ jobs:
         run: make test-elpscheck
 
       # The seal contract's static half.  golangci-lint above checks Go style;
-      # this checks the three rules that exist because a raw Go write can
+      # this checks the four rules that exist because a raw Go write can
       # launder around the seal bit -- package-level LVal tables, writes to
-      # LVal fields on values the function did not construct, and escaping
-      # uncopied *token.Location pointers.  ~3s.
+      # LVal fields on values the function did not construct, escaping
+      # uncopied *token.Location pointers, and native payloads minted
+      # without a fork-safety classification.  ~3s.
       #
       # Runs TWO passes.  The second is the whole reason this is a make target
       # rather than one line here: elpsvet accepts -tags and silently ignores
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -122,7 +122,62 @@ Functions return `*LVal`. Errors are LVal values with type `LError`. Check with 
 - **`analysis/`** — Semantic analysis: scope building, symbol resolution, reference counting. `prescan()` uses a two-phase approach (definitions first, exports second) so `(export 'name)` before `(defun name ...)` works correctly.
 - **`lint/`** — Static analysis modeled after `go vet`. Each check is an `Analyzer` with a `Run func(pass *Pass) error`. Uses `Walk()`/`WalkSExprs()` for AST traversal, plus custom walkers for context-sensitive checks.
 - **`diagnostic/`** — Rust-style annotated source snippets for error and lint output. Zero dependencies on `lisp/`.
-- **`cmd/elpsvet/`** — Go static analyzers over elps's *own* Go source, enforcing invariants the compiler cannot: the freshness rule (writes rooted at `lisp.LVal` storage) and the slice-alias tracking that catches writes laundered through a local alias first (issues #369, #371).
+- **`cmd/elpsvet/`** — Go static analyzers over elps's *own* Go source, enforcing invariants the compiler cannot: the freshness rule (writes rooted at `lisp.LVal` storage), the slice-alias tracking that catches writes laundered through a local alias first (issues #369, #371), and the native-payload audit (see "Go static analysis over elps's own sources" below).
+
+## Go static analysis over elps's own sources
+
+`cmd/elpsvet` is a `golang.org/x/tools/go/analysis` harness over elps's own
+Go code: focused analyzers, an audited `//elpsvet:allow` suppression carrying
+a justification, and testdata-driven `analysistest` fixtures under
+`cmd/elpsvet/testdata/src`. CI runs it as `make elpsvet`, which is two passes
+of `go run ./cmd/elpsvet -test=false ./...` (the second under
+`GOFLAGS=-tags=elpscheck`, because elpsvet silently ignores `-tags`; see the
+Makefile comment). The package list is `./...`, so there is no hand-scoped
+list to drift away from where the code lives and no scope-drift check to
+maintain.
+
+Its fourth rule, `elpsnativepayload` (`cmd/elpsvet/nativepayload.go`), closes
+a class of bug the fork/isolation tests cannot see from the outside: `Fork`
+shares `lisp.LVal.Native` **by reference**, so a mutable native payload that
+reaches a template is state shared by every fork in the process, and the
+isolation oracle in `elpstest` only recognises a payload as stateful when its
+type declares `lisp.NativeCloner`. The rule is ported from the substrate
+repository's nativepayload analyzer and adapted to the module that defines
+the constructors.
+
+A native construction is any of these SPELLINGS, and the list is the first
+thing to check when the gate looks suspiciously quiet: `lisp.Native(x)`, the
+typed `lisp.NativeOf[T](x)` (inferred or explicitly instantiated — the latter
+wraps the callee in an `*ast.IndexExpr`, which `calleeFunc` unwraps), a
+`lisp.Value(x)` the compiler can see falling through to Native (a defined
+type over `[]byte` DOES fall through; the type switch matches the unnamed
+type only), a `lisp.LVal{Native: x}` literal, and a write to a `.Native`
+field. `NativeOf` needs its own arm even though it is *implemented* as a call
+to `Native`: a generic instantiation resolves to the generic `*types.Func`,
+whose name is never `Native`.
+
+A construction is reported unless the payload's static type has a basic
+underlying type (a value of which is immutable inside an interface —
+`unsafe.Pointer` excluded), declares `lisp.NativeCloner` (the kernel's clone
+protocol, checked structurally as an interface assertion would), is on the
+audited allowlist in `cmd/elpsvet/nativepayload.go` — the kernel's own
+representation slots (`*funData`, `*[]byte`, `*MapData`, `*CallStack`, each
+with an explicit fork/detach arm), `*regexp.Regexp`, `time.Time`, `error`,
+libschema's `*validatorTag`, libjson's `*ownMessage`, each row carrying its
+reason — or a `//elpsvet:allow <justification>` comment covers the line (or
+the enclosing function's doc). **A bare `//elpsvet:allow` with no
+justification does not suppress.** An interface-typed payload (or a type
+parameter) is REPORTED, not skipped: this module is where `interface{}`
+enters the system, so the constructors themselves (`Native`, `NativeOf`,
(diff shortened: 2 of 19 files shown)


sam-at-luther:

Template publication now rejects unknown native payloads by default, and library-owned time/regexp wrappers have explicit ownership contracts. This is not the repository-wide constructor audit proposed here: [elpsvet still registers three analyzers](https://github.com/luthersystems/elps/blob/v1.61.0/cmd/elpsvet/main.go), not elpsnativepayload. The additional analyzer and its construction-spelling controls remain independent work. Its old Fork/NativeCloner and raw time/regexp exemptions must be re-audited against the released contracts rather than copied unchanged.

sam-at-luther:
Closing in favour of #642, which carries the analyzer with its rules re-audited against template admission.

Kept as reviewed: the AST and type matching (every construction spelling, `NativeOf`'s own arm, field-object matching through `ErrorVal` and promotion, `&v.Native`), the `//elpsvet:allow-native` marker with its three-word floor, and the fixtures.

Changed, because #623's merged tests refute the rules as written: the "implements `NativeCloner` therefore safe" tier is gone (`lisp/native.go` says it is not an admission protocol; `lisp/template_test.go` asserts rejection) and is replaced by the `templatepolicy.Marker` struct-value tier, mirroring `templateInventory.native` exactly. The `time.Time` and `*regexp.Regexp` rows are gone (`lisp/lisplib/template_natives_test.go` shows both mutable through Go); `*lisp.CallStack` is reported with a pointer at `checkDiagnosticPayload` rather than allowlisted; `error`, `*libjson.ownMessage` and `*validatorTag` rows are gone (interface, rejected-at-publication pointer, and Marker-tier value respectively). The diagnostic no longer cites Fork or recommends `NativeCloner`. Rebased unchanged, the gate was red on main with four findings, three of them the types #623 audited and the runtime admits.

---

PR #609, opened 2026-09-05
Bound LSP document parsing by the size limit, not only analysis

### LINKED ISSUE #607: LSP: DocumentStore.Open/Change parse the whole document before the size gate; a large didChange hangs the session
## What

`DocumentStore.Change` (lsp/document.go:113-127) and `DocumentStore.Open` (lsp/document.go:98-109) parse the entire document unconditionally. `maxDocumentBytes` is consulted only in `analyzeAndPublish` (lsp/diagnostics.go:124-130), after the parse has already run. So the size cap that the fuzz target's header assumes makes over-limit documents cheap does not cover the parse step at all: an untrusted client can make every keystroke cost a full parse of an arbitrarily large document.

The parser is linear (measured about 250 to 400 ns per byte with `ParseProgramFaultTolerant` at 2^16 to 2^21 lines), so this is untrusted size times ungated work, not a complexity bug. A 32 MiB document costs about 17 s and 16.7 M AST nodes per `didChange`.

## How it surfaced

A 10-minute CI fuzz campaign (run 33941631011, shard 10) on `FuzzLSPSession` produced crasher `36ec6d2dd72dfd1f`: the fuzz worker died with `exit status 2`. It is a hang, not a panic: with `-timeout 12s` the session goroutine is `[runnable]` in the lexer mid-parse of a 0xffffff-byte document:

```
lsp.(*DocumentStore).Change(..., {0xc02513a000, 0xffffff})   lsp/document.go:121
lsp.(*Document).parse                                         lsp/document.go:70
parser/rdparser.(*Parser).ParseProgramFaultTolerant           parser.go:1155
parser/lexer.(*Lexer).readNumber                              lexer.go:355
parser/token.(*Scanner).EmitToken                             scanner.go:90
```

Reproduces identically on `main` (a32ada7), so it is pre-existing and unrelated to the fork/alias work in the #599 stack, where it was found.

## Minimal repros (verified on main)

1. Fuzz seed (26 ops, document B empty): hits the harness's 30 s watchdog.
   ```
   go test fuzz v1
   []byte("0")
   []byte("")
   []byte("\xaa" x 104)
   ```
   Decode: every op is `didChange` with the harness's "duplicate" edit shape (`cur + "\n" + cur`), so document A doubles on each op.

2. Harness-free, in package `lsp` (fails in about 17 s): `New(WithMaxDocumentBytes(16384))`, open `"0"`, then `textDocumentDidChange` with `strings.Repeat("0\n", 16<<20)`. Afterwards `s.docs.Get(uri).ast` holds 16,777,216 expressions; only the later `analyzeAndPublish` emits "File exceeds size limit for semantic analysis".

## Fix options

- Check `maxDocumentBytes` in `Open` and `Change` before parsing (keep the text, skip the parse, mark the document over-limit so diagnostics report it as they do today).
- Harness: the duplicate edit shape in `lsp_fuzz_test.go` is uncapped, so even with the server fixed this input would build a 2^49-byte string; cap `editedContent`'s doubling at a small multiple of `fuzzMaxDocBytes`.
- Commit the crasher as a permanent seed once the fix lands.

Not fixed in the #599 stack (#604) because it is a `main` defect in the LSP, independent of the walkers.

sam-at-luther:
## Summary

Fixes #607: the pre-existing LSP hang the nightly 10-minute fuzz campaign found (run 33941631011, shard 10) as `FuzzLSPSession/36ec6d2dd72dfd1f`. It reproduces on plain `main` (a32ada7).

The server's `maxDocumentBytes` limit only gated analysis and lint in `analyzeAndPublish`; `DocumentStore.Open`/`Change` parsed the full document first, on every didOpen and every didChange. Parsing is the expensive part (the AST is ~100 bytes per source byte, rebuilt per edit), so the limit bounded nothing that mattered. This implements the three fix options from #607, plus the QA review's follow-ups.

## Root cause, with evidence

The script is 48 `didChange` ops (`0xaa % 29 = 25`), each taking the harness's duplicate shape (`a.b % 3 = 2`, `cur + "\n" + cur`), so doc A doubles per op. Replaying it op by op with timers:

```
op 18 docA=1048575   didChange(parse)=952ms    analyzeAndPublish=46us
op 20 docA=4194303   didChange(parse)=3.85s    analyzeAndPublish=36us
op 22 docA=16777215  didChange(parse)=5.34s    analyzeAndPublish=251us
op 24 docA=67108863  didChange(parse)=14.09s   analyzeAndPublish=37us  heap=10877MB
```

A CPU profile of that replay puts 65.6% of samples under `lsp.(*Server).textDocumentDidChange` -> `(*DocumentStore).Change` -> `(*Document).parse` -> `rdparser.(*Parser).ParseProgramFaultTolerant`, with nearly all the rest in GC marking the AST parse allocated. `analyzeAndPublish` returned in microseconds every time once the doc passed 16 KiB: the analysis-side guard worked and was irrelevant. Matches #607: a linear parser on untrusted size, ungated.

The harness has its own defect on the same input: 48 doublings of one byte is 2^48 bytes. With only the server fix applied, the crasher still dies with `fatal error: out of memory` in the harness's `decode` -> `strings.Split` on a 512 MiB document at op 29, so no server change alone makes that input tractable.

## The fix (the options in #607)

1. `lsp/document.go`, `lsp/diagnostics.go`: `Document.parse(maxBytes)` checks the limit before parsing. An over-limit document keeps its text, gets no AST or parse errors, and is marked `overLimit`. `analyzeAndPublish` reports from that mark through the **same version guard** as the analysed path, so `publishedVersion`/`publishedDiagnostics` stay current for code actions and a stale over-limit result cannot overwrite a newer publication. The diagnostic now reads "File exceeds size limit: parsing, semantic analysis, formatting and AST-based features are disabled for this document". `DocumentStore` carries the server's `maxDocumentBytes`. Default (no limit) unchanged.
2. `lsp/formatting.go`, `lsp/hover.go`: both honour the mark. `formatter.FormatFile` parses the text itself (3s on an 8 MiB over-limit doc before this), and hover's registry fallback `strings.Split`s the whole text per request to find the word under the cursor. Both are linear over a document the limit exists to keep off the hot path.
3. `lsp/lsp_fuzz_test.go`: `editedContent`'s duplicate shape is bounded by an absolute `fuzzMaxContentBytes = 16 MiB`, checked on the result, past which a duplicating edit halves (undoing the previous duplication exactly). `decode` computes line count/width without materialising `strings.Split` (128 MiB of headers per op at this size). Oracle unchanged; input not skipped.
4. `lsp/testdata/fuzz/FuzzLSPSession/36ec6d2dd72dfd1f` committed as a permanent seed.

**Godoc correction vs. the review:** the didSave path is *not* bounded by `ScanConfig.MaxFileBytes`: `updateFileDefinitions` and `updateFileRefs` do a bare `os.ReadFile` + parse with no size check. `WithMaxDocumentBytes`'s godoc says so explicitly rather than claiming otherwise. (Workspace scanning is bounded by `MaxFileBytes`.)

## Why the cap is 16 MiB (QA item 1)

The seed only discriminates if an unfixed server does more than 30s of parsing across the session. At a 4 MiB cap the unfixed tree finished the seed in **15.9s and PASSED** (parser runs ~100-160 ns/byte). At 16 MiB the session parses ~550 MiB on an unfixed server, which no box finishes inside the budget, while the fixed tree stores each edit in microseconds. `TestLSPFuzzDuplicateEditIsBounded` pins that the shape never exceeds the cap, peaks at the largest 2^k-1 under it, and that both steady-state legs are over `fuzzMaxDocBytes` (the over-to-under transition comes from the truncate/swap shapes, as the comment now says).

## Tests

- `TestDidChange_OverLimitDocumentIsCheap` — #607's harness-free repro, bounded by allocation (`TotalAlloc` delta < 1 MiB; the parse allocates gigabytes), not wall clock.
- `TestDocumentSizeGuard_OversizedDocIsNotParsed` — store skips the parse on open and on change, re-parses back under the limit, no limit still parses every atom.
- `TestDocumentSizeGuard_OverLimitDocumentHasNoASTFeatures` — table over hover, definition, completion, documentSymbol, foldingRange, semanticTokensFull, selectionRange, documentHighlight, prepareCallHierarchy: non-empty under the limit, empty with nil error over it.
- `TestFormatting_OverLimitDocumentIsCheap` — under-limit doc formats; 1 MiB over-limit doc returns no edits with < 1 MiB allocated.
- `TestDocumentSizeGuard_OverLimitPublishIsVersionGuarded` — over-limit publish records its version and diagnostics; a stale result is not re-sent over a newer version.
- `TestLSPFuzzDuplicateEditIsBounded` — harness self-test, above.
- `runDebouncedNow` is the shared stop-timer/run-inline helper used by the tests and the fuzz harness.

## Test plan

```
# Seed discrimination (QA item 1): origin/main's server files + this PR's harness + seed
go test ./lsp -run 'FuzzLSPSession/36ec6d2dd72dfd1f' -count=1
  --- FAIL: FuzzLSPSession/36ec6d2dd72dfd1f (30.00s)
      the LSP session did not terminate within 30s of SCHEDULED time
      (wall 30s, of which 30s scheduled and 0s lost to scheduler stall)
# Same seed on this branch's head
  --- PASS: FuzzLSPSession/36ec6d2dd72dfd1f (0.36s)

go test ./lsp -run TestDidChange_OverLimitDocumentIsCheap -count=1
  origin/main + the test: FAIL
      "an over-limit didChange must not build an AST" (16,777,216 nodes)
      didChange took 7.09s; document not marked
  head:  --- PASS (0.10s)

go test ./lsp -run TestDocumentSizeGuard_OversizedDocIsNotParsed -count=1
  origin/main + the test: FAIL
      "an over-limit document must not be parsed on open" / "on change"
  head:  PASS

go build ./...                      ok
go vet ./lsp                        ok
gofmt -l <touched files>            clean (lsp/semantic_tokens_test.go is flagged on main already; untouched here)
golangci-lint run ./lsp             0 issues
go test ./lsp -count=1              ok (3.9s)
go test -race ./lsp -count=1        ok (6.4s)
```

Closes #607


sam-at-luther:
Closing in favour of #642, which carries this fix rebased onto v1.61.1 with two review findings folded in.

Review found four handlers still linear in document size on an over-limit document even after this PR: `definition` and `completion` (via `wordAtPosition`, which splits the whole content, and `wordBoundsAtPosition`, which splits it again), `signatureHelp` (`enclosingCallText`) and `foldingRange` (`commentFoldingRanges` runs unconditionally after the nil-AST walk). Measured at 8×, 8×, 4× and 4× the document in allocation. The `WithMaxDocumentBytes` godoc and the user-visible diagnostic named `definition`, `completion` and `folding` as disabled, and `TestDocumentSizeGuard_OverLimitDocumentHasNoASTFeatures` asserted emptiness only, so `foldingRange`'s row passed because the fixture had no comment lines.

In #642: one `Document.OverLimit()` helper replaces the three copied lock blocks and guards all seven sites; the fixture carries a comment block; a new test holds each handler under 64 KiB of allocation on an 8 MiB document (proven to fail with the guards reverted). The godoc sentence saying the didSave path is unbounded is dropped, since #612's commit bounds it. The per-request double split in `wordAtPosition` under the limit is a separate bug, filed as #641.

---

PR #612, opened 2026-09-05
Bound the LSP didSave disk reads by the workspace scan's file size limit

### LINKED ISSUE #611: LSP: the didSave path re-reads and parses the saved file with no size check (same class as #607)
## What

`textDocument/didSave` (`lsp/diagnostics.go:71`) calls `updateFileDefinitions` and `updateFileRefs` (`lsp/server.go:618` and `:562`), which do a bare `os.ReadFile` of the saved file and parse it with no size check. `maxDocumentBytes` (after #609) bounds the didOpen/didChange/formatting/AST-feature paths, and `ScanConfig.MaxFileBytes` bounds the workspace scan, but neither reaches this path. `lsp/position_encoding.go:354` has another bare `os.ReadFile`.

So a client can make each save cost a full parse of an arbitrarily large file, the same untrusted-size, ungated-work class as #607, on a different entry point.

## How it surfaced

Found while applying review to #609: the `WithMaxDocumentBytes` godoc was going to claim the didSave path is bounded by `ScanConfig.MaxFileBytes`, and reading the code showed it is not. The godoc in #609 now states the truth instead.

## Fix options

- Apply the same size gate (`maxDocumentBytes` or `ScanConfig.MaxFileBytes`, whichever the server treats as the disk-read bound) before the parse in `updateFileDefinitions` / `updateFileRefs`, and at `position_encoding.go:354`.
- Prefer `os.Stat` then a bounded read, so the file is not even read into memory when over the limit.
- Pin with a test in the style of `TestDidChange_OverLimitDocumentIsCheap` (allocation-bounded, not wall-clock): save an over-limit file, assert the definitions/refs tables are untouched and the allocation delta is small.
- `FuzzLSPSession` cannot reach this path unless it writes files; a unit test is the right pin.

Follow-up from #609 / #607.

sam-at-luther:
## Summary

Fixes #611. `textDocument/didSave` refreshes the workspace index from the saved file on disk, and the three reads on that path — `updateFileDefinitions` (`lsp/server.go`), `updateFileRefs` (`lsp/server.go`), and `documentTexts.get` (`lsp/position_encoding.go`, the rename fallback for a closed file) — were bare `os.ReadFile` calls followed by a full parse, with no size check. They now honour the same per-file limit the workspace scan applies, stat-first so an over-limit file is never read into memory.

Independent of #609: no helper from that branch is used, so either can merge first.

## Root cause

The workspace scan that builds `ExtraGlobals` and `workspaceRefs` skips any `.lisp` file over `ScanConfig.MaxFileBytes` (default `analysis.DefaultMaxFileBytes`, 5 MiB) by its `Stat` size before opening it (`analysis/workspace.go` `collectLispFilesWithConfig`). The incremental refresh of those same tables on save had no such check, so a file the scan would never index was read and parsed in full on every save — the same untrusted-size, ungated-work class as #607, on a different entry point. On `main` the new test measures `updateFileDefinitions` allocating **499,359,464 bytes** for one over-limit save, and the whole before-run took 350s.

## Which limit, and why

The scan's `MaxFileBytes`, not `maxDocumentBytes`. `maxDocumentBytes` bounds text the client sends; this path reads files from disk to refresh the scan's own tables, and a save must not index what the scan refused to. Stated in the `scanConfig`/`readWorkspaceFile` godoc.

## The fix

- `analysis/workspace.go`: export `ScanConfig.EffectiveMaxFileBytes()` (wrapper over the existing unexported method) so the LSP uses the scan's number rather than a copy.
- `lsp/server.go`: `Server.scanConfig()` is now the single source of the scan config for both `buildWorkspaceIndex` and the incremental refresh. `readWorkspaceFile(path)` does `os.Stat` and rejects a directory or a file over the limit before opening it, then reads through `io.LimitReader` at limit+1 so a file that grows between the `Stat` and the read is still rejected rather than parsed. A missing, unreadable or over-limit file returns `ok=false` and the caller leaves its tables untouched, exactly as if the scan had skipped it. `updateFileDefinitions` and `updateFileRefs` use it.
- `lsp/position_encoding.go`: `documentTexts.get` uses the same bounded read; an over-limit closed file yields no text and the caller leaves the range in byte columns, as it already does for an unreadable file.

## Tests

`lsp/didsave_size_gate_test.go`, allocation-bounded in the style of `TestDidChange_OverLimitDocumentIsCheap`:

- `TestDidSave_OverLimitFileLeavesIndexUntouched` — sentinel entries in `ExtraGlobals` and `workspaceRefs`; positive control: an under-limit save of a two-function program adds its definitions and its `target<-caller` reference; then an over-limit save of the same program repeated past 5 MiB leaves both tables deep-equal to the sentinels and each of `updateFileDefinitions`/`updateFileRefs` allocates under 1 MiB.
- `TestDocumentTexts_OverLimitFileIsNotRead` — under-limit closed file is read; over-limit one yields `("", false)` with under 1 MiB allocated.

Both compile unchanged against `main` (they only call symbols that already exist there), so the before/after is a like-for-like run.

## Test plan

```
go test ./lsp -run 'TestDidSave_OverLimitFileLeavesIndexUntouched|TestDocumentTexts_OverLimitFileIsNotRead' -count=1 -v
  origin/main + the tests:
    --- FAIL: TestDidSave_OverLimitFileLeavesIndexUntouched
        "an over-limit save must leave ExtraGlobals untouched"
        "499359464" is not less than "1048576"
        updateFileDefinitions on an over-limit file must not read or parse it (allocated 499359464 bytes)
    FAIL  github.com/luthersystems/elps/lsp  350.746s
  head:
    --- PASS: TestDidSave_OverLimitFileLeavesIndexUntouched (0.02s)
    --- PASS: TestDocumentTexts_OverLimitFileIsNotRead (0.01s)
    ok    github.com/luthersystems/elps/lsp  0.070s

go build ./...                              ok
go vet ./lsp ./analysis                     ok
gofmt -l <touched files>                    clean
golangci-lint run ./lsp ./analysis          0 issues
go test ./lsp ./analysis -count=1           ok (3.4s / 1.0s)
go test -race ./lsp -count=1                ok (5.4s)
```

Closes #611

---

### DIFF
--- a/analysis/workspace.go
+++ b/analysis/workspace.go
@@ -51,6 +51,14 @@ func (c *ScanConfig) effectiveMaxFiles() int {
 	return c.MaxFiles
 }
 
+// EffectiveMaxFileBytes returns the per-file size limit the scan applies,
+// with the default filled in when MaxFileBytes is unset. Exported so that
+// callers refreshing the scan's tables one file at a time (the LSP's didSave
+// path) skip exactly the files the scan would have skipped.
+func (c *ScanConfig) EffectiveMaxFileBytes() int64 {
+	return c.effectiveMaxFileBytes()
+}
+
 // effectiveMaxFileBytes returns the file size limit, applying the default if zero.
 func (c *ScanConfig) effectiveMaxFileBytes() int64 {
 	if c == nil || c.MaxFileBytes <= 0 {
--- a/lisp/loadcache_reader_fuzz_test.go
+++ b/lisp/loadcache_reader_fuzz_test.go
@@ -57,6 +57,23 @@ import (
 // Also asserted, in every mode: termination under the scheduled-time
 // watchdog, and no recovered Go panic (lisp.IsInternalPanic) — both from the
 // shared budgeted harness.
+//
+// "Indistinguishable" means BY WHAT A PROGRAM CAN OBSERVE: every load's
+// value, and every binding's content, sealed-ness, mutable-storage aliasing
+// and sealed-node PROVENANCE (the frozen parse location every error and
+// stack note through that node prints -- see envStateFingerprintProv, which
+// this target uses and its siblings do not).  It does not mean the same
+// storage: a cache hit serves the
+// sealed nodes the miss served (that is the contract, pinned by
+// TestLoadCacheServesTheSameNodes), so a file loaded twice through the cache
+// rebinds its literals to nodes an earlier load already captured, where two
+// fresh parses give two equal nodes.  The environment digest used to record
+// that identity and reported the contract as a divergence (issue #613, seed
+// bfe6dc31a652ced0: A binds a literal, B captures it, A is loaded again);
+// valueFingerprint now digests sealed nodes by content alone (memoised by
+// pointer, so a sealed DAG or cycle stays cheap), and
+// TestLoadCacheRebindAcrossLoadsMatchesFreshParse pins the shape
+// deterministically in every mode.
 
 // loadCacheHostileReader wraps the real parser and then does one plausible,
 // contract-legal thing to what it returns.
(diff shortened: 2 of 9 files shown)


sam-at-luther:
Standing down on the red fuzz shard, with the reason.

Failing check: Fuzz shard 3/10 on b56a8de, target `FuzzLoadCacheHostileReader`, crasher `bfe6dc31a652ced0` (artifact 9965388451). This PR touches only `lsp/` and `analysis/`; the crasher is in package `lisp` and reproduces identically on plain `main` (a32ada7) and on the #599 stack head (b076a2f), so it is not this PR's. Tracked as #613, with a fix branch in progress; the fix will be ported into this PR when it exists so the sweep goes green here too. No re-run: the failure is deterministic and reproduced locally on `main`, so a re-run would only burn the runner.

---

sam-at-luther:
Fuzz shard 3/10 was red on `b56a8de`, and the failure was not this PR's: it was `FuzzLoadCacheHostileReader` finding `bfe6dc31a652ced0`, in `lisp`, which this PR does not touch (its diff is the LSP didSave path plus the workspace size limit).

That crasher is the pre-existing load-cache oracle artefact fixed in #614. Rather than leave this PR red waiting on that one, I have merged `fix/loadcache-zero-width-slice-613` into this branch and pushed `16e10fd`. The port is test-only (the oracle's sealed-content memo and scoped provenance, plus the seed) and no-ops once #614 lands first.

Verified locally on the merged branch:

| Check | Result |
|---|---|
| `go build ./...` | ok |
| `go test ./lisp` (incl. the `bfe6dc31a652ced0` seed) | ok, 47.4s |
| `go test ./lsp` | ok, 6.5s |
| `go test ./elpstest ./analysis` | ok |

---

sam-at-luther:
Closing in favour of #642, which carries this PR's own commit (`b56a8de`) rebased onto v1.61.1 with one review finding folded in. The merged-in `f466bb1`/`16e10fd` from the unlanded #613 branch is not carried.

Review found the fix sound and the defect worse than described: reverting it makes `TestDidSave_OverLimitFileLeavesIndexUntouched` hit the 15-minute test timeout inside `ExtractFileRefs`, which calls `FindEnclosingFunction` and `ScopeAtPosition` per reference, so the unbounded path is super-linear, not "a full parse". But the new `readWorkspaceFile` godoc claimed an over-limit file leaves the index "exactly as if the scan had skipped it", and that was only true for a file never indexed: a file indexed while small and then saved over the limit kept every definition and reference, so go-to-definition pointed at symbols the file no longer defined.

In #642: `readWorkspaceFile` returns a status; over-limit purges the file's entries through the existing `filterExtraGlobalsByFile` and a shared `dropFileRefsLocked`, while missing or unreadable leaves them alone. A new test pins the purge, that a deleted file keeps its entries, and that the purge is file-scoped; proven to fail with the purge removed.

---

PR #614, opened 2026-09-05
Digest sealed nodes by content in the fuzz env-state oracle (#613)

### LINKED ISSUE #613: Load cache changes post-load environment state for a zero-width slice of a literal (FuzzLoadCacheHostileReader/bfe6dc31a652ced0)
## What

`FuzzLoadCacheHostileReader` found a crasher on the PR CI fuzz sweep for #612 (run 33952519477, shard 3, artifact 9965388451). The `lisp` package is untouched by that PR, and the crasher reproduces on plain `main` (a32ada7) and on the #599 stack head (b076a2f), so it is pre-existing.

Input (reader mode 4):

```
A: (set'lit'(0))(and'1(slice'list lit 0 0))
B: (set'A(list lit))
```

Loading A, B, A through one environment and one shared cache gives the same three evaluation results as the same sequence with no cache, but the post-load environment fingerprint differs in its second component:

```
baseline  env state -> 9cf3603d1f4fd5e5/2c9070ae5a388967
cached    env state -> 9cf3603d1f4fd5e5/78e121c06a8e4a3b
```

The oracle's own confirmation discipline held (a fresh uncached run still matches the baseline, a second cached run still differs), so the cache is what changes the state.

## Repro

```
go test ./lisp -run 'FuzzLoadCacheHostileReader/bfe6dc31a652ced0' -count=1
```

with the crasher file from the artifact placed under `lisp/testdata/fuzz/FuzzLoadCacheHostileReader/`.

## Reading

The program takes a zero-width `slice 'list` of a sealed literal (the sealed empty carve-out) and then wraps the literal in a fresh list in a later load. With the cache, the second load of A serves the cached sealed parse, so `lit` and every view of it share the cache entry's storage across loads; without the cache each load parses afresh. Whatever `envStateFingerprint`'s second component encodes (seal state, storage identity, or sharing) is observably different between the two, which is what the target exists to forbid: a cached load must be indistinguishable from a fresh parse.

To determine: which component differs and why, whether it is observable from lisp (a write through one view reaching another load's value, or a seal refusal that differs between cached and uncached), and the fix.

Found while driving #612; same 30-second sweep that is green on #609.

sam-at-luther:
## What

`FuzzLoadCacheHostileReader` reported seed `bfe6dc31a652ced0` (reader mode 4) as a cache divergence. Loading A=`(set'lit'(0))(and'1(slice'list lit 0 0))`, B=`(set'A(list lit))`, A again through one env with a shared `LoadCache` gave the same three evaluation results as the same sequence with no cache, but a different `envStateFingerprint`.

**Classification: (iii) a harness/fingerprint artefact.** `valueFingerprint` (test-only oracle) keyed its back-reference bookkeeping on every node, sealed or not, so it encoded storage identity that a sealed cache entry legitimately shares. A cache hit serves the SAME sealed nodes the miss served (`TestLoadCacheServesTheSameNodes`), so the third load of A rebinds `lit` to the node B captured from the first load, where two fresh parses mint two equal nodes. No lisp program can observe WHICH of two equal sealed nodes it holds: a sealed node cannot be written through and `equal?` is structural.

Fix: the oracle digests SEALED nodes by content and keeps aliasing bookkeeping for UNSEALED nodes, where identity is observable. Nothing in `lisp/` (non-test) changed.

Closes #613

## Adversarial-review follow-ups

A QA review confirmed the classification (154 probes found no lisp-observable difference) but found the weakening broader than the artefact required. Three things landed on top of the original fix.

### 1. Sealed content hashes are memoised by pointer

Digesting sealed nodes by content on every reach is exponential on a sealed DAG and a budget hog on a sealed cycle — and the node budget is shared across every value in one `envStateFingerprint`, so a hog blinds the digest to every value walked *after* it. That is a correctness bug, not a slow test. Reachable from three lines of lisp.

`valueFP` gains `memo map[*lisp.LVal]uint64` and `touchedUnsealed bool`. A sealed node under budget returns its memoised sum if present; otherwise it is walked into a private `fnv.New64a()` that SHARES `seen`/`budget`/`memo` with the parent, and the sum is cached only when `!touchedUnsealed`. A sealed subtree that reaches an unsealed node is not memoised — its digest embeds `back:N` markers and `seen` ids valid only at the visit that minted them. The parser cannot build that shape (`sealAST` stops at the first non-sealable type) but an embedder calling `SealAST` on a hand-built tree can.

A memo hit emits the bytes a full walk would have emitted, so the digest still says "this content", never "the node you saw at position N".

### 2. Sealed provenance, SCOPED

A sealed node's `source` is frozen (`SetSource` is a no-op once sealed) and IS lisp-observable — every error and stack note raised through the node prints it. That is why the cache keys on the stream's name and location, not its bytes alone (`loadcache.go` "# Keying"), and the content-only rule is blind to it.

`valueFingerprintProv` / `envStateFingerprintProv` additionally mix `src:<file>:<line>:<col>` for sealed nodes, and are used **only** from `runHostilePair` — on the per-load results as well as the env state. (For byte-identical files the A,B,A sequence rebinds the same symbols from the same bytes, so the final environment converges either way; load B's own *value* does not, which is where the mutation shows.)

Deliberately **not** global: `FuzzSharedProgramMultiEnv` evaluates one parse under several stream names on purpose, and a global rule turns that target, `TestSharedProgramSeedsAgreeWithSharedParse` and `TestValueFingerprintSealedIdentityIsNotState` red on their own premise.

### 3. Test and comment fixes

- The file comment's claim that "whether two bindings hold one sealed node or two equal ones is not a fact any program can observe" was false in general. It now names the source-location exception and points at `loadcache.go`'s keying paragraph.
- `TestLoadCacheSharedLiteralIsNotLispObservable` now writes through a DERIVED value: `(append! (slice 'vector lit 0 0) 99)` (the zero-width seal carve-out) and `(append! (slice 'vector (copy lit) 0 1) 99)` (`copy` clears the seal), then re-reads `lit` and `A`. Both are safe today; these are the two shapes that would falsify the classification. The test's comment now states it is a PREMISE test (it passes on `origin/main` too), not a regression test.
- The whitespace-sensitive `assert.Contains(t, cached, "'refused 'refused 'refused")` is replaced by per-element assertions over named probe elements.
- Subtest naming uses `fmt.Sprintf("%s/mode=%d", tc.name, mode)` instead of `string(rune('0'+mode))`, which breaks past mode 9.
- `TestEnvStateFingerprintRecordsMutableAliasing`: an env-level positive control through `envStateFingerprint` itself — the function the fuzz targets call — asserting `(set 'x (vector 1 2))(set 'y x)` fingerprints differently from `(set 'x (vector 1 2))(set 'y (vector 1 2))`.
- `TestValueFingerprintSealedIdentityIsNotState`'s termination arm gains the 2-cell branching sealed cycle and the 26-node sealed DAG, asserted with deterministic NODE-COUNT bounds (`valueFingerprintNodes`) rather than only determinism, plus the hog-blindness case.
- `TestLoadCacheRebindAcrossLoadsMatchesFreshParse` gains an `identical-bytes` case (two different streams, identical content) — the shape that exercises provenance.
- The "differing ONLY in `lit`" claim is softened: the digest is a hash and that was never pinned at byte level.

## Measurements (reproduced on this box, before → after)

| Measurement | Before (PR head `06e990b`) | After (`f466bb1`) |
|---|---|---|
| 26-node sealed DAG (2^25 unfolded) | 25.9 ms | 79 µs |
| Same graph UNSEALED (aliasing-walk reference) | — | 13 µs |
| 2-cell branching sealed self-cycle | 31.4 ms (whole 65536-node budget) | 0.70 ms |
| Hog blindness: `fp([hog, Vector(1)]) == fp([hog, Vector(2)])` | `true` (blind) | `false` (discriminates) |
| `loadCacheKey` mutation (drop `name`+`loc`) caught? | not caught by the fingerprint | caught — `identical-bytes` fails on `eval 2` in all 7 transparent modes |

The `loadCacheKey` mutation was applied and reverted locally to confirm both directions: mutated → `TestLoadCacheRebindAcrossLoadsMatchesFreshParse/identical-bytes` fails in every transparent mode; unmutated → green, and the #613 seed stays green.

## Gates

- `go build ./...` — OK
- `go vet ./lisp/` — OK
- `gofmt -l` on the four touched files — clean
- `golangci-lint run ./lisp/...` (v2.6.2) — 0 issues
- `make elpsvet` (both passes, plain and `-tags=elpscheck`) — OK
- `go test ./lisp -run 'TestLoadCache|TestSealed|TestSlice|TestValueFingerprint|TestEnvStateFingerprint|FuzzLoadCache|FuzzSharedProgram|FuzzSequenceOps' -count=1` — ok 5.7 s
- same with `-tags elpscheck` — ok 17.6 s
- same with `-race` — ok 44.0 s
- `go test ./lisp -count=1` (whole package) — ok 33.2 s

### DIFF
--- a/lisp/loadcache_reader_fuzz_test.go
+++ b/lisp/loadcache_reader_fuzz_test.go
@@ -57,6 +57,23 @@ import (
 // Also asserted, in every mode: termination under the scheduled-time
 // watchdog, and no recovered Go panic (lisp.IsInternalPanic) — both from the
 // shared budgeted harness.
+//
+// "Indistinguishable" means BY WHAT A PROGRAM CAN OBSERVE: every load's
+// value, and every binding's content, sealed-ness, mutable-storage aliasing
+// and sealed-node PROVENANCE (the frozen parse location every error and
+// stack note through that node prints -- see envStateFingerprintProv, which
+// this target uses and its siblings do not).  It does not mean the same
+// storage: a cache hit serves the
+// sealed nodes the miss served (that is the contract, pinned by
+// TestLoadCacheServesTheSameNodes), so a file loaded twice through the cache
+// rebinds its literals to nodes an earlier load already captured, where two
+// fresh parses give two equal nodes.  The environment digest used to record
+// that identity and reported the contract as a divergence (issue #613, seed
+// bfe6dc31a652ced0: A binds a literal, B captures it, A is loaded again);
+// valueFingerprint now digests sealed nodes by content alone (memoised by
+// pointer, so a sealed DAG or cycle stays cheap), and
+// TestLoadCacheRebindAcrossLoadsMatchesFreshParse pins the shape
+// deterministically in every mode.
 
 // loadCacheHostileReader wraps the real parser and then does one plausible,
 // contract-legal thing to what it returns.
@@ -361,7 +378,29 @@ func runHostilePair(t *testing.T, mode uint8, a, b []byte, cache *fuzzLoadCache)
 		if !ok {
 			return programRun{}, false
 		}
-		fp := valueFingerprint([]*lisp.LVal{result})
+		// PROVENANCE variant, scoped to this function.  A cached load must
+		// serve the parse of the file that was ASKED FOR, and a sealed
+		// node's frozen source location is the lisp-observable witness of
+		// which file that was: it is printed by every error and stack note
+		// raised through the node.  A content-only digest cannot see a cache
+		// that serves one file's parse for another's -- mutate loadCacheKey
+		// to drop name and loc, load two byte-identical files, and the
+		// second file's errors name the first file's path while the
+		// content-only digest reports nothing.
+		//
+		// The per-load RESULT is where that shows up, not the env state: for
+		// two byte-identical files the A,B,A sequence rebinds the same
+		// symbols from the same content, so the final environment converges
+		// whichever parse was served.  Load B's own value does not.
+		// (TestLoadCacheRebindAcrossLoadsMatchesFreshParse's identical-bytes
+		// case is the deterministic form; under that mutation it fails on
+		// `eval 2` in every transparent mode.)
+		//
+		// Scoped because FuzzSharedProgramMultiEnv deliberately evaluates one
+		// parse under several stream names; see valueFingerprintProv
+		// (valuefp_test.go) for the three tests a global provenance rule
+		// turns red.
+		fp := valueFingerprintProv([]*lisp.LVal{result})
 		if isAdmissionRefusal(result) {
 			// Tagged rather than merely digested so hostileRefusalOnly can
 			// tell "the cache refused this load" from "the cache answered
@@ -371,7 +410,10 @@ func runHostilePair(t *testing.T, mode uint8, a, b []byte, cache *fuzzLoadCache)
 		}
 		run.results = append(run.results, fp)
 	}
-	run.state = envStateFingerprint(env)
+	// Provenance here too, for the same reason as the per-load digests above
+	// -- a binding that survives to the end of the sequence must hold the
+	// parse of the file it came from.
+	run.state = envStateFingerprintProv(env)
 	return run, true
 }
 
--- a/lisp/loadcache_rebind_test.go
+++ b/lisp/loadcache_rebind_test.go
@@ -1,0 +1,337 @@
+// Copyright © 2026 The ELPS authors
+
+package lisp_test
+
+import (
+	"fmt"
+	"strings"
+	"testing"
+	"time"
+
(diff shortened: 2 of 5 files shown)


sam-at-luther:

The sealed-content/provenance oracle and its DAG/cycle-budget, mutable-alias and original load-cache regression controls are retained in [loadcache_rebind_test.go](https://github.com/luthersystems/elps/blob/v1.61.0/lisp/loadcache_rebind_test.go). This also resolves [#613](https://github.com/luthersystems/elps/issues/613): cached sealed identity is not mutable state, while observable provenance and mutable aliasing remain checked.

PR #616, opened 2026-09-06
Gate Copy's cell-view drop on the header type, and widen the drift scan to the class that let it ship

sam-at-luther:
Stacked on the chain head (`claude/elps-perf-optimization-jcgfd2-loadcache-topology`, b076a2f) so the diff shows only these five commits. Draft: the base is itself unmerged.

Found by an adversarial code review of the chain's production interpreter change — the chain had been fuzzed extensively but never read for defects.

## 1. `(*LVal).Copy` destroys any `LNative` payload whose Go type is `*LVal`

A regression against `main`. `Copy`'s `default:` arm decided on the **payload** type:

```go
if _, isView := v.Native.(*LVal); isView { cp.Native = nil; cp.Int = 0 }
```

`LNative` falls into `default:` (only `LArray` and `LSortMap` have their own cases), so `lisp.NativeOf[*lisp.LVal](x)` — the shape `native.go` blesses for #546 — lost its payload on every copy. `detach`'s twin arm has keyed on the **header** type all along, as do the `*[]byte`, `*MapData` and `*CallStack` arms beside it.

Reachable from ordinary lisp with no Go fixture: `lvalByFun.Less` copies both elements it compares, so `(stable-sort < boxes key)` handed the key function arguments whose payload was gone. `insert-sorted` has the same shape.

The fix is the gate `detach` already had: `if v.IsCellView()`. Commit `aa0dbe4` removed it deliberately — it is right about the view case and wrong about `LNative`; the commit body says which half survives.

| Control | parent `b076a2f` | `main` `a32ada7` | this branch |
|---|---|---|---|
| `TestCopyKeepsNativeLValPayload` | FAIL — `Copy dropped the native payload: <nil>` | PASS | PASS |
| `TestCopyKeepsNativeLValPayloadUnderStableSort` | FAIL — `expected native *lisp.LVal value, got native <nil>` | PASS | PASS |

## 2. The payload drift scan was structurally blind to it

This is why the bug shipped green. `nativeSwitchCaseTypes` recognised only `case *T:` arms inside `switch ….Native.(type)`, in `fork.go` and `detach.go`. The regression's site is a type *assertion*, in `lisp.go` — invisible twice over, even though `Copy` is a registered walker with `Rebuilds: true`. And the `*LVal` `memoExemptions` row was marked "used" only because an unrelated walker happened to spell its arm as a `case`: rewrite that arm as an `if` and the shrink-only checker declares the row dead while fork still handles the payload.

The replacement keys on the **property** — a function that assigns to a `.Native` field — found by type-checking the whole package, so no file list can go stale. A payload counts however it is named: `case *T:`, `….Native.(*T)`, or a call to `IsCellView`/`CellView`/`cellsView`. 13 sites across 5 functions in 3 files.

`TestCellViewLinkIsGatedOnTheHeaderType` **fails on the parent commit**, naming `lisp.go:1729 in (*LVal).Copy (assert)` — the poka-yoke for the class, not the instance. `TestThePayloadScanSeesEveryRebuildingWalker` ties coverage back to the registry.

## 3. Smaller items

- **`detach`'s silent drop of a non-view `*LVal` on an `LSExpr` is correct** — investigated and *not* changed. A stale link is byte-identical to an embedder-set payload and `CellView()` answers false for both; stale links are reachable from ordinary lisp. Reinstating the loud arm fails exactly one test, the new one. The reasoning is now in the comment and measured by `TestDetachDropsAStaleCellViewLinkSilently`.
- **`lvalCopyExemptions` is now genuinely shrink-only**: production rows and fixture rows are separate lists, and a site only matches a row from its own list. It previously carried 5 test rows out of 11, so a production `*cp = *v` in a test-looking function was easy to miss.
- **`scripts/mutation-proof.sh` armed `trap restore EXIT` above its dirty-tree check**, so `die` reverted the uncommitted work it was telling you to commit. It discarded a real edit during this work. Trap moved below the check.

## Reviewable part

The widened scan surfaced one always-present, never-examined payload: fork rebuilds `*funData` per header. That needs a new `memoExemptions` row — I could not both widen the scan and add nothing. It is made falsifiable rather than asserted (`TestForkRebuildsFunDataPerHeaderIndistinguishably` reddens if a `funData` field becomes mutable or `env` stops being memoised), but **a reviewer should decide whether fork ought to memoise `*funData` instead**, which would be strictly better and would delete the row.

Also honest about a limit: the gate check is syntactic, not dataflow. A function mentioning `LSExpr` for an unrelated reason that also asserts `*LVal` off `.Native` would pass. It catches the class as it appears in this package and fails on the actual regression.

## Verification

`go build ./...`, `go test ./lisp/...`, `-tags elpscheck`, `-race`, whole-repo `go test ./...`, `make elpsvet` (both passes), `scripts/mutation-proof.sh` 10/10 caught, `mutation-proof-selftest.sh` all three guarantees. No gate relaxed; no needle silently stopped tripping.

Not touched: `(*LVal).Copy` stays on `knownDefectiveWalkers` for its separate, still-open `*MapData` and `*[]byte` defects.

---

### DIFF
--- a/lisp/copy_native_lval_test.go
+++ b/lisp/copy_native_lval_test.go
@@ -1,0 +1,133 @@
+// Copyright © 2026 The ELPS authors
+
+package lisp_test
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/luthersystems/elps/lisp"
+)
+
+// An LNative whose payload happens to be a *LVal is a payload like any
+// other -- lisp.NativeOf[*lisp.LVal] is a spelling the kernel blesses at
+// native.go's NativeOf (issue #546), and NativeValue[*lisp.LVal] is the
+// documented way to read it back.  It is NOT a cell view: a cell view is an
+// LSExpr whose Native names the header its Cells window (the convention on
+// cellsView), and cellsView itself gates on that header type before it
+// looks at the payload.
+//
+// Copy briefly keyed "this link is a cell view" on the payload type alone,
+// so every one of these natives came back from Copy with a nil payload.
+// These tests pin the header-type gate that detach.go's twin arm always
+// had.
+
+// TestCopyKeepsNativeLValPayload is the unit form: the payload survives a
+// Copy, and the copy still reads back through the typed accessor.
+func TestCopyKeepsNativeLValPayload(t *testing.T) {
+	payload := lisp.QExpr([]*lisp.LVal{lisp.Int(1), lisp.Int(2)})
+	v := lisp.NativeOf[*lisp.LVal](payload)
+	if v.Type != lisp.LNative {
+		t.Fatalf("NativeOf built a %v, want %v", v.Type, lisp.LNative)
+	}
+
+	cp := v.Copy()
+	if cp.Type != lisp.LNative {
+		t.Fatalf("Copy produced a %v, want %v", cp.Type, lisp.LNative)
+	}
+	if cp.Native == nil {
+		t.Fatalf("Copy dropped the native payload: %#v", cp.Native)
+	}
+	got, ok := lisp.NativeValue[*lisp.LVal](cp)
+	if !ok {
+		t.Fatalf("NativeValue[*lisp.LVal] on the copy: not ok, payload %T", cp.Native)
+	}
+	if got != payload {
+		t.Errorf("Copy replaced the payload: got %p, want %p", got, payload)
+	}
+	// Copy shares native leaves by identity (TestCopySharesFunctionAndNativeLeaves),
+	// so Int -- which the cell-view convention uses as the view's offset --
+	// must not be zeroed either.
+	if cp.Int != v.Int {
+		t.Errorf("Copy changed Int: got %d, want %d", cp.Int, v.Int)
+	}
+}
+
+// TestCopyKeepsNativeLValPayloadUnderStableSort is the reachable form, and
+// the reason this matters outside a unit test: no Go code has to call Copy
+// for the payload to vanish.  stable-sort's comparator copies both
+// elements it compares (lvalByFun.Less), so ordinary lisp sorting a list
+// of these natives hands the key function arguments whose payload is gone
+// -- surfacing to an embedder as "expected native *lisp.LVal value, got
+// native <nil>".  builtinInsertSorted copies the same way, so insert-sorted
+// is exercised alongside it.
+func TestCopyKeepsNativeLValPayloadUnderStableSort(t *testing.T) {
+	env := copyTestEnv(t)
+
+	// `boxed-rank` is the embedder-side reader: it asks for the payload
+	// through the typed accessor exactly as RequireNative documents, and
+	// reports the failure the way an embedder would see it.
+	var seen, nilPayload int
+	rank := lisp.FunInPackage(lisp.DefaultUserPackage, lisp.DefaultUserPackage+":boxed-rank", lisp.Formals("x"),
+		func(env *lisp.LEnv, args *lisp.LVal) *lisp.LVal {
+			seen++
+			box := args.Cells[0]
+			inner, lerr := lisp.RequireNative[*lisp.LVal](box)
+			if lerr != nil {
+				nilPayload++
+				return lerr
+			}
+			return inner.Cells[0]
+		})
+	if lerr := env.PutGlobal(lisp.Symbol("boxed-rank"), rank); lerr.Type == lisp.LError {
+		t.Fatalf("PutGlobal boxed-rank: %v", lerr)
+	}
+
+	box := func(n int) *lisp.LVal {
+		return lisp.NativeOf[*lisp.LVal](lisp.QExpr([]*lisp.LVal{lisp.Int(n)}))
+	}
+	env.PutGlobal(lisp.Symbol("hi"), box(2))
+	env.PutGlobal(lisp.Symbol("lo"), box(1))
+
+	sorted := env.LoadString("copy_native_lval_test.lisp",
+		`(stable-sort < (list hi lo) boxed-rank)`)
+	if sorted.Type == lisp.LError {
+		t.Fatalf("stable-sort over *LVal-payload natives failed (%d/%d arguments arrived with a nil payload): %v",
+			nilPayload, seen, sorted)
+	}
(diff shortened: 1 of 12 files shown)


sam-at-luther:
## Fuzz shard 8/11 red on `6dd1252` — what I know, and what I can't yet establish

`FuzzSharedTreeEval` found a new crasher, `423b7dd9e421bd27`, in 13s. Recording this rather than re-checking silently, because I have **not** been able to attribute it either way.

**I am not dismissing it as someone else's.** This PR changes `lisp/lisp.go`, and the change makes `Copy` *retain* `*LVal` native payloads it previously destroyed. A payload that now survives a copy is exactly the kind of thing that can surface a latent crash downstream, and `FuzzSharedTreeEval` is in the same package. So it is plausibly mine and I treated it that way.

### What I did

| Attempt | Result |
|---|---|
| `FuzzSharedTreeEval`, 150s × 2 workers, on parent `b076a2f` | no crasher |
| same, on this branch `6dd1252` | no crasher |
| `FuzzSharedTreeEval`, **720s × 2 workers**, on parent `b076a2f` | no crasher |
| same, on this branch `6dd1252` | no crasher |

24 minutes of directed fuzzing across both arms, from the same committed seed corpus, reproduced nothing on either side. The only seed present in `lisp/testdata/fuzz/FuzzSharedTreeEval/` on both trees is `cfba74b0f2cbb127`.

### What blocks a definitive answer

The crasher input exists only as a CI artifact (`fuzz-crashers-34042739278-shard8`). I have no way to download a workflow artifact from this environment — the tooling available to me lists artifacts but cannot fetch them — so I cannot seed the exact input and get a deterministic answer. That is the blocker, and it is mine to state rather than to work around.

### Relevant context on the base

The base branch is not quiescent under fuzzing. At the **identical commit** `b076a2f`:

- the `pull_request` Fuzz run ([33949925959](https://github.com/luthersystems/elps/actions/runs/33949925959)) **passed**;
- a longer `workflow_dispatch` Fuzz run ([33950054036](https://github.com/luthersystems/elps/actions/runs/33950054036)) **failed** — shard 11/11.

So the chain's fuzz targets do surface fresh crashers on the base itself. That is context, **not** an alibi: the base's failure was a different shard, so it does not cover shard 8.

### What happens next

I have re-run the failed jobs once — my single permitted re-run, spent here to see whether the same target reddens again on this head. If it does, the input is reachable enough to chase and I will chase it. If it does not, that still leaves the question open rather than closed, and the honest resolution is to obtain the artifact — which needs someone with the means to download it, or a corpus seed committed to the branch.

I am keeping this PR watched until it is green or the crasher is understood. I will not mark it ready while this is unresolved.

---

sam-at-luther:
## Resolved: the crasher is the base's, and it is fixed in `5fb298a`

Correcting my previous comment. I said the artifact was unobtainable — **the input was in the job log all along**, in the failure body I had not paged back far enough to read:

```lisp
(defun s()(let()(s)))(s)(s)(s)
```

| arm | error |
|---|---|
| private tree | `tail-call iteration limit exceeded: 100001` |
| shared tree | `context-cancelled: context deadline exceeded` |

Both arms are the same runaway tail-recursive loop. They differ only in **which budget stopped it** — one denominated in turns, one in time. The shared arm skips a parse and so runs at a different speed, and under load the two arms disagree about which limit tripped first. The comparison read that as "evaluating a SHARED parse tree changed the program's meaning".

### Attribution, measured

Seeding the reconstructed input on `b076a2f` — the chain head this branch is based on, which predates every commit here:

| | sequential (5 runs) | concurrent (8 runs) |
|---|---|---|
| parent `b076a2f` | 5/5 pass | **8/8 FAIL** |
| this branch (before the fix) | 5/5 pass | — |
| this branch (with `5fb298a`) | 5/5 pass | **8/8 pass** |

The failure reproduces on the base, which does not contain the `Copy` gate. It is **not this PR's**. Eleven fuzz shards on one runner is the load that surfaces it.

### The fix, and why it is here

Collapse the two budgets into one outcome — **only when both arms hit one**. One arm terminating while the other is stopped by a budget stays a divergence and is still reported; that asymmetry is not explained by scheduling.

It is fixed in this PR rather than left for the chain because it reddens this PR on every run and the chain is currently unowned. The change is confined to the target's own test file and no-ops if the chain fixes it first.

`sharedtree_backstop_test.go` pins the narrowness in both directions — including the case that must **not** collapse: a program is free to raise an error whose text mentions a limit, and that is its meaning, not a backstop. A fix that quietly narrows an oracle would be worse than the flake it removes.

Verified: `go test ./lisp/...`, `-tags elpscheck`, `make elpsvet` all clean.

---

sam-at-luther:

The cell-view payload-drop regression is obsolete because the advisory metadata was removed; [legacy Copy/native-sort controls](https://github.com/luthersystems/elps/blob/v1.61.0/lisp/template_legacy_regression_test.go) preserve the surviving behavior. The broader whole-package Native-write drift scan and shrink-only struct-copy exemptions are not in the release and remain relevant with [#599](https://github.com/luthersystems/elps/pull/599) / [#598](https://github.com/luthersystems/elps/issues/598). If that mutation-proof infrastructure is retained, preserve this PR's dirty-tree cleanup-trap safety fix too; the script itself is absent from the released tree.

sam-at-luther:
Closing: subsumed by the template work in #623 and by #642.

Five of the six commits repair a bug that the #599 stack itself introduced (`aa0dbe4`'s `v.Native.(*LVal)` cell-view gate, which never reached main) or edit scans that exist only on that stack (`lisp/walkers.go`'s payload scan, `lvalcopy_drift_test.go`, `scripts/mutation-proof.sh`). Main already carries the two properties `6314a95` tested: `TestTemplateLegacyCopyKeepsNativeLValPayload` and `TestTemplateLegacySortKeepsNativeLValPayload` in `lisp/template_legacy_regression_test.go`, and both pass unmodified there. `fd3a75b` comments a `*LVal` arm in `detach` that main does not have.

Kept: `5fb298a` (the `FuzzSharedTreeEval` runaway-budget flake fix) is in #642 as its own commit. `e1dd503`'s idea, keying the payload drift scan on "a function that assigns to a `.Native` field" rather than a file list, is carried in #642's registry commit.

---

PR #617, opened 2026-09-06
Close the demonstrated blind spots in the isolation/parity oracle

sam-at-luther:
Stacked on the chain head (`claude/elps-perf-optimization-jcgfd2-loadcache-topology`, b076a2f) so the diff is only these seven commits. Draft: the base is itself unmerged.

From an adversarial review asking not "does the oracle pass" but **"what real bug could it miss"**. It found 8 misses with runnable proofs. Its own summary of the shape: the oracle is strongest exactly where elps's historical bugs were — pointer-held payloads reached from lisp source — and weakest exactly where substrate's risk lives, **Go-side payloads**.

## Commits

| | Finding closed |
|---|---|
| `79391de` | Census a native payload held by any reference kind, not only a pointer |
| `60d92db` | Call a native payload stateless only when its own type is basic |
| `27b0320` | Let a caller render a native payload's contents into both parity channels |
| `0aab15a` | Probe a header's cell slot, and both ends of a bytes buffer |
| `837375f` | Normalise the two ID counters where they are IDs, not wherever they match |
| `a667f95` | Replace the wrong reason a builtin's implementation is unencoded with the measured one |
| `8239420` | Check fork independence on the concurrent arm, not just the template |

## The three that composed

`isPointerPayload` keyed on `reflect.Pointer`, so a payload of kind **Map, Slice, Chan or Func** was invisible to `reachableNatives` (property 4 could not fire), to `NativeDeclarations` (the exported pre-ship census an embedder is told to run), *and* to the fingerprint. Separately, `isStatelessPayload` unwrapped one pointer before checking for a basic type, so a `*int` — a shared mutable cell — reported `stateless=true`. And no channel ever compared a native's **contents**, so parity, the backstop, went silent whenever the observation was itself a native.

Together: a map- or func-typed handle shared by every transaction was invisible to the census, invisible to the fingerprint, and invisible to parity *even when a transaction observed it*. The substrate-relevant instance is a func payload — a closure over mutable Go state, which presents as `reflect.Func`.

Independently verified for this PR, with a test written outside the branch: on the parent the census reports **0** shared payloads for a map both forks demonstrably write through; on this branch it reports **1**.

The widening documents its own limits rather than overclaiming: a zero-length, zero-capacity slice has no identity, and Go may emit one static funcval for non-capturing literals. The ordinal is keyed on a `uintptr`, not the payload value — a map- or slice-typed payload used as a Go map key would otherwise panic.

## One fix implemented, measured worthless, and reverted

Adding `reflect.ValueOf(v.Builtin()).Pointer()` as a builtin identity ordinal **does not work**: every builtin is registered as `LBuiltinDef.Eval`, a method value, so reflect returns the compiler's shared wrapper address and `car` and `cdr` collapse to the same ordinal. A token asserting "the whole stdlib shares one implementation" is worse than no token. An absolute pointer is not an option either — a closure builtin gets a fresh address per environment, so cold-vs-fork would redden on correct code.

What shipped instead is the wrong justification replaced with the two measured ones, plus `TestABuiltinIsNotIdentifiedByItsGoPointer`, which reddens the day reflect starts distinguishing them. **Closing the absolute-implementation half needs a design decision** — it needs unsafe (the eface data word), which has no precedent in this repo.

## New finding, pinned rather than fixed

The last-byte probe made visible that **Fork, `copy` and detach all de-alias an overlapping bytes view**: `(slice 'bytes b 3 6)` shares `b`'s array on a cold load and comes out of every rebuild with its own. For *cells* that is a Fork contract (#602) and a documented exemption for the other two; for **bytes nothing states either**, so a fork diverges from a cold load. It is the bytes analogue of #600 gap 3. `TestTheBytesProbeSeesAnOverlappingView` records it — green today, and it fails the day someone fixes it, at which point the fix is to state the contract, not to weaken the probe.

## Residual, narrowed and documented

An error's rendering is still normalised as free text — an error is a message, not a structure, so a FID or gensym reaches it through frame names with nothing to key on. A divergence confined to an *error message* matching either pattern is still erased. Narrowed from every value to that one case, and said at the call site.

## Gates

`go build ./...`; `go test ./elpstest/... ./lisp/...`; `-race ./elpstest/...`; `make elpsvet` and `make static-checks` both passes; `scripts/mutation-proof.sh` **10/10 caught with no needle changed**; `mutation-proof-selftest.sh` all three guarantees. Each of the seven commits builds and passes `./elpstest/` independently.

Two gate touches, neither a relaxation: `TestAProbeCountEqualToTheCapIsNotReportedAsPartial` now carries a cap equal to each graph's *real* site count with the arithmetic stated (the graphs genuinely hold more mutable payloads now, and "one over the cap" became the same graph at `cap-1` rather than a bigger graph); and the cell-slot write/undo carries an audited `//elps:mutates` with justification, the same hatch `elpstest/sortedmap.go`'s probe already uses.

`ProbeCellSlot` is excluded from the **alias-class** comparison for `WalkerCopy` — copy and detach declare they do not preserve cells backing-array sharing, and without the exemption both went red on documented behaviour. The leak check still covers every cell-slot site of every walker, which is what catches an identity copier.

## Not in this PR

Deliberately held: measuring cell-slot overlap independently instead of trusting the `cellsView` declaration (an undeclared window still reproduces the exact #602 signature invisibly); fuzzing with `lisplib` loaded; `Runtime`-level shared state (`Reader`, `Library`) which no walk reaches; and the arms that have no control. Those are design changes, not repairs.

---

### DIFF
--- a/elpstest/aliasguard.go
+++ b/elpstest/aliasguard.go
@@ -265,6 +265,49 @@ const (
 	// one would report a documented behaviour as a leak -- and would be a
 	// write into storage the guard does not own, which is the very defect
 	// class checkStamp exists for.
+	//
+	// ARRAY HEADERS ARE SKIPPED TOO, for a different reason: an LArray's
+	// own Cells is not the storage a vector is mutated through.  It is a
+	// fixed two-slot structural record -- [0] the dims list, [1] the data
+	// holder (lisp/lisp.go's LArray doc) -- and every writer of a
+	// vector's ELEMENTS goes through seqCells/seqHolder, which for an
+	// LArray hand back Cells[1].Cells, the HOLDER's array.  Nothing in this
+	// repository assigns an array header's own slot: cmd/elpsvet's alias
+	// rule forbids `v.Cells[i] = x` on a header the function did not
+	// construct, and the only such assignment reachable from a vector
+	// operation is the probe write below.  So a site here measures an alias
+	// no writer can exploit -- and it COSTS a false positive.
+	//
+	// lisp.Quote's struct copy gives `(quasiquote (unquote vec))` a second
+	// LArray header over the SAME two-slot array (lisp/walkers.go's
+	// lvalCopyExemptions: the intentional aliasing).  Fork memoises per
+	// *LVal and per registered payload kind, and a Cells backing array is
+	// neither, so the fork lands three private two-slot arrays whose SLOTS
+	// hold the same dims header and the same holder header.  Measured on
+	//
+	//	(set 'v1 (vector 3 1 2))
+	//	(set 'v2 (quasiquote (unquote v1)))
+	//	(set 'v3 (quasiquote (unquote v1)))
+	//	(set 'probe (list v1 v2 v3))
+	//
+	// which FuzzAliasGuard found (crasher cbfd49dfda179e65): the sweep
+	// called Fork's rebuild a leak, and EVERY mutation the language can
+	// perform through any of the three names -- stable-sort, append!,
+	// append! past capacity -- agrees between a cold load and a fork.
+	// TestAnArrayHeaderIsNotAProbeSite and
+	// TestAQuasiquotedVectorForksLikeAColdLoad pin the two halves.
+	//
+	// Fork's ONE cells-sharing contract is the cell VIEW, and a view is an
+	// LSExpr: lisp.cellsView returns nothing for any other type, so no view
+	// is lost here.  It is asserted by cellViewWitnesses
+	// (aliasguard_cellview.go) as well as by the sweep.
+	//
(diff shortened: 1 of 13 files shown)


sam-at-luther:

The replacement independently tests native reference kinds, caller-provided native contents in both comparison channels, actual cell slots/byte overlap, user identifiers and concurrent hidden writes. See [oracle visibility controls](https://github.com/luthersystems/elps/blob/v1.61.0/elpstest/forkcheck_visibility_test.go) and [parity negative controls](https://github.com/luthersystems/elps/blob/v1.61.0/elpstest/template_fuzz_test.go). Opaque Go closure identity remains an explicit limit, not a claimed proof. Broader copy/detach class-guard work remains separately open in [#599](https://github.com/luthersystems/elps/pull/599).

PR #642, opened 2026-09-09
Consolidate the pre-template fixes that survive v1.61.1

### LINKED ISSUE #606: LQSymbol has no rendering arm: String() falls through to %#v and prints heap addresses

### LINKED ISSUE #607: LSP: DocumentStore.Open/Change parse the whole document before the size gate; a large didChange hangs the session

### LINKED ISSUE #611: LSP: the didSave path re-reads and parses the saved file with no size check (same class as #607)

### LINKED ISSUE #643: Validate custom-map keys before Copy invokes native callbacks
Fourth-pass review of #642 found a remaining error-path ordering gap. A valid custom Map with distinct integer keys and distinguishable NativeCloner values may emit Entries in either order. Sorting by (Str, Type) ties those keys, and Copy invokes a value callback before the stock destination rejects its unsupported key. Both copies fail identically but invoke different callbacks. Validate unsupported key types before copying their values, preserve fail-stop behavior and error wording, and add a rotating-entry regression test with a supported-key positive control. This is not a template-isolation failure.

sam-at-luther:
## Summary

Every open non-dependabot PR was triaged against main after #623/#640 (immutable templates). This PR carries the parts that still apply, rebased onto v1.61.1 as one isolated commit per work item. The rest was subsumed and is closed with a reason on each PR. The three open dependabot bumps ride along as their own commits.

| Source | Kept as | Notes |
|---|---|---|
| #615 | 1 commit | LQSymbol rendering. Taken as reviewed. Closes #606. |
| #609 | 1 commit | LSP parse size bound. Folded in from review: four more handlers (definition, completion, signatureHelp, foldingRange) were still linear on an over-limit document; one `Document.OverLimit()` helper replaces the three copied lock blocks; the test now asserts cost (64 KiB ceiling on an 8 MiB doc), not only emptiness; the godoc sentence claiming didSave is unbounded is gone because the next commit bounds it. Closes #607. |
| #612 | 1 commit | LSP didSave size bound. Only its own commit; the merged-in #613 commit is dropped. Folded in from review: an over-limit save now purges that file's stale index entries instead of leaving definitions it no longer has. The unbounded path was super-linear, not "a full parse". Closes #611. |
| #616 | 1 commit | Only the `FuzzSharedTreeEval` flake fix. The other five commits repaired a bug the #599 stack introduced and main already carries the equivalent regression tests. |
| #604 | 6 commits | `(*LVal).Copy` memoisation, macro-expansion drop, LArray walk, comparator-by-reference, allocation pin, load-path memo pre-size. All five Copy defects were measured live on main. Registry and mutation-gate hunks dropped; test helpers re-homed out of `elpstest`. |
| #599 | 4 commits | Walker memo registry with drift guard (rows for `templateInventory`, `templateCompiler`, `copier`), the LVal struct-copy scan (three template-compiler sites audited), the Copy-path sort benchmarks, and the `forkcheck.go` doc fix removing a false #579 citation. The fingerprint oracle and mutation gate are not ported; that remainder is tracked on #598. |
| #605 | 2 commits | Native-payload elpsvet rule. Machinery and fixtures as reviewed; rules re-audited: the NativeCloner exemption is replaced by the `templatepolicy.Marker` struct-value tier mirroring `templateInventory.native`; the `time.Time`, `*regexp.Regexp`, `*CallStack`, `error`, `*ownMessage` and `validatorTag` rows are gone (each was refuted by a merged test or is handled by the runtime); the diagnostic names the three real remedies. Wired into `make elpsvet` and the gate self-test. |
| gate fix | 1 commit | The first CI run failed the benchmark gate on the load path (`TextLoaderLoad` +15.2% time / +8.9% bytes, the 78 KiB `reader-cache-copy` arm +8.6% bytes): the memoised Copy recorded every header, and a parse tree is mostly leaves. Copy now memoises only containers and payload headers. A leaf cannot cycle, cannot share storage, and its identity is unobservable from Lisp, which is also what main's un-memoised Copy did. Admission's hint counts the same set. Measured against main: bytes and allocations on all three rows are identical, timing unchanged. |
| #618 | 1 commit | dependabot: vscode-languageclient 10.1.0 to 10.1.1 in `editors/vscode`. Cherry-picked as authored. |
| #636 | 1 commit | dependabot: js-yaml 4.3.1 to 4.3.2 in `editors/vscode`. Cherry-picked as authored. |
| #619 | 1 commit | dependabot: tree-sitter-cli 0.26.13 to 0.27.0 in `tree-sitter-elps`. Cherry-picked as authored. |

The three dependabot PRs stay open; dependabot closes them itself once main carries the same versions. `npm ci` passes in both `editors/vscode` and `tree-sitter-elps` on the final head, so each lockfile agrees with its manifest.

## Adversarial review follow-up, round one (seven commits, one per finding)

The first review found six defects and one API-visibility issue in the bundle above. Each is fixed in its own commit with a regression test that was red before the fix and a negative control. The template engine is untouched. The full mapping with proofs is in the first reply comment.

| # | Finding | Commit |
|---|---|---|
| 1 | Failed `Copy` left a memoised header over the source's `*MapData` | 249e616 |
| 2 | Over-limit closed-file rename emitted byte columns as UTF-16 | 9bc273a |
| 3 | `FuzzSharedTreeEval` collapsed results by rendered text | 74f0888 |
| 4 | Native-payload analyzer admitted `uintptr`, `[]**LVal`, and kernel-slot rows through constructors | cb5fd5c |
| 5 | Sorted-map copy invoked host clone hooks in Go map order | 0d3ece9 |
| 6 | Code actions bypassed the over-limit document guard | b983eb4 |
| 7 | Walker registry exported an API with no external caller | 02349a8 |

## Adversarial review follow-up, round two (five commits)

The second review found four gaps left by round one and asked for benchmark coverage. The full mapping with proofs, and the load-path bisection, are in the second reply comment.

| Gap | Commit | Fix |
|---|---|---|
| Cyclic Copy failure could return a container holding a half-built map | c79cec5 | The copier is fail-stop: the first failing arm's error is what `Copy` returns, nothing partial escapes |
| A program could capture the tail-budget error, diverge, and replay it to fool the fuzz oracle | 27164a7 | No value-based classification at all. The tail budget is deterministic across arms; an expired harness context makes the input inconclusive from that expression on |
| `.Native` field writes and `LVal{Type: LNative, Native: row}` literals were exempt from the native-payload rule | 5687def | A row exempts only a site inside package `lisp` whose literal `Type:` key names the row's header; everything else is reported, and the paired runtime test covers both spellings |
| A host `Map`'s `Entries` order reached clone hooks unsorted | cfdff7d | Pairs sorted by (key text, key type) before any value is copied; every arm skips the sort when all values are leaves |
| No benchmark exercised Copy over a sorted map | 7389cf2 | `BenchmarkCopySortedMap` in scalar, nested and custom-map shapes at 64 and 512 keys |

## Adversarial review follow-up, rounds three and four (four commits)

The third review found three small gaps; the fourth found one error-path ordering gap and fixed it directly on the branch. The full mapping with proofs, and the customer-phylum benchmark, are in the third reply comment.

| Gap | Commit | Fix |
|---|---|---|
| A host `Map` with all-scalar values could copy differently by entry order (`"a"` and `'a` collapse to whichever came last) | 09814f4 | The `Entries` arm sorts on every copy; a text collision between string-like keys fails the copy naming the key |
| The native-payload rule accepted a function-local `const LBytes LType = LNative` as the header constant | 73e86dd | The `Type:` constant must be the package-scope object of package `lisp`, checked by identity |
| The shared-tree fuzz harness could become always-inconclusive with every test green | 37dbfaa | `sharedTreeProperty` returns a per-arm verdict; a live-path control and the seed corpus assert conclusive comparisons; three mutations each turn it red |
| With unsupported (non-string-like) keys the sort ties on empty text, so a value's clone hook could run before the stock map rejected its key (#643) | 0a0476b | Keys are validated before any value is copied; rejection is order-independent and no hook runs for a rejected entry. Closes #643 |

Cost notes. The leaf fast path removes the sort's allocation on scalar maps. The residual load-path timing (about +3 to 5% on the load rows in CI's gate runs, allocations identical) was bisected to the `Copy` memoisation's bookkeeping, not to hashing, and is left as the price of a memoising `Copy`; the second reply comment has the method and the reasons the obvious loader-hint fix is unsafe. On a real 4.7 MB customer phylum through substrate (elps main vs candidate 5687def, substrate held constant, ten interleaved rounds under SCHED_FIFO), template publication, per-request checkout, retained heap per VM and the phylum's own test-suite wall clock are all within noise, with allocation counts identical on the checkout path; the third reply comment has the table. The commits after that candidate touch only generic-map copying and the static analyzer, not publication or NewVM.

## Behaviour change, isolated in one commit

"Pass sort comparator arguments by reference instead of copying per comparison" changes two things: a `stable-sort`/`insert-sorted` predicate now receives the list's own elements (as `map` and `foldl` already do), and a predicate that writes through an argument that is a sealed program literal raises `modify-literal-error` instead of silently sorting a throwaway copy. The commit message carries the census (zero affected call sites in two customer phylum snapshots). Drop that commit and the allocation-pin commit after it if you do not want this in this release.

## Verification on the final head (0a0476b)

- `go build ./...`, `go vet`, `gofmt` clean on every touched file
- `go test ./... -count=1`: all packages ok
- `go test -tags elpscheck ./lisp/ ./elpstest/ ./lisp/lisplib/...`: ok
- `go test -race ./lisp/ ./lsp/ ./elpstest/`: ok, no data races
- `go run ./cmd/elpsvet -test=false ./...` in default and `elpscheck` modes: exit 0, no site needed a new annotation
- `scripts/ci-gates-test.sh`: 394 passed, 0 failed
- Drift guards proven to fail on a planted unregistered memo field and a planted struct copy; the struct-copy scan classifies `copier.go`'s copy as a registered-walker site
- Over-limit handler cost test proven to fail with the guards reverted (code actions included); stale-index purge test proven to fail with the purge removed
- Each review-finding test proven red on the previous head and green after its commit, with per-half negative controls and mutation controls where the finding was about a test's own power
- `FuzzSharedTreeEval` smoke on the final oracle, no new crasher
- CI on 0a0476b: 25/25, benchmark gate 0 rows at or above threshold (timing 15%, allocation 5%)

Closes #606
Closes #607
Closes #611
Closes #643

Supersedes #599, #604, #605, #609, #612, #615, #616.

https://claude.ai/code/session_01BHMPMnEJgGRuJioE5ddat6


sam-at-luther:
# Adversarial review of ELPS #642

**Verdict: request changes.** The immutable-template engine has not been reverted, and I found no new defect in its publication/NewVM implementation. However, this bundle introduces two concrete production regressions and weakens a fuzz comparison; those should be fixed before merge.

Reviewed head: `5664864069204c5c6885b61c9218b5a6d4b2e223`, against v1.61.1 / `03e7c95e3e46339c9a504c53a680adb468649592`. All **25/25 current-head checks pass**. The findings below come from additional synthetic, targeted probes, not from treating passing CI as proof.

## Findings

| Priority | Finding | Classification |
|---|---|---|
| P1 | Failed custom-map Copy leaves an alias pointing at the source | New production ownership regression |
| P1 | Closed-file rename can emit byte offsets as UTF-16 edits | New production source-corruption regression |
| P1 | Resource-budget normalization hides differing successful values and user errors | New fuzz-oracle blind spot |
| P2 | Native-payload analyzer exempts constructions runtime admission rejects | New static-check gaps; runtime remains fail-closed |
| P2 | Deep map copying invokes native clone hooks in unspecified map order | New host-callback determinism risk |
| P2 | Code actions still process oversized documents | Preexisting path omitted from this hardening |

### 1. Failed Copy retains source-owned map storage

[Memo/error path](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/lisp/copier.go#L299), [failed map return](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/lisp/copier.go#L313).

The new header memo is seeded while `cp.Native` still points at the source MapData. If a custom `StringKeyRanger` fails, the first encounter returns an error but leaves that unfinished header in the memo. A second encounter returns the unfinished copy.

Reproduction: construct a custom map whose `RangeStringKeys` calls its callback once and returns an error. Copy `QExpr([]*LVal{src, src})`. The result is effectively `[error, source-sharing-map]`. Writing `k=99` through the second element changes the original map from `k=1` to `k=99`.

The targeted probe fails on this head, including checked/race builds. Restoring only the base public Copy implementation makes it pass.

**Fix:** make failed header/payload memo entries safe and propagate or consistently memoize failure; never return the partially initialized source-backed destination. Cover repeated headers, distinct headers over the same failing payload, and failure after partial traversal. Removing one memo entry alone needs care around cycles/backreferences.

This is a demonstrated Go custom-map/Copy boundary defect, **not** a demonstrated cross-VM escape through NewTemplate.

### 2. Oversized closed-file rename corrupts UTF-16 edits

[New bounded read](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/lsp/position_encoding.go#L356), [unconverted-range fallback](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/lsp/position_encoding.go#L366).

Index a small closed file containing `(list "é" (target))`. Append only comments until the file exceeds the workspace size cap, then rename `target` from another open file before a watcher refresh. The first line and indexed reference have not moved.

Before growth, the correct UTF-16 range is **[11,17)**. On this head, the bounded read fails and the rename emits **[12,18)**: applying that edit replaces `arget)`, corrupting source.

The real-index probe fails on this head. Overlaying only the base `position_encoding.go` makes it pass.

**Fix:** retain the bounded read, but fail the whole rename explicitly if required coordinate conversion cannot be performed. Do not emit byte columns as UTF-16, and do not silently perform only part of a rename. Add a non-ASCII cross-file rename regression where the closed file grows after indexing.

### 3. The fuzz exception is based on arbitrary rendered text

[Comparison](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/lisp/sharedtree_fuzz_test.go#L266), [substring classifier](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/lisp/sharedtree_fuzz_test.go#L367).

`bothHitAResourceBackstop` returns true for these two **successful string values**:

```text
"context deadline exceeded: original"
"context deadline exceeded: corrupted"
```

It also suppresses differences between user-raised errors containing that phrase. Both targeted negative controls fail, including checked/race builds. The existing forged-error test uses `rate limit exceeded`, which never exercises either actual substring trigger.

**Fix:** preserve structured evaluation/budget provenance before rendering; only normalize genuine evaluator-budget outcomes. Keep successful values, user errors, and one-budget/one-result outcomes exact, even when their text contains either trigger. The flake motivation is valid, but this implementation enlarges the set of invisible divergences.

### 4. Static admission still differs from runtime admission

[Basic-type classification](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/cmd/elpsvet/nativepayload.go#L432), [Value slice matching](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/cmd/elpsvet/nativepayload.go#L538).

The actual built analyzer emits no native-payload diagnostic for any of:

```go
lisp.Native(uintptr(1))
lisp.Value([]**lisp.LVal{})
b := []byte{1}
lisp.Native(&b)
```

The actual runtime rejects all three at publication without a native policy.

- `uintptr` is treated as safe because all basic kinds except unsafe.Pointer pass. Runtime explicitly excludes uintptr.
- `[]**LVal` matches the `[]*LVal` fast path because the helper strips another pointer layer. It actually falls through Value to an opaque native.
- `*[]byte` is exempted by payload type alone, even when constructed as LNative. Runtime's LNative arm calls native admission, not the LBytes storage copier. The same distinction matters for other kernel-slot exemptions.

**Fix:** mirror the exact admitted scalar kinds and exact Value conversion types; restrict kernel-slot exemptions to proven representation sites, or annotate those few audited internal stores instead of globally exempting their payload types. Add real-analyzer-versus-real-runtime negative controls. Do not weaken runtime admission to make the analyzer match.

This does **not** bypass runtime fail-closed validation, but it undercuts the promised CI forcing function and can allow unnoticed fallback/performance degradation.

### 5. Map traversal now exposes unspecified order to host callbacks

[Sorted-map value copying](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/lisp/copier.go#L375), [custom ranger copying](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/lisp/copier.go#L398).

Copy now invokes `NativeCloner` while iterating Go map entries. A synthetic cloner that does not mutate its receiver but consumes an external sequence counter assigns a different clone value to key `a` on repeated Copy of identical input with the counter reset. Restoring the old Copy makes the probe stable because that path did not call the hooks.

The published contract forbids receiver mutation but does not require state-independent hooks. This is a host-extension risk, not evidence that ordinary pure-Lisp map copying is nondeterministic.

**Fix:** use deterministic key order before invoking host hooks, including custom-map paths, or explicitly narrow the clone contract and enforce the supported behavior. Benchmark any sorting cost rather than assuming descriptor-compilation's pure map iteration argument applies to callback execution.

### 6. Code actions bypass oversized-document protection

[Handler](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/lsp/code_actions.go#L30), [whole-document split](https://github.com/luthersystems/elps/blob/5664864069204c5c6885b61c9218b5a6d4b2e223/lsp/code_actions.go#L234).

Ordinary client sequence: open a small document, receive a real lint diagnostic, round-trip it through JSON, grow the document to 8 MiB before diagnostic debounce, then request the stale diagnostic's code action.

Despite `OverLimit()==true`, the request allocates **67,118,464 bytes** and returns an edit. No forged diagnostic is required.

This route predates the PR, but it defeats the claimed handler size bound. The new cost test omits code actions.

**Fix:** guard code actions consistently and include them in the behavioral allocation test, using a genuine stale diagnostic.

## Did it undo the template fixes?

| Earlier protection | Review result |
|---|---|
| Template admission, runtime snapshot, storage overlap/index and checked sorted-span guard | No production changes |
| Per-VM instantiation | Only an analyzer justification comment added |
| Private immutable time/regexp wrappers and raw-host rejection | Preserved |
| Template parity fuzz implementation | Unchanged |
| Shared-tree parity guard | Weakened by finding 3 |
| General Copy ownership | Expanded intentionally, but finding 1 breaks its error path |

The new walker registry introduces exported types/constants/functions used only by tests in this tree. Given private-by-default API policy, move that registry/support into private or test-only scope unless a real downstream consumer is identified. No need to restore arbitrary Fork/Clone APIs.

The sort comparator-by-reference change is deliberately documented and tested. It remains a real behavior change for predicates/key functions that mutate or retain arguments, including sealed literals; do not describe the entire bundle as behavior-preserving.

## Performance evidence

[Exact-head CI benchmark](https://github.com/luthersystems/elps/actions/runs/34365391045/job/102512657471): original attempt, ten interleaved samples per arm, Go 1.25.13, GOMAXPROCS=2. The unchanged **ELPS** gate is 15% timing / 5% allocations, not Substrate's 10% gate.

| CI path | Base | PR | Result |
|---|---:|---:|---|
| Template publish, 250 functions | 2.327 ms | 2.362 ms | No significant change, p=.631 |
| NewVM, 250 functions | 497.2 µs | 489.1 µs | No significant change, p=.052 |
| Reader-cache-copy, 1 KiB | 189.7 µs | 195.8 µs | **+3.22%, p=.002** |
| Reader-cache-copy, 78 KiB | 5.086 ms | 5.158 ms | No significant change, p=.165 |
| TextLoaderLoad | 3.282 ms | 3.334 ms | No significant change, p=.739 |

Template allocation counts are unchanged: publication 414/1673 and NewVM 1111/4865 for the two fixtures. TextLoader admission adds one allocation, about 1.8 KiB / 0.05%. The CI gate reports 108 delta rows plus 688 no-change rows, **37 significant adverse movements**, and zero gate breaches. A pass is not evidence of zero slowdown.

Independent local check: frozen base/head binaries, Go 1.25.13, Apple M3, six alternating AB/BA rounds, GOMAXPROCS=2, 200ms per row. All 24 processes succeeded; all samples retained. Publication/NewVM timing showed no significant change and allocations matched. Reader-cache-copy was **2.80% / 2.54% slower** for the small/library fixtures. Other loader timing changes were not significant. Short/noisy local intervals limit the strength of equivalence claims.

**Bottom line:** no evidence here of the large template-performance regression you were worried about; there are modest load-path slowdowns, and the correctness defects still block approval. Re-run the same gates after fixing them, particularly if deterministic callback traversal changes costs.

No customer application was accessed or benchmarked for this review. These are public synthetic fixtures, not an Acre performance claim. Substrate #483 still pins v1.61.1; its separate passing benchmark does not validate unpublished #642 code.

## Verification and handoff

- Existing targeted Copy/sort/loader/template and LSP size/rendering tests pass.
- Additional Copy, oracle, rename, code-action, and analyzer probes fail as described.
- Copy probes were checked against the base public Copy implementation; rename against the base coordinate-conversion file. Those controls pass.
- Copy/oracle probes also fail in checked/race builds without race reports.
- Review used isolated detached worktrees. No author-branch source changes, commits, pushes, or test deletions.
- Node dependency functional checks: exact tree-sitter CLI 0.27.0 passes all 75 corpus cases with the committed parser, then all 75 after regeneration/rebuild; both query files compile and produce expected captures. VS Code extension build and all five grammar test files pass against the locked dependencies. Full Node-addon packaging is not verified: its existing build route lacks configuration and Node bindings in both base and head. This is distinct from the working CLI, Go/cgo and editor checks.

Requested merge gate: fix findings 1–3 before approval; close the static-check, callback-order and code-action gaps in this bundle with focused regression tests. Keep the runtime's fail-closed boundary and the unchanged performance thresholds.

sam-at-luther:
All seven items are addressed on head 0d3ece9, one commit per finding, each with a regression test that was red before its fix and a negative control. Nothing in the template engine (`NewTemplate`, `NewVM`, admission, storage overlap) is touched.

| # | Finding | Commit | What changed | Proof |
|---|---|---|---|---|
| 1 | Failed `Copy` retains source-owned map storage | 249e616 | The memoised destination header is overwritten in place with the error, so every later encounter of that header yields the same error rather than a sorted-map over the source's `*MapData`. Every payload-copy failure poisons the map memo with a nil sentinel, so a second header over the same payload fails instead of adopting the half-built copy. An adjacent panic is closed too: the generic arm indexed `Entries` cells without checking for `LError`. | `TestCopyFailedMapCopyLeavesNoSourceBackedHeader`: a ranger that fails part-way and an `Entries` path that fails after a partial copy, each as one header reached twice and as two headers over one payload. Red on the old head (write through the copy reached the source; nil-pointer panic on the two-header shape). Removing the in-place overwrite reddens the one-header shapes; removing the memo poison reddens the two-header shapes. |
| 2 | Oversized closed-file rename emits byte columns as UTF-16 | 9bc273a | A conversion that cannot be performed fails the whole rename with an error naming the file and the reason. No partial `WorkspaceEdit`, no unconverted range. Taken one step further than requested: a file the index saw but the disk no longer offers fails the same way, since that was the same corruption with a rarer trigger. utf-8 is unaffected. | `TestRenameOverLimitClosedFileFailsRatherThanEmittingByteColumns` reproduces your numbers exactly (UTF-16 [11,17) vs bytes [12,18), replacing `arget)`), plus the deleted-file and utf-8 arms. Two existing cross-file rename tests that inject refs to files not on disk now negotiate utf-8. |
| 3 | Fuzz exception based on rendered text | 74f0888 | `evalTreeOnce` classifies each result where the evaluator's `*LVal` is still in hand: the tail budget by `errors.As` on the `*TailIterationLimitError` the evaluator stores in the error's data cell (not constructible from Lisp); the deadline by `CondContextCancelled` admitted only when that run's own context actually expired. Only two genuine budgets collapse. `isResourceBackstop(string)` is gone. | `TestBackstopNegativeControls`: two string values containing the phrase, two user-raised errors containing it, and a budget against a value containing it. All three were collapsed by the substring classifier and are divergences now. `TestRunawayLoopStillCollapses` keeps the positive case (tail budget in one arm, deadline in the other). |
| 4 | Static admission differs from runtime admission | cb5fd5c | `runtimeScalarKinds` mirrors `templateInventory.native`'s scalar arm kind for kind, so `uintptr` is reported. `lisp.Value`'s fall-through mirror is exact in both directions, so `[]**LVal` is reported. The kernel-slot rows (`*[]byte`, `*MapData`, `*funData`) exempt only a `.Native` field write or a keyed `LVal{Native: ...}` literal; through a constructor they get their own diagnostic, because a constructor always builds an `LNative` that `native()` refuses. No site in the tree needed an annotation. | `TestNativePayloadAnalyzerMirrorsTemplateAdmission`: each case is one construction spelled twice, once in a fixture the real analyzer runs over and once as a value a real `lisp.NewTemplate` is asked to publish in-process, and both verdicts must agree. Mutation-verified: putting `uintptr` back fails this test, not only the fixture. |
| 5 | Map traversal exposes Go map order to host callbacks | 0d3ece9 | The sortedmap arm walks the keys sorted; the ranger arm sorts its collected pairs before copying any value. The `Entries` arm is sorted by contract. | `TestCopyMapValueCloneOrderIsDeterministic`: a counter-driven `NativeCloner` over a 64-key map, copied three times, key-to-clone assignment must agree. Both arms failed before. Cost: none of the gate rows moves and their allocation counts are identical to the byte (StableSortMaps, InsertSortedMaps, AliasConcatCopy, ForkDecodedMaps, AssocDecodedMap, interleaved before/after). The cost lands on the map-copy path itself, which no gate row exercises: about +52% on a 64-entry sorted-map copy, +73% at 512, one extra allocation. That is documented at the site. |
| 6 | Code actions bypass the oversized-document guard | b983eb4 | `doc.OverLimit()` guard in `textDocumentCodeAction`, returning nothing. | A `codeAction` row in `TestDocumentSizeGuard_OverLimitHandlersAreCheap`, driven by a genuine published diagnostic captured while the document was small. Before the guard it allocated 10.5 MB on an 8 MiB document and returned a quick fix. |
| 7 | Walker registry API visibility | 02349a8 | The registry types and accessors in `walkers.go` had no caller outside the package's own drift tests. All unexported; two accessors nothing called are dropped. | Drift tests unchanged in what they check. |

On the performance evidence: the reader-cache-copy 1 KiB +3.22% you measured on the previous head predates these commits. Nothing here touches the load path; finding 5 adds cost only to sorted-map copies, which that row does not perform. The CI benchmark gate on this head is the number to read.

---

sam-at-luther:
Second round, on head 5687def. The four remaining gaps are closed, one commit each, plus benchmark coverage for the Copy-of-map path. The load-path slowdown was bisected rather than dismissed; the verdict is at the end.

| Gap | Commit | What changed | Proof |
|---|---|---|---|
| Cyclic Copy failure returns an unfinished map | c79cec5 | The copier is fail-stop. The first failing arm records the error, every level returns it without descending, the cell loop stops as soon as a child sets it, and `Copy` returns the error rather than a container with the error inside it. Nothing half-built escapes. The in-place overwrite and the memo poison from round one stay, so the memo is consistent while the walk unwinds. | `TestCopyFailedMapCopyLeavesNoSourceBackedHeader` (extended): both failing fixtures (ranger part-way, `Entries` after a partial copy) across one header reached twice, two headers over one payload, and the map reaching itself through the other header in both orders. Red on the previous head: the `Entries` shape returned a list whose second cell was a sorted-map over the half-built payload and writing through it hit the nil-pointer panic you described. Green after; reverting the fail-stop reddens all ten shapes. |
| Captured budget-error replay fools the fuzz comparison | 27164a7 | No value-based classification remains. The tail-iteration budget is deterministic across the private and shared arms (same program, same fresh env, same limit), so exact comparison already agrees and it needs no exception. The only nondeterministic budget is the wall-clock deadline, and it is handled as harness state: if either arm's own context has expired at expression j, the input is inconclusive from j on and expressions before j still compare exactly. A program cannot expire the harness's context. | Red before: `(handler-bind ([condition (lambda (c &rest r) (set! captured (car r)) 'caught)]) (s))` then `(error 'replay "<divergent payload>" captured)` collapsed two results that differ. `TestOracleReportsCapturedBudgetReplay` now reports it. Also pinned: two tail-budget arms compare equal with no exception, an expired context is inconclusive only from that expression on, and a program raising the reserved `context-cancelled` condition under a live context is compared exactly. 60s fuzz smoke, no new crasher. |
| Native field assignments bypass the static check | 5687def | A row exempts a site only when it is inside package `lisp` and, for a keyed literal, the same literal's `Type:` key names the row's header (`LBytes`, `LSortMap`, `LFun`), resolved through the type checker. `LVal{Type: LNative, Native: &b}` is reported. Every `.Native` write or row literal outside package `lisp` is reported, which is where the bypass lived. One correction to the brief I gave: seven guarded `.Native` field writes of row types do exist inside the kernel (copier, detach, planner), each preceded by a check of the header's own `Type`; those stay exempt by the in-kernel condition and are named in the analyzer header. | Paired runtime test extended: `v := lisp.Native(int64(0)); v.Native = &b` and `&lisp.LVal{Type: lisp.LNative, Native: &b}`, each reported by the analyzer and refused by `lisp.NewTemplate` with the same `native *[]uint8 has no template immutability declaration`. Mutation: dropping the `Type:` condition turns the paired test red. Both elpsvet modes still exit 0 with no new annotation. |
| Custom-map `Entries` order is nondeterministic | cfdff7d | The `Map` interface promises sorted `Keys` and says nothing about `Entries`, so round one's "sorted by contract" was wrong for a host map. The `Entries` arm now sorts pairs by (key text, key type) before copying any value. And the optimisation you prototyped: when every value in the map is a leaf (no cells, no payload) no host hook can run, so all three arms skip the sort. | `TestCopyMapValueCloneOrderIsDeterministic/custom_entries_order`: a host `Map` whose `Entries` returns a different permutation on alternate calls, counter-driven clones, 64 keys, three copies must agree. Red before, green after. Negative controls: deleting the `Entries` sort reddens only that arm; weakening the leaf predicate to "no payload" reddens the nested-native test on all three arms. |
| Benchmark coverage | 7389cf2 | `BenchmarkCopySortedMap` over 64 and 512 keys in three shapes: `scalar` (leaf fast path), `nested` (stock arm with the sort live), `custom` (host map on the `Entries` arm with the sort live). The gate now has rows on this path. | Interleaved across the leaf fast path, n=6: scalar drops one allocation and about 10% of bytes at both sizes (p=0.002), timing directionally faster and significant at 512 (-39%). I did not reproduce your 33 to 55% as a significant result at n=6 on this box. Nested is within noise; the sort still runs there. |

**Load path, bisected.** The 3 to 6% is real and it is carried by the `(*LVal).Copy` memoisation commit from #604. The gate-fix commit removed all of the allocation half and about two thirds of the timing half. What survives to head is about +5 to 6% on `Copy` itself over a parse tree, which dilutes to about +2 to 3% on the load rows. Allocations are byte-identical to main. Method: seven build points from main to head, each row in its own process under SCHED_FIFO because the shared box's spread without it was ±37% to ±641%, interleaved round-robin, n=10 to n=25, benchstat. The mechanism is not hashing: on both benchmark sources every top-level form memoises exactly 16 headers, which is the inline memo's size, so no map is ever allocated and pprof shows no pointer-keyed map access. The cost is a failed 16-entry linear scan per container that never hits (a parse tree shares nothing), one store per container, and the memoise test on every node. I did not take the fix I had in mind (a loader hint to skip the memo): `loaderWalk`'s admission policy admits a repeated composite, so an interning reader's DAG is admitted and the memo is what preserves that sharing across the copy. Skipping it would silently unfold the DAG. The only semantics-preserving levers are inside the copier and are worth well under 1% by the profile. My recommendation is to accept the residual as the price of a `Copy` that memoises, which is what fixed the five aliasing defects this PR measured live on main. If you want it paid down anyway, say so and I will do it as a separate PR with its own measurement.

CI's gate on this head agrees with the bisection: 0 rows at or above threshold; the 1 KiB reader-cache-copy row is +4.10% (p=0.000) and the 78 KiB row and TextLoaderLoad are within noise (p=0.12, p=0.11); bytes and allocations identical on all three. The six new CopySortedMap rows now appear in the report and will carry a delta from the next PR on.

One side finding from the bisection worth a follow-up: both load benchmarks generate uniform forms with exactly 16 memoisable headers each, so they measure one design point of the memo and sit one node below the spill to a map. A benchmark over a real library source would be a better gate row. Not in this PR.

---

sam-at-luther:
Third round, on head 09814f4. The three remaining gaps are closed, one commit each. The Acre benchmark you asked for is at the end: publication, checkout and endpoint execution are performance-neutral on the real phylum.

| Gap | Commit | What changed | Proof |
|---|---|---|---|
| Custom maps with scalar values copy differently by entry order | 09814f4 | The generic `Entries` arm now sorts by (key text, key type) on every copy; the leaf fast path stays only on the two arms whose keys are unique Go strings by construction. After the sort, two adjacent string-like keys with the same text (`"a"` and `'a`, or a duplicate) fail the copy naming the key, since the stock map keys on text and would otherwise keep whichever came last. | `TestCopyHostMapEntriesOrderIsIrrelevant`: a host `Map` backed by a pair slice that rotates its `Entries` order on alternate calls. Red before: the string-and-symbol case and the duplicate case each copied successfully with DIFFERENT contents (`(sorted-map 'a 2 "z" 3)` then `(sorted-map 'a 1 "z" 3)`). Green after: both copies fail with the same collision error; the unique-key case copies identically either way. Negative control: reinstating the fast path reddens exactly the two all-scalar collision cases. |
| Static checker accepts a locally shadowed header constant | 73e86dd | `headerTypeNamed` now requires the `Type:` key's constant to BE the package-scope object: `konst.Pkg().Scope().Lookup(konst.Name()) == konst`. A function-local `const LBytes LType = LNative` inside package `lisp` fails that identity and the literal is reported with a reason naming the failed condition. Identity rather than value, because `const LBytes LType = LBytes` is a legal shadow with the same value that must still be reported. | In-kernel fixtures `shadowedTypeKey` and `shadowedTypeKeySameValue` (both reported) and paired case `PairShadowedHeaderConst`: the analyzer reports it and `lisp.NewTemplate` refuses what it constructs with `native *[]uint8 has no template immutability declaration`. Mutation: deleting the identity check turns four expectations red across both roots. Both elpsvet modes still exit 0. |
| Real fuzz runner can become always-inconclusive without tests failing | 37dbfaa | `sharedTreeProperty` returns a verdict per shared arm (inconclusive reason, expressions compared) instead of only logging. `TestSharedTreeLiveResultIsConclusive` drives the REAL path (real env, real deadline, real `evalTreeOnce`) on a five-form program with a value, a defun, a call, a caught error and an uncaught one, and requires every arm conclusive with every expression compared. `TestSharedTreeSeedsAgree` requires the same per terminating seed and, in `t.Cleanup`, that at least one seed was conclusive. | Three mutations, each applied and reverted: `expired` pinned true, `compareTreeEvals` unconditionally inconclusive, and the arm context cancelled at creation. Each reddens the new test; the first and third also redden all 14 seed subtests. 30s fuzz smoke, no crasher. |

CI on 09814f4: 25/25, benchmark gate 0 rows at or above threshold. The load rows read +4.61% (1 KiB reader-cache-copy), +2.94% (78 KiB) and +4.21% (TextLoaderLoad) in timing with bytes and allocations identical to main on all three, which is the same residual the bisection attributed to the Copy memo's bookkeeping in the previous round; nothing in this round touches that path.

**Acre phylum, elps main (03e7c95, v1.61.1) vs PR head (5687def), substrate held constant at b17c3cb.** The customer phylum (4.7 MB, ~50 production `.lisp` files) with one local-only change: `client_utils.lisp:4`'s load-scope `(set 'now-date (timestamp-to-date (cc:timestamp (cc:now))))` replaced by a placeholder, since `main.lisp` already sets `now-date` in a tx pre-hook and the load-scope read only stalled the preheater at that line. `shirotester preheat-check` went from that single flagged read to `fully preheatable` on both the directory and the deployed zip; the phylum's own test suite gives the same single pre-existing failure before and after. Ten interleaved rounds per benchmark, `GOMAXPROCS=2`, SCHED_FIFO, three processes per arm per round:

| Row | main | head | Verdict |
|---|---|---|---|
| Publication (`prodPhylum/cold`) | 65.50 ms | 66.40 ms | ~ p=0.218; B/op and allocs identical |
| Checkout, NewVM per request (`prodPhylum/fork`) | 12.08 ms | 12.62 ms | ~ p=0.052; 3.344 MiB and 29.5k allocs/op identical (p=0.97, 0.54) |
| Retained heap per VM (`retained-fork`) | 2.979 MiB | 2.979 MiB | all ten samples equal |
| Concurrent checkout, fork path | 3.339 ms | 3.422 ms | ~ p=0.105 |
| Concurrent cold start, cold path | 473.6 ms | 474.4 ms | ~ p=0.631 |
| Concurrent checkout, cold fallback path | 49.32 ms | 49.87 ms | ~ p=0.796; B/op +0.34%, allocs +0.28% (p=0.015, 0.016) |
| `shirotester unit-tests`, 29 files, wall clock, 5 rounds | 101.16 s | 100.15 s | -1.0%, p=0.17, not separable from run order |

Every timed row is within noise in both directions; the only significant rows are a third of a percent of bytes and allocations on the cold fallback path. The six fixture-phylum preheat rows (hard-wired to `testfixtures/shiro.lisp`) are within noise too. What this box resolves is about a 5% wall-clock change on the prod rows and about 1% on allocations. No benchmark in the preheat package invokes an endpoint through the router, so endpoint cost is covered only by the unit-test wall clock. Nothing from this run is checked in; the phylum copy stayed in scratch.

---

sam-at-luther:
Third round, on head 09814f4. The three remaining gaps are closed, one commit each. The Acre benchmark you asked for is at the end: no clear timing regression was detected at the precision of the customer benchmark; endpoint timing was assessed only indirectly through the unit suite.

| Gap | Commit | What changed | Proof |
|---|---|---|---|
| Custom maps with scalar values copy differently by entry order | 09814f4 | The generic `Entries` arm now sorts by (key text, key type) on every copy; the leaf fast path stays only on the two arms whose keys are unique Go strings by construction. After the sort, two adjacent string-like keys with the same text (`"a"` and `'a`, or a duplicate) fail the copy naming the key, since the stock map keys on text and would otherwise keep whichever came last. | `TestCopyHostMapEntriesOrderIsIrrelevant`: a host `Map` backed by a pair slice that rotates its `Entries` order on alternate calls. Red before: the string-and-symbol case and the duplicate case each copied successfully with DIFFERENT contents (`(sorted-map 'a 2 "z" 3)` then `(sorted-map 'a 1 "z" 3)`). Green after: both copies fail with the same collision error; the unique-key case copies identically either way. Negative control: reinstating the fast path reddens exactly the two all-scalar collision cases. |
| Static checker accepts a locally shadowed header constant | 73e86dd | `headerTypeNamed` now requires the `Type:` key's constant to BE the package-scope object: `konst.Pkg().Scope().Lookup(konst.Name()) == konst`. A function-local `const LBytes LType = LNative` inside package `lisp` fails that identity and the literal is reported with a reason naming the failed condition. Identity rather than value, because `const LBytes LType = LBytes` is a legal shadow with the same value that must still be reported. | In-kernel fixtures `shadowedTypeKey` and `shadowedTypeKeySameValue` (both reported) and paired case `PairShadowedHeaderConst`: the analyzer reports it and `lisp.NewTemplate` refuses what it constructs with `native *[]uint8 has no template immutability declaration`. Mutation: deleting the identity check turns four expectations red across both roots. Both elpsvet modes still exit 0. |
| Real fuzz runner can become always-inconclusive without tests failing | 37dbfaa | `sharedTreeProperty` returns a verdict per shared arm (inconclusive reason, expressions compared) instead of only logging. `TestSharedTreeLiveResultIsConclusive` drives the REAL path (real env, real deadline, real `evalTreeOnce`) on a five-form program with a value, a defun, a call, a caught error and an uncaught one, and requires every arm conclusive with every expression compared. `TestSharedTreeSeedsAgree` requires the same per terminating seed and, in `t.Cleanup`, that at least one seed was conclusive. | Three mutations, each applied and reverted: `expired` pinned true, `compareTreeEvals` unconditionally inconclusive, and the arm context cancelled at creation. Each reddens the new test; the first and third also redden all 14 seed subtests. 30s fuzz smoke, no crasher. |

CI on 09814f4: 25/25, benchmark gate 0 rows at or above threshold. The load rows read +4.61% (1 KiB reader-cache-copy), +2.94% (78 KiB) and +4.21% (TextLoaderLoad) in timing with bytes and allocations identical to main on all three, which is the same residual the bisection attributed to the Copy memo's bookkeeping in the previous round; nothing in this round touches that path.

**Acre phylum, elps main (03e7c95, v1.61.1) vs measured candidate (5687def; before final head 09814f4), substrate held constant at b17c3cb.** A complex customer phylum was tested with an authorized scratch-only adjustment to a load-time clock read. Both source-directory and deployed-archive preheat checks then reported fully preheatable. The customer unit suite retained the same pre-existing failure before and after the adjustment. No customer source is included here. Ten interleaved rounds per benchmark, `GOMAXPROCS=2`, SCHED_FIFO, three processes per arm per round:

| Row | main | head | Verdict |
|---|---|---|---|
| Publication (`prodPhylum/cold`) | 65.50 ms | 66.40 ms | ~ p=0.218; B/op and allocs identical |
| Checkout, NewVM per request (`prodPhylum/fork`) | 12.08 ms | 12.62 ms | ~ p=0.052; 3.344 MiB and 29.5k allocs/op identical (p=0.97, 0.54) |
| Retained heap per VM (`retained-fork`) | 2.979 MiB | 2.979 MiB | all ten samples equal |
| Concurrent checkout, fork path | 3.339 ms | 3.422 ms | ~ p=0.105 |
| Concurrent cold start, cold path | 473.6 ms | 474.4 ms | ~ p=0.631 |
| Concurrent checkout, cold fallback path | 49.32 ms | 49.87 ms | ~ p=0.796; B/op +0.34%, allocs +0.28% (p=0.015, 0.016) |
| `shirotester unit-tests`, 29 files, wall clock, 5 rounds | 101.16 s | 100.15 s | -1.0%, p=0.17, not separable from run order |

None of the reported timed differences reached statistical significance; this is not an equivalence test or proof of zero overhead. the only significant rows are a third of a percent of bytes and allocations on the cold fallback path. The six fixture-phylum preheat rows (hard-wired to `testfixtures/shiro.lisp`) are within noise too. What this box resolves is about a 5% wall-clock change on the prod rows and about 1% on allocations. No benchmark in the preheat package invokes an endpoint through the router, so endpoint cost is covered only by the unit-test wall clock. Nothing from this run is checked in; the phylum copy stayed in scratch.


Privacy correction: this sanitized comment replaces the earlier benchmark reply. Its customer source excerpt has been removed rather than preserved in comment edit history. The benchmark candidate was 5687def; subsequent production changes are confined to generic-map copying and the static analyzer, not template publication or NewVM.

sam-at-luther:
Third round, on head 09814f4 (this replaces an earlier version of this comment that quoted customer source; that version is deleted). The three remaining gaps are closed, one commit each. The customer-phylum benchmark you asked for is at the end: publication, checkout and endpoint execution are performance-neutral on the real phylum.

| Gap | Commit | What changed | Proof |
|---|---|---|---|
| Custom maps with scalar values copy differently by entry order | 09814f4 | The generic `Entries` arm now sorts by (key text, key type) on every copy; the leaf fast path stays only on the two arms whose keys are unique Go strings by construction. After the sort, two adjacent string-like keys with the same text (`"a"` and `'a`, or a duplicate) fail the copy naming the key, since the stock map keys on text and would otherwise keep whichever came last. | `TestCopyHostMapEntriesOrderIsIrrelevant`: a host `Map` backed by a pair slice that rotates its `Entries` order on alternate calls. Red before: the string-and-symbol case and the duplicate case each copied successfully with DIFFERENT contents (`(sorted-map 'a 2 "z" 3)` then `(sorted-map 'a 1 "z" 3)`). Green after: both copies fail with the same collision error; the unique-key case copies identically either way. Negative control: reinstating the fast path reddens exactly the two all-scalar collision cases. |
| Static checker accepts a locally shadowed header constant | 73e86dd | `headerTypeNamed` now requires the `Type:` key's constant to BE the package-scope object: `konst.Pkg().Scope().Lookup(konst.Name()) == konst`. A function-local `const LBytes LType = LNative` inside package `lisp` fails that identity and the literal is reported with a reason naming the failed condition. Identity rather than value, because `const LBytes LType = LBytes` is a legal shadow with the same value that must still be reported. | In-kernel fixtures `shadowedTypeKey` and `shadowedTypeKeySameValue` (both reported) and paired case `PairShadowedHeaderConst`: the analyzer reports it and `lisp.NewTemplate` refuses what it constructs with `native *[]uint8 has no template immutability declaration`. Mutation: deleting the identity check turns four expectations red across both roots. Both elpsvet modes still exit 0. |
| Real fuzz runner can become always-inconclusive without tests failing | 37dbfaa | `sharedTreeProperty` returns a verdict per shared arm (inconclusive reason, expressions compared) instead of only logging. `TestSharedTreeLiveResultIsConclusive` drives the REAL path (real env, real deadline, real `evalTreeOnce`) on a five-form program with a value, a defun, a call, a caught error and an uncaught one, and requires every arm conclusive with every expression compared. `TestSharedTreeSeedsAgree` requires the same per terminating seed and, in `t.Cleanup`, that at least one seed was conclusive. | Three mutations, each applied and reverted: `expired` pinned true, `compareTreeEvals` unconditionally inconclusive, and the arm context cancelled at creation. Each reddens the new test; the first and third also redden all 14 seed subtests. 30s fuzz smoke, no crasher. |

CI on 09814f4: 25/25, benchmark gate 0 rows at or above threshold. The load rows read +4.61% (1 KiB reader-cache-copy), +2.94% (78 KiB) and +4.21% (TextLoaderLoad) in timing with bytes and allocations identical to main on all three, which is the same residual the bisection attributed to the Copy memo's bookkeeping in the previous round; nothing in this round touches that path.

**Customer phylum, elps main (03e7c95, v1.61.1) vs PR head (5687def), substrate held constant at b17c3cb.** A real production phylum (4.7 MB, about 50 production `.lisp` files) with one local-only change: a single load-scope clock read, the one thing `shirotester preheat-check` flagged, was moved into per-transaction scope (the phylum already sets the same value in a transaction pre-hook, so the load-scope read only stalled the preheater at that line). After the change preheat-check reports the phylum fully preheatable on both the source directory and the deployed zip, and the phylum's own test suite gives the same single pre-existing failure before and after. Ten interleaved rounds per benchmark, `GOMAXPROCS=2`, SCHED_FIFO, three processes per arm per round:

| Row | main | head | Verdict |
|---|---|---|---|
| Publication (`prodPhylum/cold`) | 65.50 ms | 66.40 ms | ~ p=0.218; B/op and allocs identical |
| Checkout, NewVM per request (`prodPhylum/fork`) | 12.08 ms | 12.62 ms | ~ p=0.052; 3.344 MiB and 29.5k allocs/op identical (p=0.97, 0.54) |
| Retained heap per VM (`retained-fork`) | 2.979 MiB | 2.979 MiB | all ten samples equal |
| Concurrent checkout, fork path | 3.339 ms | 3.422 ms | ~ p=0.105 |
| Concurrent cold start, cold path | 473.6 ms | 474.4 ms | ~ p=0.631 |
| Concurrent checkout, cold fallback path | 49.32 ms | 49.87 ms | ~ p=0.796; B/op +0.34%, allocs +0.28% (p=0.015, 0.016) |
| `shirotester unit-tests`, 29 files, wall clock, 5 rounds | 101.16 s | 100.15 s | -1.0%, p=0.17, not separable from run order |

Every timed row is within noise in both directions; the only significant rows are a third of a percent of bytes and allocations on the cold fallback path. The six fixture-phylum preheat rows (hard-wired to `testfixtures/shiro.lisp`) are within noise too. What this box resolves is about a 5% wall-clock change on the prod rows and about 1% on allocations. No benchmark in the preheat package invokes an endpoint through the router, so endpoint cost is covered only by the unit-test wall clock. Nothing from this run is checked in; the phylum copy stayed in scratch.

---

sam-at-luther:
Fourth-pass leftovers implemented directly on this PR in **0a0476b**.

- Unsupported custom-map keys now fail before value-clone callbacks (#643). The regression test was red before the change; direct and nested cases now pass, and the supported-map control verifies real cloning.
- Targeted tests pass normally and under `elpscheck` + race; focused elpsvet and independent Go/security/QA review pass.
- The affected benchmark reply was replaced with a [sanitized version](https://github.com/luthersystems/elps/pull/642#issuecomment-5608022984). Benchmark revision, endpoint scope and statistical claims are corrected.
- No customer source in the code change; no stock/ranger or template/NewVM path changes.

Exact-head CI: **25/25 green on 0a0476b**, including the [benchmark gate](https://github.com/luthersystems/elps/actions/runs/34399571372), attempt 1. Zero rows breach the ELPS gate (15% timing / 5% allocation). Template-plan publication and checkout timings show no significant change. The small/larger cached-reader Copy rows are +1.98% / +3.25%, and TextLoaderLoad is +2.34% versus main; these are same-run comparisons, not a claimed improvement over earlier runners.

The fourth-pass leftovers are complete. This PR is ready from my review; no merge or branch cleanup was performed.

PR #647, opened 2026-09-09
lint: report side effects in sort comparators and writes to an iterated collection

### LINKED ISSUE #644: lint: comparator-mutation, report side effects inside a stable-sort or insert-sorted predicate
## Problem

A `stable-sort` or `insert-sorted` predicate runs an unspecified number of times in an unspecified order. Any side effect inside it is therefore nondeterministic. After #642 the predicate receives the list's own elements by reference, so a write through an argument now lands on the real element instead of on a throwaway copy, and a write through a sealed program literal raises `modify-literal-error`. Nothing warns the author at edit time.

## Proposed check

New analyzer `comparator-mutation`, severity **error**, in `lint/analyzers.go`, registered in `DefaultAnalyzers()`.

Report any call to a mutating builtin anywhere inside the body of the predicate passed to `stable-sort` or `insert-sorted`, whatever the target: `assoc!`, `dissoc!`, `append!`, `append-bytes!`, `set!`, and a nested `stable-sort` (which sorts in place). Cover:

- an inline `lambda` predicate;
- a predicate named by symbol that resolves to a `defun` in the same file (one hop, no recursion);
- nested lambdas inside the predicate body.

Diagnostic: `stable-sort predicate mutates state: <builtin> is called inside a comparator` with a note explaining that comparators run in unspecified order and count, and that after v1.62 arguments are the list's own elements. Suppression via `; nolint:comparator-mutation` works through the existing `filterSuppressed`.

## Non-goals

- Mutation reached through a helper function defined elsewhere or in another file.
- A predicate passed as an arbitrary expression.

## Acceptance

- Positive and negative fixtures in `lint/lint_test.go`, including a pure comparator, a comparator using `let` without mutation, the accumulator idiom outside a comparator (must not report), and the same-file `defun` hop.
- `TestDefaultAnalyzers` count bumped; `docs/lang.md` gains a paragraph under the sort section; `elps doc -m` clean.
- `make test`, `make static-checks`, `scripts/ci-gates-test.sh` green.

### LINKED ISSUE #645: lint: iteration-mutation, report writes to the collection being iterated or to the current element
## Problem

Mutating a collection while iterating over it, or mutating the element handed to the callback, produces order-dependent results and after #642 the element aliases the collection's own storage in every higher-order builtin. Nothing warns the author at edit time.

## Proposed check

New analyzer `iteration-mutation`, severity **warning**, in `lint/analyzers.go`, registered in `DefaultAnalyzers()`.

For each of `map`, `foldl`, `foldr`, `select`, `reject`, `zip`, `all?`, `any?`: when the collection argument is a plain symbol, report a mutating call inside the callback body (`assoc!`, `dissoc!`, `append!`, `append-bytes!`, `set!`, `stable-sort`) whose target is

- that same collection symbol, or
- one of the callback's own parameters (the element).

Resolution is syntactic. The accumulator idiom stays clean: in `(foldl (lambda (acc x) (assoc! acc k v)) (sorted-map) xs)` the target `acc` is neither the collection nor an element, so nothing is reported. Cover inline lambdas and a callback named by symbol that resolves to a same-file `defun` (one hop).

Diagnostic: `<builtin> mutates <symbol> while <form> iterates over it` (or `... mutates the element <param> of <form>`), with a note recommending a copy or building a new collection. Suppression via `; nolint:iteration-mutation`.

## Non-goals

- A collection passed as an expression rather than a symbol.
- Mutation reached through a helper defined elsewhere.

## Acceptance

- Positive and negative fixtures in `lint/lint_test.go`: collection symbol mutated, element mutated, accumulator idiom clean, unrelated map mutated clean, same-file `defun` hop, `nolint` suppression.
- `TestDefaultAnalyzers` count bumped; `docs/lang.md` paragraph; `elps doc -m` clean.
- `make test`, `make static-checks`, `scripts/ci-gates-test.sh` green.

### LINKED ISSUE #649: lint: bound mutation analysis for nested callbacks and distinct collection names
Follow-up review of #647 found quadratic work despite deduplicated output: an 800-deep, 60 KB callback tree allocates about 96 MB; a 44 KB helper reused across 600 distinct collection names allocates about 46 MB. Avoid repeated flattened descendant lists and repeated matching/formatting; add regression bounds that fail when the work optimization is removed. Fix belongs in #647.

### LINKED ISSUE #650: lint: match ELPS nested quasiquote and qualified quote evaluation
Follow-up review of #647 found mutation lint disagrees with actual evaluation: lisp:quote and lisp:quasiquote data produce false errors; an unquote inside nested quasiquote executes in ELPS but is missed. Correct the walker without changing interpreter semantics, and pair regression tests with real runtime behavior. Fix belongs in #647.

### LINKED ISSUE #651: lint: mutation checks miss lisp:-qualified spellings and scan the wrong duplicate defun
Found by an adversarial re-review of #647 at head 97e5ac6, after the quote-traversal fix (#650) had landed.

## 1. Qualified spellings evade both checks (medium)

Commit a741148 taught `walkEvaluated` to treat `lisp:quote` and `lisp:quasiquote` as data, but every hazard-side head comparison still matches only the bare name. Each of these runs and mutates, and each draws zero diagnostics on its own:

```lisp
(lisp:stable-sort (lambda (a b) (assoc! a "k" 1) (< a b)) xs)
(stable-sort (lisp:lambda (a b) (assoc! a "k" 1) (< a b)) xs)
(stable-sort (lambda (a b) (lisp:assoc! a "k" 1) (< a b)) xs)
(lisp:map 'list (lambda (x) (assoc! x "k" 1) x) xs)
```

`lisp` is the only package exporting these operators, and the interpreter resolves both spellings to the same function. `lisp:assoc!` and `lisp:lambda` both already occur in this repository's own `.lisp` sources, so the spelling is not synthetic.

Note that `lisp:unquote` inside a template must stay data: `getUnquoteType` in `lisp/macro.go` compares the bare string literally, and `(quasiquote (a (lisp:unquote v) b))` evaluates to `'(a (lisp:unquote v) b)`.

## 2. Duplicate `defun` resolves to the first definition (low)

`sameFileDefuns` keeps the first definition of a duplicated name, but `defun` overwrites, so the interpreter runs the last one. A clean-first, dirty-second duplicate is a silent miss, and a dirty-first, clean-second duplicate reports against a body no call reaches.

```lisp
(defun dup (a b) (< a b))
(defun dup (a b) (assoc! a "second" 1) (< a b))
(stable-sort dup xs)   ; runs the second body and mutates; no diagnostic
```

## Not in scope, documented as limits

A same-file rebinding such as `(defun quote (x) x)` shadows the special form, the write runs, and both checks stay silent. That is a property of every syntactic analyzer in `lint/`, not of these two, and is now named in `docs/lint-checks.md`.

sam-at-luther:
## Summary

Two conservative structural lint checks, with callback-work bounds and quote handling matched to the ELPS interpreter.

| Check | Severity | Reports |
|---|---|---|
| `comparator-mutation` | error | Known mutating calls inside `stable-sort` or `insert-sorted` predicates. This intentionally includes callback-local scratch writes; it is not proof that every reported write changes the result. |
| `iteration-mutation` | warning | Writes through a collection symbol or callback element in `map`, `foldl`, `foldr`, `select`, `reject`, `all?` or `any?`. Fold accumulators and `set!` rebinding are exempt. |

Both follow inline lambdas and same-file defuns one hop, respect `nolint`, and skip quoted data. ELPS unquotes inside nested quasiquotes still execute and must be checked. Canonical `lisp:quote` and `lisp:quasiquote` spellings are recognized, and so is the `lisp:` spelling of every operator on the hazard side.

This supports the comparator-by-reference change in #642; it does not alter runtime evaluation, copying or template performance. MCP help-table drift is handled separately by #648.

## Limits

- These are syntactic checks, not an effect system: cross-file helpers, indirect calls and aliases are not resolved.
- The iteration check does not track inner rebinding/shadowing; collection expressions rather than plain symbols are not followed.
- A same-file rebinding of `quote`, `quasiquote` or a builtin name (for example `(defun quote (x) x)`) is not tracked; the checks then read the shadowed meaning. Every syntactic analyzer in `lint/` shares this property.
- `zip` has no callback and is intentionally outside the check.

## Final review follow-up

- #649: bound work for deeply nested callbacks and helpers reused across distinct collection names, not just diagnostic count.
- #650: pair quote diagnostics with actual interpreter execution; correct the previous nested-quasiquote expectation without changing the interpreter.
- Clarify the comparator rule's conservative policy in help text and docs.
- #651: recognise `lisp:`-qualified spellings on the hazard side (the quote fix had only covered the data side), and resolve a duplicated same-file `defun` to the last definition, which is the one the interpreter runs. `lisp:unquote` inside a template deliberately stays data, matching `getUnquoteType`.

## Verification

- Focused lint package tests and mutation tests under `-tags elpscheck -race` pass on Go 1.25.13.
- CI-matching golangci-lint 2.6.2 reports zero issues in the changed package.
- 28 paired evaluator/linter quote cases; eight failed before the fix. Two further qualified-spelling contexts added with #651's fix, run through the real evaluator.
- Independent before/after comparison preserves diagnostic messages and positions for 1,080 synthetic callback cases.
- Allocation bounds fail on the old implementation. Disabling range merging also fails the new 1,000-case exact-once model and allocation guard, even though output counts remain correct.
- 800-deep comparator allocations: approximately 2.68 million before, 124 thousand after. Distinct-collection fixture: 3.26 million before, 71 thousand after. These include parsing and diagnostics, not just the walker.
- New `BenchmarkMutationChecksScaling` rows cover nested comparators, nested iteration and distinct collections at 200/800 sizes. A short local warm MCP lint comparison found no significant timing difference; overhead was eight allocations (+0.13%) and about 330 bytes (+0.05%). No VM/request execution code changed. The #651 commits leave the scaling allocations unchanged to within one allocation.
- Independent QA/security review: no remaining blocker on 97e5ac6. Independent adversarial re-check of the final-round commits against the interpreter: work bounds exactly linear, dedupe swallows nothing, fold accumulator positions correct, docs claims hold; its two findings are the #651 fixes.
- Final head `265113f` passes all 25 CI checks, including full tests, fuzz shards and the benchmark gate.
- Combined #647 + #648 overlay passes the MCP Help/Lint tests under race, including the generated analyzer registry table with both new rows.

Closes #644
Closes #645
Closes #649
Closes #650
Closes #651

### DIFF
--- a/docs/lang.md
+++ b/docs/lang.md
@@ -661,6 +661,21 @@ $ elps lint sort.lisp
 error: stable-sort predicate mutates state: assoc! is called inside a comparator (comparator-mutation)
 ```
 
+The higher-order functions carry a milder version of the same hazard: a `map`,
+`foldl`, `foldr`, `select`, `reject`, `all?` or `any?` callback that writes
+through the sequence it is walking — or through an element that sequence handed
+it — is traversing a value that changes underneath it.  The
+`iteration-mutation` check reports that shape, leaving a fold's own accumulator
+alone, since threading it is the point of the fold.  `set!` is not a write of
+this kind and is not reported: it rebinds a name rather than the value the name
+held, so the traversal goes on walking exactly the sequence it was handed.
+
+```lisp
+(map 'list (lambda (x) (append! xs x)) xs)
+; lint: append! mutates xs while map iterates over it
+(foldl (lambda (acc x) (assoc! acc x 1)) (sorted-map) xs)  ; fine
+```
+
 ### Sorted Maps
 
 A sorted map is a mapping between keys and values which ensures that key
--- a/docs/lang.md
+++ b/docs/lang.md
@@ -635,6 +635,32 @@ empty literal-derived input — there is nothing in it to modify — and hand
 back fresh storage.  `(stable-sort < (rest xs))` therefore behaves the same
 however short `xs` is.
 
+#### The predicate must be pure
+
+`stable-sort` and `insert-sorted` call their predicate an unspecified number
+of times, in an unspecified order, and hand it the list's **own** elements —
+not copies.  A side effect inside a comparator therefore has no defined
+schedule, and writing through an element it was handed writes through to the
+list being sorted, while it is being sorted.
+
+```lisp
+;; BAD -- runs an unknown number of times, in an unknown order
+(stable-sort
+  (lambda (a b) (assoc! a 'visited true) (< (get a 'n) (get b 'n)))
+  records)
+
+;; GOOD -- compares, and does nothing else
+(stable-sort (lambda (a b) (< (get a 'n) (get b 'n))) records)
+```
+
+The `comparator-mutation` lint check reports a mutating call anywhere inside a
+comparator, whatever that call writes to:
+
+```
+$ elps lint sort.lisp
+error: stable-sort predicate mutates state: assoc! is called inside a comparator (comparator-mutation)
+```
+
 ### Sorted Maps
 
 A sorted map is a mapping between keys and values which ensures that key
--- a/docs/lint-checks.md
+++ b/docs/lint-checks.md
@@ -274,6 +274,32 @@ branch matches. While sometimes intentional, this is often an oversight.
   (error 'test "data"))
 ```
 
+### `comparator-mutation`
+
+**Flags a mutating call inside a `stable-sort` or `insert-sorted` predicate.**
+(Severity: error)
+
+A comparator runs an unspecified number of times, in an unspecified order, and
+it is handed the list's own elements — so any side effect inside one is
+nondeterministic, whatever it writes to. Every mutating builtin is reported:
+`assoc!`, `dissoc!`, `append!`, `append-bytes!`, `set!`, and `stable-sort`
+itself, which sorts in place despite carrying no `!`.
+
+```lisp
+;; BAD — the predicate mutates
+(stable-sort (lambda (a b) (assoc! a 'visited true) (< a b)) xs)
+(insert-sorted 'list xs (lambda (a b) (append! log a) (< a b)) item)
+
+;; GOOD — the predicate only compares
+(stable-sort (lambda (a b) (< a b)) xs)
+```
+
+Two spellings of the predicate are followed: an inline `lambda`, and a plain
+symbol naming a `defun` **in the same file**. That hop is one level deep — the
+named function's own body is scanned, but a call it in turn makes is not
+followed, so a comparator that mutates two hops away is not reported. Quoted
+subtrees are data and are skipped whole.
+
 ### `with-cleanup-forms`
 
 **Flags a degenerate `with-cleanup` spec list.** (Severity: warning)
--- a/docs/lint-checks.md
+++ b/docs/lint-checks.md
@@ -300,6 +300,46 @@ named function's own body is scanned, but a call it in turn makes is not
 followed, so a comparator that mutates two hops away is not reported. Quoted
 subtrees are data and are skipped whole.
 
+### `iteration-mutation`
+
+**Flags a callback that mutates the collection it is iterating.** (Severity:
+warning)
+
+Covers `map`, `foldl`, `foldr`, `select`, `reject`, `all?` and `any?`. A
+mutating call — `assoc!`, `dissoc!`, `append!`, `append-bytes!` or
+`stable-sort` — is reported when the value it writes through is the collection
+argument itself, or one of the callback's own parameters, which holds an
+element of that collection.
+
+`set!` is deliberately **not** on that list, unlike in `comparator-mutation`.
+It rebinds a name rather than writing through the value the name held, so
+`(set! x 1)` gives the callback's own parameter a new value and leaves the
+element alone, and `(set! xs ...)` rebinds the caller's variable while the
+builtin goes on walking the sequence it was already handed.
+
+```lisp
+;; BAD — writes through the sequence being walked
+(map 'list (lambda (x) (append! xs x)) xs)
+
+;; BAD — writes through an element the traversal handed over
+(map 'list (lambda (x) (assoc! x 'seen true)) xs)
+
+;; GOOD — the fold's own accumulator is neither the collection nor an element
+(foldl (lambda (acc x) (assoc! acc x 1)) (sorted-map) xs)
+
+;; GOOD — an unrelated binding
+(map 'list (lambda (x) (assoc! out x 1)) xs)
+```
+
+Both an inline `lambda` and a plain symbol naming a `defun` **in the same
+file** are followed, one hop deep.
+
+Blind spots worth knowing: the check is syntactic and keeps no scope of its
+own, so a callback parameter that an inner `let` rebinds is still treated as
+the element; a collection passed as an expression rather than a symbol has no
+name to match against and is invisible; and `zip` is not covered, because it
+takes no callback at all.
+
 ### `with-cleanup-forms`
 
 **Flags a degenerate `with-cleanup` spec list.** (Severity: warning)
--- a/docs/lint-checks.md
+++ b/docs/lint-checks.md
@@ -297,8 +297,15 @@ itself, which sorts in place despite carrying no `!`.
 Two spellings of the predicate are followed: an inline `lambda`, and a plain
 symbol naming a `defun` **in the same file**. That hop is one level deep — the
 named function's own body is scanned, but a call it in turn makes is not
-followed, so a comparator that mutates two hops away is not reported. Quoted
-subtrees are data and are skipped whole.
+followed, so a comparator that mutates two hops away is not reported.
+
+Data is skipped whole, and in all three spellings: a reader-quoted form
+(`'(assoc! a b)`), an explicit `(quote ...)` form, and a `quasiquote`
+template. The one exception is the standard one — an `(unquote ...)` or
+`(unquote-splicing ...)` subtree inside a template is evaluated where it
+stands, so a mutation in one is still reported. This applies to the sort form
+itself as much as to the predicate's body, and a `defun` written inside data
+defines nothing, so a symbol naming one resolves to no callback at all.
 
 ### `iteration-mutation`
 
@@ -332,7 +339,8 @@ builtin goes on walking the sequence it was already handed.
 ```
 
 Both an inline `lambda` and a plain symbol naming a `defun` **in the same
-file** are followed, one hop deep.
+file** are followed, one hop deep. Quoting is honoured exactly as it is for
+`comparator-mutation` above.
 
 Blind spots worth knowing: the check is syntactic and keeps no scope of its
 own, so a callback parameter that an inner `let` rebinds is still treated as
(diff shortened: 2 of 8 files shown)


sam-at-luther:
Both blockers are fixed on head b8a869c, one commit each, test-first with the reviewer's reproductions.

| Blocker | Commit | Fix | Proof |
|---|---|---|---|
| Diagnostic explosion: a same-file `defun` used as a callback was rescanned once per reference site and every mutation inside it reported again at the same position, so N references times M sites gave N*M identical diagnostics; nested lambdas were walked once per enclosing form | e0b5cfa | A per-pass `mutationRun` memoises the mutation sites of each callback body by its root node, so a body is scanned once however many forms name it, and deduplicates reports by (position, message). A scope key of (form, body, iterated collections) keeps distinct messages distinct: the same `defun` under `stable-sort` and under `insert-sorted`, or over two different collections, still gets one diagnostic each. An inner lambda is scanned once and its sites folded into the enclosing body's list. | `TestComparatorMutation_FanoutReportsEachSiteOnce` and the iteration twin: a `defun` with 20 mutating calls referenced 300 times reported 6000 diagnostics before and reports 20 after. `TestComparatorMutation_NestedCallbacksScannedOnce` and twin: five nested callbacks, exactly five diagnostics. `TestMutationChecks_FanoutAllocationBound`: a 300-reference file allocates about 29,000 times per lint against an asserted ceiling of 200,000, so a return to quadratic fails the test. |
| Quoted data reported as mutation: only the reader's quote flag was honoured, so an explicit `(quote ...)` form inside a callback, and a sort or iteration form sitting inside quoted data, were both treated as executing | b8a869c | Both the outer traversal and the body walk use one quote-aware walker: a reader-quoted node, an explicit `quote` form, and a `quasiquote` template are skipped whole, except `unquote` and `unquote-splicing` subtrees inside a template, which are evaluated where they stand and so are still checked, with nesting levels tracked so an unquote under a nested quasiquote stays data. `sameFileDefuns` uses the same walker, so a `defun` written inside data defines no callback. | Red before, green after: `TestComparatorMutation_Negative_ExplicitQuoteForm`, `_Negative_FormInsideQuotedData`, `_Negative_QuasiquoteTemplate`, `_Negative_NestedQuasiquoteLevel`, `TestMutationChecks_DefunInsideQuotedDataIsNotACallback`, each with its iteration twin where applicable; positive controls `_Positive_QuasiquoteUnquote` and `_Positive_UnquotedFormInTemplate` keep the evaluated subtrees reported. |

Docs updated in `docs/lint-checks.md` and both analyzer `Doc` strings to state the three data spellings and the unquote exception. Verification on b8a869c: `go test ./lint/` and `go test ./...` ok, `go test -race ./lint/` ok, elpsvet both modes exit 0, `elps lint --checks=comparator-mutation,iteration-mutation` over every `.lisp` file in the repo still reports nothing.

---

sam-at-luther:
Force-pushed with lease: head is now 60adde2. Same two fix commits as the comment above (now b633504 and 60adde2) with CI's golangci-lint v2.6.2 findings folded into the first one: two loops as `for range`, and the count assertions through a small `diagCount` helper rather than `assert.Len`, since a failing `Len` would print every one of the thousands of diagnostics the fixture exists to produce. The local linter (v2.5.0) had not reported them; a v2.6.2 build with CI's Go 1.25.13 toolchain reports zero issues on this head. The tree is byte-identical to the verified state: `go test ./...`, `go test -race ./lint/`, elpsvet both modes, gate self-test and the repo-wide `elps lint` sweep all clean.

---

sam-at-luther:
Implemented the remaining review fixes directly on this branch, in separate commits: 7bbbe11 bounds callback analysis (#649), a741148 matches quote traversal to real ELPS evaluation (#650), and 97e5ac6 clarifies the conservative lint policy. All 1,080 synthetic diagnostic comparisons agree with the prior implementation outside the intentional quote corrections. New tests fail before the fixes; disabling span merging makes both work-bound and exact-once tests fail. Final combined head passes targeted tests, checked/race mutation tests, and the exact CI linter. No runtime/template code or customer source changed. CI is now running on 97e5ac6.

sam-at-luther:
Independent re-check of 7bbbe11, a741148 and 97e5ac6 against the real interpreter. The two blockers they fixed hold: work bounds are exactly linear (allocations 45,974 / 91,537 / 182,599 / 364,669 at 200 / 400 / 800 / 1,600 sites, 3,000-deep nesting fine, self- and mutual recursion terminate), the `(position, message)` dedupe swallowed no distinct finding across every pairing tried, the fold accumulator positions match `builtinFoldLeft` / `builtinFoldRight` at runtime, and every claim in the docs commit lints exactly as printed. Overlaying #648 on this head merges clean and its registry drift test passes with both new rows present.

The re-check did find one gap the quote fix left behind, plus a smaller one. Both fixed on head 265113f, one commit each, test-first (#651).

| Gap | Commit | Fix | Proof |
|---|---|---|---|
| Qualified spellings evade both checks. a741148 made the data side aware of `lisp:quote` / `lisp:quasiquote`, but the hazard side matched bare names only, so `lisp:stable-sort`, `lisp:lambda`, `lisp:assoc!` or `lisp:map` each alone gave zero diagnostics while the code runs and mutates. | d52f6b7 | One canonicaliser strips a leading `lisp:` (and nothing else) at every head comparison in the two checks; diagnostics name the canonical spelling. `lisp:unquote` inside a template is deliberately left as data, because `getUnquoteType` compares the bare string literally and the interpreter treats it as data. | 14 table cases plus two qualified contexts in the paired evaluator/linter suite: 40 subtests red before, 90 green after. `foo:map` / `foo:assoc!` negative control stays silent. |
| Duplicate `defun`: the checker scanned the first definition, the interpreter runs the last. | 265113f | `sameFileDefuns` is last-wins. Memo and dedupe untouched. | Dirty-second now reported, dirty-first no longer reported, last-of-three wins: 6 subtests red before, 7 green after. Interpreter prints only `"SECOND body ran"`. |

Also documented as a limit, not fixed: a same-file `(defun quote (x) x)` shadows the special form and both checks then read the shadowed meaning. That is a property of every syntactic analyzer in `lint/`, not of these two.

Verification on 265113f: `go test ./... -count=1` all ok; `-tags elpscheck -race ./lint/` ok; elpsvet both modes exit 0; golangci-lint 2.6.2 `0 issues`; `scripts/ci-gates-test.sh` 389 passed, 0 failed; `elps doc -m` clean; `elps lint --list` transcript unchanged. `BenchmarkMutationChecksScaling` allocations unchanged to within one allocation and still 3.97 to 3.99x from 200 to 800.

---

sam-at-luther:
Final verification: 97e5ac6 is now green on all 25 checks, including full tests, fuzz and the benchmark gate. A local combined #647 + #648 overlay also passes the MCP Help/Lint tests under race. No remaining blocker from this review; ready for merge approval.

sam-at-luther:
Note for whoever approves: the comment above names 97e5ac6, but the head is now 265113f, which adds the two #651 fixes on top of it (qualified `lisp:` spellings and last-wins duplicate `defun`). 265113f is green on all 25 checks as of 03:48 UTC, including the benchmark gate, and the #648 overlay result holds on it too. Nothing else changed.

---

PR #648, opened 2026-09-09
mcpserver: generate the help prompt's lint analyzer table from the registry

### LINKED ISSUE #646: mcpserver: the "Lint Analyzers" prompt table names analyzers that do not exist
## Problem

`mcpserver/service.go` carries a hard-coded "Lint Analyzers" table in its prompt text (around lines 2195 to 2215). Several rows name analyzers that are not registered in `lint.DefaultAnalyzers()` and never were, for example `handler-bind-structure`, `sort-stable-comparator` and `assert-arity`, and the table omits real ones. An MCP client reading the prompt is told to expect checks that `elps lint` cannot run, and is not told about checks it does run.

Found while adding `comparator-mutation` (#644) and `iteration-mutation` (#645); left untouched there to keep those commits focused.

## Proposed fix

Generate the table from `lint.DefaultAnalyzers()` at prompt-build time (name, severity, first line of `Doc`) instead of maintaining a literal, and add a test that the prompt's analyzer names equal `lint.AnalyzerNames()` so it cannot drift again. `docs/nolint.md`'s `elps lint --list` transcript had the same drift and is refreshed by the #644/#645 PR; the same test could cover it if the docs are read by a test.

sam-at-luther:
## Summary

The MCP server's help prompt carried a hand-written "Lint Analyzers" table that had drifted a long way from `lint.DefaultAnalyzers()`: nine rows named analyzers that do not exist (`assert-arity`, `cond-else`, `deftype-structure`, `export-form`, `handler-bind-structure`, `lambda-structure`, `missing-package-qualifier`, `sort-stable-comparator`, `test-structure`), eleven registered analyzers were missing, and `in-package-toplevel` was listed as `error` when the registry says `warning`. An MCP client reading the prompt was told to expect checks `elps lint` cannot run.

The table is now built at package init from the registry: one row per analyzer, sorted by name, with its name, severity, the first sentence of its `Doc`, and a new `Semantic` column saying whether it needs workspace analysis. The three existing columns keep their header text and order, so a consumer reading the first three cells sees the same shape. `mcpserver` already imported `lint` for its lint tool, so no new dependency and no cycle.

## Test

`TestHelpAnalyzerTableMatchesRegistry` parses the rendered prompt and asserts the row set equals `lint.AnalyzerNames()`, and that each row's severity and semantic flag match the registered analyzer. On the previous head it fails naming the nine fictional and eleven missing rows; on this head it passes. The existing `TestHelpTool` substring assertions are unchanged and still pass.

## Also found, not fixed here

Other hand-maintained analyzer lists outside `mcpserver`: `docs/lint-checks.md` has no section for `duplicate-definition`; `docs/lsp-guide.md` has a partial prose enumeration; two historical design docs under `docs/` list old tiers. `docs/nolint.md`'s `elps lint --list` transcript is regenerated in #647.

## Verification

- `go build ./...`, `go vet ./mcpserver/` clean; `gofmt -l mcpserver/` reports only the pre-existing `types.go` alignment on main, untouched
- `go test ./mcpserver/ -count=1 -v`: new test red on the previous head, green here; `go test ./... -count=1`: all ok
- `go run ./cmd/elpsvet -test=false ./...` in both modes: exit 0
- `make static-checks`: only the two documented pre-existing `nolintlint` findings from the golangci-lint version skew
- `scripts/ci-gates-test.sh`: 0 failed; `./elps doc -m`: all documented

Closes #646

https://claude.ai/code/session_01BHMPMnEJgGRuJioE5ddat6

---
_Generated by [Claude Code](https://claude.ai/code/session_01BHMPMnEJgGRuJioE5ddat6)_

### DIFF
--- a/mcpserver/help_analyzers_test.go
+++ b/mcpserver/help_analyzers_test.go
@@ -1,0 +1,138 @@
+package mcpserver
+
+import (
+	"sort"
+	"strings"
+	"testing"
+
+	"github.com/luthersystems/elps/lint"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// helpAnalyzerRow is one parsed row of the help prompt's "Lint Analyzers"
+// table.
+type helpAnalyzerRow struct {
+	Name        string
+	Severity    string
+	Semantic    string
+	HasSemantic bool
+}
+
+// parseHelpAnalyzerRows extracts the rows of the "## Lint Analyzers" markdown
+// table out of the help prompt, resolving cells by their header name so the
+// test does not depend on column order.
+func parseHelpAnalyzerRows(t *testing.T, content string) []helpAnalyzerRow {
+	t.Helper()
+	_, after, found := strings.Cut(content, "## Lint Analyzers")
+	require.True(t, found, "help prompt must carry a Lint Analyzers section")
+	section, _, _ := strings.Cut(after, "\n## ")
+
+	cells := func(line string) []string {
+		line = strings.TrimSpace(line)
+		line = strings.TrimPrefix(line, "|")
+		line = strings.TrimSuffix(line, "|")
+		out := strings.Split(line, "|")
+		for i := range out {
+			out[i] = strings.TrimSpace(out[i])
+		}
+		return out
+	}
+
+	var header []string
+	var rows []helpAnalyzerRow
+	for _, line := range strings.Split(section, "\n") {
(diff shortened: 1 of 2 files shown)


sam-at-luther:
Final recheck at 87683b9: no remaining code blocker. The corrected semantic-help text matches the implementation; generated analyzer table still derives from the current registry. Focused MCP Help/Lint tests pass locally and all 25 checks are green on this exact head. No additional source change needed; the remaining lint fixes are now on #647.

Issue #638, opened 2026-09-09
v1.61.0 VS Code publication blocked by repeated Marketplace gallery timeouts


sam-at-luther:
## Observed failure

The v1.61.0 Go release and immutable tag are published at `31d1116e16a2fbf63ad973da27a998c6269ef2e2`, and all four release binary builds pass. VS Code Marketplace publication is blocked by a reproducible three-minute `Request timeout: /_apis/gallery` in `vsce publish`.

- Workflow: https://github.com/luthersystems/elps/actions/runs/34292740674
- Attempt 1: universal and darwin-x64 publishing timed out; remaining platform legs were cancelled by matrix fail-fast.
- One scoped failed-job retry reused successful binary builds. Attempt 2: universal and linux-x64 again timed out; remaining platform legs were cancelled.
- Attempt 2 universal evidence: https://github.com/luthersystems/elps/actions/runs/34292740674/job/102284941833 (publishing 00:05:47 UTC; timeout 00:08:47 UTC on September 9).
- Dependency install, extension build, and packaging succeeded. The reported error is a network timeout, not an explicit invalid-token response. Root cause beyond this is unconfirmed.
- Public Marketplace listing still referenced 1.60.0 before retry; that alone cannot exclude an upload accepted but not indexed.
- Prior four release workflows (v1.57.0 through v1.60.0) succeeded on original attempts.

## Recovery requirements

- [ ] Diagnose Marketplace/service connectivity from the publishing runner, without exposing or rotating credentials speculatively.
- [ ] Check version/platform state before further publishing attempts; publish only missing packages.
- [ ] Complete publication of universal, linux-x64, linux-arm64, darwin-x64, and darwin-arm64 packages at 1.61.0.
- [ ] Verify the public listing and record recovery evidence.

Do not move the release tag or weaken test/benchmark gates. This is separate from the missing-CLI release-helper failure in #637. ELPS #623 is merged; coordinated Substrate #483 can consume the published Go module regardless of Marketplace availability.

sam-at-luther:
[...]

Issue #637, opened 2026-09-08
Release-tag helper fails before execution: Claude native binary missing


sam-at-luther:
The required v1.61.0 dry run failed before publishing anything on merged commit31d1116: https://github.com/luthersystems/elps/actions/runs/34292534442/job/102281990685 . The Claude SDK reports executable_not_found / ENOENT for /home/runner/.local/bin/claude. Checkout and authentication setup succeeded; no release/tag was created by this run. Investigate the floating anthropics/claude-code-action@v1 installation/runtime contract and add a dry-run regression that actually reaches release validation. For the operator-approved v1.61.0 release, use the existing documented gh release path with the exact tested commit, then verify the tag-triggered publication workflow. Do not mask this failed dry run or change repository permissions/secrets to bypass it.

sam-at-luther:
Recovered upstream; verified without a local workflow, permission, runner-image or secret change. The current release-helper dry run successfully starts the CLI and reaches version validation, matching main SHA, commit collection and generated notes: https://github.com/luthersystems/elps/actions/runs/34441785806 . Its retained trace records only three read-only tool calls and no release/tag creation: https://github.com/luthersystems/elps/actions/runs/34441785806/artifacts/10138203049 . The failed run used CLI/SDK 2.1.265/0.3.265; this successful run uses 2.1.267/0.3.267. The exact upstream installer repair is not established, but the missing-executable failure and required dry-run acceptance are demonstrably resolved. Marketplace publication remains separately tracked in #638; this closure does not claim that publishing is repaired.

PR #653, opened 2026-09-10
Close ELPS backlog gaps in LSP lookup and walker verification

### LINKED ISSUE #598: The alias oracle only covers Template.NewVM; copy, detach, (*LVal).Copy and the macro stamper are undriven

### LINKED ISSUE #600: Alias-guard channels still unwatched in copy/detach/(*LVal).Copy and the macro stamper

### LINKED ISSUE #641: LSP: definition, completion and hover split the whole document on every request
## Problem

`wordAtPosition` (`lsp/position.go:134`) does `strings.Split(content, "\n")` over the entire document, and the `wordBoundsAtPosition` it calls (`lsp/position.go:158`) splits the entire document again. `wordRangeAtPosition` (`lsp/position.go:145`) calls `wordBoundsAtPosition` too, so hover pays a third split.

Callers: `lsp/definition.go:33`, `lsp/completion.go:33`, `lsp/hover.go:44` and `:45`.

So every `textDocument/definition` and `textDocument/completion` request allocates two full copies of the document's lines, and hover three, regardless of size. Measured with an 8 MiB document: definition and completion allocate 8× the document, hover more.

The over-limit guard from #609 stops this only for documents above `maxDocumentBytes`. Under the limit (default: no limit; 5 MiB is common), a large file still pays two splits per keystroke of completion.

## Fix

`lineOf(content, line)` already exists at `lsp/position_encoding.go:131`. It walks to the requested line with `strings.IndexByte` and allocates nothing. `wordAtPosition` and `wordBoundsAtPosition` should take the single line from it instead of splitting the document.

Found during the review of #609.

sam-at-luther:
## Summary

ELPS backlog closeout on one branch, with a separate commit for each work item. Substrate is untouched and remains under user review.

| Ticket | Work / status |
|---|---|
| #641 | Remove whole-document line-slice allocations from LSP word/bounds/range lookups; preserve cursor and symbol semantics. |
| #598 / #600 | Shared behavioral checker over current copy/detach/Go Copy/macro-stamp/template operations, with real mutation controls and bounded fuzzing. |
| #637 | Closed separately: actual release-helper dry run proves upstream CLI startup recovery, without local workflow/secret changes. |
| #638 | Add guarded read-only manual Marketplace diagnostics and recovery runbook. Both runner hosts answered promptly; authenticated discovery and publication remain unverified. **Not claimed fixed or closed**. |

Closes #641. Closes #598. Closes #600. Partially addresses #638.

## Verification

- LSP: six allocation assertions fail old code and pass the fix. Independent token-boundary model and UTF-8/UTF-16 cursor tests; targeted normal, checked and race tests pass. Exact CI linter passes.
- LSP allocation counts: word 2→0, bounds 1→0, range 2→1. Eighteen before/after benchmark rows are directionally faster; noisy timing intervals prevent exact timing certification. Late-line lookup still scans preceding lines.
- Walkers: five real adapters, structural registry coverage, independent alias/isolation/metadata observations, and per-channel broken-adapter controls. Nine production mutations were each detected and restored. Tests/docs only: no runtime performance or public API change. Targeted checked/race tests pass; bounded fuzz smoke completed 7,831 executions in 10 seconds. Independent final QA passes. See `docs/walker-oracle.md` for the finite supported contracts, exceptions, and reproduction commands; existing cold/template parity tests remain intact.
- Diagnostics: 12 real HTTP, locked SDK, and subprocess tests plus workflow isolation checks pass. Removing body/deadline/environment/redirect protections fails controls. Independent QA/security review passes.
- Manual diagnostics have read-only contents permission, no publishing token and no lifecycle scripts. Build and publication jobs require a tag-push event. Diagnostic success is not evidence of successful publication.
- Actual read-only [runner comparison](https://github.com/luthersystems/elps/actions/runs/34444476770): public inventory HTTP 200, unauthenticated SDK HTTP 401 in 67 ms on ARM / 163 ms on x64; all production jobs skipped. Latest public version remained 1.60.0. This does not establish authenticated recovery. Shipped diagnostics use one literal production-equivalent ARM runner.
- First-head CI correctly rejected the diagnostic runner matrix under two existing fleet guards. The follow-up removes the matrix; no fleet exemption or guard weakening. Final local gate in benchmark-CI configuration: 210 passed, 0 failed.
- Combined-head CI also caught a direct header struct copy in the new truncated-cycle negative control. Reproduced locally, then routed through the existing audited header-view constructor, preserving the exact intentional defect. The source-copy guard and behavioral controls now pass together under checked/race; no exemption was added.
- Previously reviewed head `83c6aee`: full tests, full race, lint, both analyzer modes and all 11 fuzz shards passed in CI. The new walker target ran 18,842 executions during its 30-second CI budget. These historical run counts are not comparable across revisions/machines.
- [Benchmark gate on reviewed head 83c6aee](https://github.com/luthersystems/elps/actions/runs/34445842926): PASS, 47 delta rows + 801 no-change rows, zero rows at or above the unchanged 15% timing / 5% allocation thresholds; ten interleaved samples per arm on the same runner. Five statistically significant bad-direction metric rows were below the gate (timing changes at most +2.74%); this is not a claim that every timing row improved. LSP allocation improvement is separately measured above; these new LSP microbenchmarks are not in the existing CI comparison set.

## Review follow-up

| Finding | Commit | Fix / proof |
|---|---|---|
| Post-write snapshot errors mislabeled as isolation failures | ab7db35 | Wrap the real cause with arm/phase context. A 200-map regression fails old code; both error cause and classification are asserted. |
| Control suites silently change walkers after registry reorder | f9c6d69 | Select by stable names, reject missing/duplicate names, and run the real control suites under every rotation. Reverting their actual selections to indices fails. |
| Mutation/throughput documentation overclaim | e70ed91 | State that the common test was run alone, not that other tests stayed green. Remove stale machine-specific throughput from the committed guide. |
| Lexicographic version ordering | 0e27792 | Numeric triplet comparison without integer rounding, tested across digit boundaries and components above 2^53. |
| Unbounded manual diagnostics | 50b25bc | Fixed cross-branch, diagnostic-only job concurrency. Tag publication cannot share its cancellation group; mutation tests guard scope. |
| Missing Node handling differs from existing gate | 3c2a65b | Explicit optional-Node skip; a present Node must still propagate failed test execution. Both paths run in the test. |

Combined local verification on `e70ed91`: targeted checked/race oracle and structural guards pass (7.872s); exact CI linter reports zero issues; 11 Node and 4 Python tests pass; benchmark-CI-config gate 210/0. Independent QA/security review: PASS.

Final-head CI on `e70ed91`: **25/25 checks passed**, including full tests/race, both analyzer modes, all 11 fuzz shards and the unchanged benchmark gate. The walker fuzz target completed 18,744 executions during its 30-second budget. [Benchmark result](https://github.com/luthersystems/elps/actions/runs/34494138586): 49 delta rows + 799 no-change rows, zero rows at or above the 15% timing / 5% allocation thresholds, ten interleaved samples per arm. Nine significant adverse timing rows remained below the gate (largest +4.88%); not a claim that every timing row improved.

The [updated manual diagnostic run](https://github.com/luthersystems/elps/actions/runs/34494073897) passed with all three production build/publish jobs skipped. No authenticated probe or publication was performed.

Pre-existing MCP whole-document splitting and ampersand lookup mismatch independently reproduced and filed as #654. No MCP runtime change is bundled here. The minor existing repeated hover line scans are not a correctness regression; this follow-up leaves that separate optimization unchanged.

Not merged. Marketplace #638 remains explicitly unresolved, with authenticated diagnosis/publication recovery outside this read-only implementation.

No customer code, new embedding APIs, gate waivers, or production publication are included.

### DIFF
--- a/.github/workflows/vscode-publish.yml
+++ b/.github/workflows/vscode-publish.yml
@@ -9,16 +9,12 @@ on:
 
 jobs:
   diagnose-marketplace:
-    name: Diagnose Marketplace (${{ matrix.runner }})
+    name: Diagnose Marketplace (ARM)
     if: github.event_name == 'workflow_dispatch'
     permissions:
       contents: read
-    runs-on: ${{ matrix.runner }}
+    runs-on: ubuntu-24.04-arm
     timeout-minutes: 3
-    strategy:
-      fail-fast: false
-      matrix:
-        runner: [ubuntu-24.04-arm, ubuntu-24.04]
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7
         with:
@@ -34,6 +30,9 @@ jobs:
       - name: Report unauthenticated discovery and published version inventory
         run: node scripts/marketplace-diagnostics.cjs
 
+  # TODO(marketplace-638): authenticated discovery and missing-target publication
+  # recovery remain unverified; diagnostics above intentionally cannot use a PAT
+  # or publish. Track recovery at https://github.com/luthersystems/elps/issues/638.
   build-binaries:
     if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     name: Build elps (${{ matrix.target }})
--- a/.github/workflows/vscode-publish.yml
+++ b/.github/workflows/vscode-publish.yml
@@ -1,12 +1,41 @@
 name: Publish VS Code Extension
 
 on:
+  # Manual runs are read-only diagnostics; only tag pushes can publish.
+  workflow_dispatch:
   push:
     tags:
       - "v*"
 
 jobs:
+  diagnose-marketplace:
+    name: Diagnose Marketplace (${{ matrix.runner }})
+    if: github.event_name == 'workflow_dispatch'
+    permissions:
+      contents: read
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 3
+    strategy:
+      fail-fast: false
+      matrix:
+        runner: [ubuntu-24.04-arm, ubuntu-24.04]
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7
+        with:
+          persist-credentials: false
+      - uses: actions/setup-node@820762786026740c76f36085b0efc47a31fe5020  # v7
+        with:
+          node-version: "20"
+      - name: Install the locked publishing client without lifecycle scripts
+        working-directory: editors/vscode
+        run: npm ci --ignore-scripts --no-audit --no-fund
+      - name: Verify the locked SDK probe against local HTTP servers
+        run: node --test scripts/marketplace-sdk.test.cjs
+      - name: Report unauthenticated discovery and published version inventory
+        run: node scripts/marketplace-diagnostics.cjs
+
   build-binaries:
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     name: Build elps (${{ matrix.target }})
     # ARM for the reason spelled out on elps.yml's lint-and-test job -- this
     # is the HOST the cross-compile runs on, and nothing else.
@@ -65,6 +94,7 @@ jobs:
           path: editors/vscode/bin/
 
   publish-platform:
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     name: Publish (${{ matrix.target }})
     needs: build-binaries
     # ARM for the reason on elps.yml's lint-and-test job. `--target` below is
@@ -126,6 +156,7 @@ jobs:
           VSCE_PAT: ${{ secrets.VSCE_PAT }}
 
   publish-universal:
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
     name: Publish (universal)
     needs: build-binaries
     # ARM for the reason on elps.yml's lint-and-test job; same npm/esbuild
--- a/.github/workflows/vscode-publish.yml
+++ b/.github/workflows/vscode-publish.yml
@@ -15,6 +15,11 @@ jobs:
       contents: read
     runs-on: ubuntu-24.04-arm
     timeout-minutes: 3
+    # One read-only probe at a time, including dispatches from other branches.
+    # Job scope is deliberate: tag publication must never share cancellation.
+    concurrency:
+      group: elps-marketplace-readonly-diagnostics
+      cancel-in-progress: true
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7
         with:
--- a/docs/marketplace-recovery.md
+++ b/docs/marketplace-recovery.md
@@ -1,0 +1,75 @@
+# VS Code Marketplace diagnostics and recovery
+
+The Go release and the Marketplace extension are separate publications. A
+published Go module does not establish that every VSIX target is available.
+
+## Diagnose without publishing
+
+The existing `Publish VS Code Extension` workflow has a **read-only manual
+mode**. Tag pushes still build and publish the original five packages. Manual
+runs cannot build or publish, and do not receive `VSCE_PAT`.
+
+```sh
+gh workflow run vscode-publish.yml --repo luthersystems/elps --ref <reviewed-branch>
+```
+
+Two jobs compare Ubuntu 24.04 ARM and x64 using Node 20, the publishing runtime.
+They report only public addresses, timings, HTTP status codes, and a validated
+version/platform inventory for `LutherSystems.elps-lang`:
+
+- unauthenticated `OPTIONS /_apis/gallery`, including DNS, connect, TLS and
+  first-byte timing;
+- the public, read-only extension query (POST is the API's query method);
+- the locked `azure-devops-node-api` client's `getExtension` call used by `vsce`
+  before upload, **without** an authentication handler.
+
+Raw requests have a 15-second absolute deadline and a 1 MiB response limit. The
+SDK subprocess has an independent 20-second deadline. Response headers, bodies,
+exception messages, cookies, and environment variables are never logged. The
+SDK subprocess receives an empty environment; dependency lifecycle scripts are
+disabled. Its authenticated behavior is deliberately not tested.
+
+**A green diagnostic job means the diagnostic ran, not that publishing works.**
+Read its JSON report. `401` from unauthenticated discovery proves an HTTP
+response arrived, not that a PAT is invalid. A successful public query establishes
+what is currently indexed, not whether an unindexed upload has been accepted.
+Compare raw transport with the SDK result and compare both runner hosts before
+attributing a failure to the client or host. An SDK-only failure warrants examining
+the locked client's request behavior; neither result justifies rotating secrets.
+
+Local checks (Node and Python with PyYAML, also used by the existing CI guards):
+
+```sh
+node --test scripts/marketplace-diagnostics.test.cjs
+python3 scripts/marketplace-workflow-test.py
+```
+
+## Recovery requires an explicitly approved publishing operation
+
+1. Record the diagnostic run and the desired release's five targets: universal,
+   linux-x64, linux-arm64, darwin-x64, darwin-arm64. Confirm missing targets in the
+   publisher's management view too: the public index may lag an accepted upload.
+2. Resolve the diagnosed cause. Do not speculate about PAT expiry from a network
+   timeout, or switch runner architectures merely because the failing host is ARM.
+3. Publish **only missing targets**, using the immutable release source and
+   artifacts. A failed-job rerun is safe only after checking that none of those
+   jobs already uploaded its target. Do not move a release tag or rerun all jobs
+   blindly. Manual diagnostic dispatch is not a recovery/publishing command.
+4. Verify all five version/target entries after indexing and record the recovery
+   evidence on [#638](https://github.com/luthersystems/elps/issues/638).
+
+## Current evidence for #638
+
+The original [v1.61.0](https://github.com/luthersystems/elps/actions/runs/34292740674)
+and [v1.61.1](https://github.com/luthersystems/elps/actions/runs/34308282015) failures
+stop at `/_apis/gallery` discovery, before the extension lookup/upload completes.
+Both time out at 180 seconds; binary builds and packaging succeed. The same ARM
+workflow successfully published v1.60.0. The relevant `vsce` 3.9.2,
+`azure-devops-node-api` 12.5.0 and transport dependency pins did not change.
+
+On September 9, 2026, an unauthenticated local probe returned discovery HTTP 401
+and public-query HTTP 200; the public inventory contained all five v1.60.0
+targets and no v1.61.0/v1.61.1 targets. This is **not runner-side or authenticated
+recovery evidence**. #638 remains open until Marketplace publication is actually
+recovered. The diagnostic workflow supplies the missing runner-side evidence
+without turning an investigation into an accidental release.
--- a/docs/marketplace-recovery.md
+++ b/docs/marketplace-recovery.md
@@ -13,8 +13,8 @@ runs cannot build or publish, and do not receive `VSCE_PAT`.
 gh workflow run vscode-publish.yml --repo luthersystems/elps --ref <reviewed-branch>
 ```
 
-Two jobs compare Ubuntu 24.04 ARM and x64 using Node 20, the publishing runtime.
-They report only public addresses, timings, HTTP status codes, and a validated
+The diagnostic job uses Ubuntu 24.04 ARM and Node 20, matching publication.
+It reports only public addresses, timings, HTTP status codes, and a validated
 version/platform inventory for `LutherSystems.elps-lang`:
 
 - unauthenticated `OPTIONS /_apis/gallery`, including DNS, connect, TLS and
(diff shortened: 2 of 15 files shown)


sam-at-luther:
Review findings 1–5 and the optional-Node cleanup are addressed in six separate commits on e70ed91; the PR body maps each finding to its fix and regression proof.

- The 200-map case now reports a wrapped traversal-limit error, not a false isolation failure.
- The real negative-control suites run under every registry rotation; restoring positional selections turns them red.
- Mutation claims are corrected; versions sort numerically; manual diagnostic concurrency is isolated from tag publishing; absent Node skips explicitly while actual Node test failures still fail.

Independent QA/security review passes. Final-head CI is **25/25 green**: full tests/race, both analyzer modes, all fuzz shards, and the unchanged benchmark gate (zero rows at or above threshold). The walker target completed 18,744 executions in its 30-second budget. The updated manual workflow passed with every production build/publish job skipped.

The separately reproduced MCP lookup issue is #654. The minor existing repeated hover scans are left unchanged; no VM/runtime Go code changed in this review follow-up. Not merged; #638 remains open for authenticated publishing diagnosis/recovery.

